PR #7, opened 2014-10-30
added div statements to 01-filedir lesson that center images (refactor t...

psteinb:
...his to sofwarecarpentry/bc css?)


gvwilson:
This one now has a ton of conflicts - can it be resurrected, or should we close and move on?

wking:
On Fri, Jan 16, 2015 at 05:53:56AM -0800, Greg Wilson wrote:
> This one now has a ton of conflicts - can it be resurrected, or
> should we close and move on?

It should be resurectable (along with all the other pull requests
based on the bc content that weren't addressed before the transition
to the new layout).  We just need someone to go through and review the
content changes they suggest.  In this case, that review is just for
the final commit [1].

[1]: https://github.com/psteinb/swc-modular-shell/commit/aa1a5a3c336c246cf26ab0223e8fd96e52b18cf5

PR #16, opened 2014-11-15
Csh and its incompatibility with bash mentioned

kpalamartchouk:
A lot of C-shell code is still in use in science, so it would be nice to warn the students about the incompatibilities between csh and bash


gvwilson:
Replaced and closed by 9951dff3

PR #18, opened 2014-11-15
Csh and its incompatibility with bash mentioned

kpalamartchouk:
A lot of csh code is in use in science. Students should be warned about the incompatibilities between csh and bash.


ChristinaLK:
These changes seem to have been moved to the `discussion.md` file.  Anything else you want me to do here @gvwilson ?

PR #23, opened 2014-11-24
Small changes in shell course for an assignment of the instructor's training

rahmanisahar:
Add a note for changing directory part in case of file name's contains space on it, how ppl should type it after cd command. This is one of most command issues for novices during shell course, especially ppl who works with Windows.


ChristinaLK:
I'll turn this into an issue for a future contributor.

PR #31, opened 2014-12-19
02-filedir.md: Fill in FIXME challenge titles

wking:
It's hard to name these in a way that's both descriptive and avoids
giving away the answer, but I've tried ;).


gvwilson:
Closed as a side effect of 227cf48056.

PR #33, opened 2014-12-19
04-loop.md: Fill in FIXME challenge titles

wking:
Naming the last challenge is difficult, because it's not really about
loops.  That challenge dates back to 8e77786c (Importing shell lesson
for novices, 2013-11-11), which doesn't explain the pedagogical
motivation for the challenge.  Maybe it's just a trick to keep
students from being to focused on the purported topic when crafting
their answers?  Maybe it's intentionally a throwback to more basic
shell syntax couched in the decoration of the current topic?  Maybe
it's a hint that you can do more with loops than vary arguments to a
single command?  In any event, I'd prefer removing this challenge
entirely, or reworking the content as an example (instead of a
challenge).

### DIFF
--- a/04-loop.md
+++ b/04-loop.md
@@ -393,7 +393,7 @@ so she decides to get some coffee and catch up on her reading.
 > then she can re-run `goostats` on `NENE01729B.txt` simply by typing
 > `!458`.
 
-> ## FIXME {.challenge}
+> ## Loop reading comprehension {.challenge}
 > 
 > Suppose that `ls` initially displays:
 > 
@@ -410,7 +410,7 @@ so she decides to get some coffee and catch up on her reading.
 > done
 > ~~~
 
-> ## FIXME {.challenge}
+> ## Loop reading comprehension with `echo` {.challenge}
 >
 > In the same directory, what is the effect of this loop?
 > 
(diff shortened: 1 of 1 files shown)


gdevenyi:
Duplicate of another merge, its names were much better.

wking:
On Sun, Feb 01, 2015 at 05:44:40PM -0800, Gabriel A. Devenyi wrote:
> Duplicate of another merge, its names were much better.

Which merge?  I still see challenge FIXMEs in the current gh-pages:

  $ git fetch origin
  …
  From git://github.com/swcarpentry/shell-novice
	   a46d1b0..aa0b1b1  gh-pages   -> origin/gh-pages
  $ git grep FIXME origin/gh-pages 04-loop.md
	origin/gh-pages:04-loop.md:> ## FIXME {.challenge}
	origin/gh-pages:04-loop.md:> ## FIXME {.challenge}
	origin/gh-pages:04-loop.md:> ## FIXME {.challenge}
	origin/gh-pages:04-loop.md:> ## FIXME {.challenge}

I agree that the names aren't great, but I think that's because the
challenges are not tightly focused ;).  I'm happy to discuss revisions
to both the headings and the challenges themselves.

PR #36, opened 2014-12-19
index.md: Fill in the Prerequisites FIXME

wking:
I also added anti-prerequisites, since it seems useful to bracket our
target audience on both sides (both "not ready yet" and "too advanced
already").

We try and make this lesson as accessible as possible, but I think
folks who are manipulating files on a computer for the first time
would still have trouble keeping up.  The lesson should be for folks
who have done similar filesystem manipulations before, just not via
the shell interface.  Phrasing that in a way that makes sense to folks
who aren't used to the word "filesystem" is difficult, but I've tried
my best ;).

### DIFF
--- a/index.md
+++ b/index.md
@@ -12,9 +12,14 @@ so that they don't have to type the same things over and over again.
 
 > ## Prerequisites
 >
-> FIXME
-> A short paragraph describing what learners need to know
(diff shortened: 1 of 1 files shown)


gdevenyi:
Looks good.

PR #61, opened 2015-02-03
.gitignore: Remove _site, .DS_Store, and *~

wking:
Explanations in the commit messages.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,4 @@ FAQ.html
 DESIGN.html
 *~
 *.pyc
-_site
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@ FAQ.html
 DESIGN.html
 *~
 *.pyc
-.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,4 @@ README.html
 LAYOUT.html
 FAQ.html
 DESIGN.html
-*~
 *.pyc


gvwilson:
We should probably put everything in lesson-template's .gitignore so that it can be copied over as needed to other projects...

wking:
On Tue, Feb 03, 2015 at 10:34:31AM -0800, Greg Wilson wrote:
> We should probably put everything in lesson-template's .gitignore so
> that it can be copied over as needed to other projects...

I imagine this file is low-churn enough that we don't need the extra
efficiency there.  I also expect that different lessons may want to
ignore different files (e.g. there may be some R-specific files that
they want to ignore in the R lesson).

r-gaia-cs:
> We should probably put everything in lesson-template's .gitignore so that it can be copied over as needed to other projects...

I would prefer that people use [a global .gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore). But I'm OK with local `.gitignore`.

wking:
On Thu, Feb 05, 2015 at 01:24:02PM -0800, Raniere Silva wrote:
> > We should probably put everything in lesson-template's .gitignore
> > so that it can be copied over as needed to other projects...
>
> I would prefer that people use a global .gitignore. But I'm OK with
> local `.gitignore`.

That's mostly for stuff that *you* create that shouldn't be versioned
(E.g. .DS_Store, *~).  For common side-effects of repository-specific
actions (e.g. *.pyc files), it's better to use an in-repo .gitignore.
That way folks don't have to create a per-user .gitignore unless they
create wonky stuff like .DS_Store.  It won't cause a problem if you
have *.pyc in *both* your per-user $HOME/.config/git/ignore (or
~/.gitignore_global if you decide to set core.excludesfile and point
it there).

Issue #41, opened 2015-01-16
Add a dotfile to the image related to ls -F -a command


willingc:
dotfiles are a concept that trip up new users of the command line and shells. In file `01-filedir.md`, the description of the `..` directory is great. Although `.` is mentioned related to files, it would be helpful to learners to see a dotfile in the output image or to have a sentence explaining dotfiles for configuration information.

![screen shot 2015-01-16 at 11 57 18](https://cloud.githubusercontent.com/assets/2680980/5783380/3f1df878-9d77-11e4-8064-aa990d48e05d.png)

lexnederbragt:
Hmmm... It is not straightforward to come up with a dotfile that is relatively easy to explain. The file should be there in the actual folder when they clone/download the repo. I was thinking a .bashrc file with just one command, but it should then be one that doesn't trip the student. Even a simple thing as an alias (alias la='ls -la') would require a few lines to explain. Is there a type of dotfile that makes sense, but has no content, or one that can be easily understandable?

wking:
On Mon, Feb 02, 2015 at 06:52:35AM -0800, Lex Nederbragt wrote:
> The file should be there in the actual folder when they
> clone/download the repo.

.gitignore?

lexnederbragt:
On 2 Feb 2015, at 18:13, W. Trevor King wrote:

> On Mon, Feb 02, 2015 at 06:52:35AM -0800, Lex Nederbragt wrote:
> > The file should be there in the actual folder when they
> > clone/download the repo.
> 
> .gitignore?

Maybe. It would mean adding a .git folder as well, with the burden of having to explain these - and why would the entire 'nelle' folder be under version control…

	Lex

wking:
On Tue, Feb 03, 2015 at 01:59:03AM -0800, Lex Nederbragt wrote:
> Maybe. It would mean adding a .git folder as well, with the burden
> of having to explain these - and why would the entire 'nelle' folder
> be under version control…

I'd have the students getting the example files via a Git repository
(either via a ‘git clone …’ or via an archive zip [1]).  That's going
to come with a .gitignore (and .nojekyll) out of the box.  The clone
will also come with a .git.  The shell lesson usually lands before the
Git lesson, so I'd just say something like:

  The .gitignore file has a leading dot to mark it as a hidden file
  (or directory).  Dotfiles are just like a regular files, except that
  `ls` doesn't list them by default and the shell doesn't use them
[...]

PR #63, opened 2015-02-05
Add .bash_profile dotfile to output of ls -F -a, and explanation thereof.

lexnederbragt:
Adresses #41.

### DIFF
--- a/01-filedir.md
+++ b/01-filedir.md
@@ -337,7 +337,7 @@ it also displays another special directory that's just called `.`,
 which means "the current working directory".
 It may seem redundant to have a name for it,
 but we'll see some uses for it soon.  
-Finally, we now see a file called `.bash_profile`. This file usually contains settings to customize the shell (terminal). It may also be called `.bashrc` or `.bash_login`. For this lesson material it does not contain any settings.
+Finally, we also see a file called `.bash_profile`. This file usually contains settings to customize the shell (terminal). There may also be similar files called `.bashrc` or `.bash_login`. For this lesson material it does not contain any settings.
 
 
 
--- a/01-filedir.md
+++ b/01-filedir.md
@@ -323,10 +323,10 @@ If we want to display it, we can give `ls` the `-a` flag:
 $ ls -F -a
 ~~~
 ~~~ {.output}
-./          Desktop/             pizza.cfg
-../         molecules/           solar.pdf
-creatures/  north-pacific-gyre/  writing/
-data/       notes.txt
+./                  creatures/          notes.txt
+../                 data/               pizza.cfg
+.bash_profile       molecules/          solar.pdf
+Desktop/            north-pacific-gyre/ writing/
 ~~~
 
 `-a` stands for "show all";
@@ -336,7 +336,10 @@ As you can see,
 it also displays another special directory that's just called `.`,
 which means "the current working directory".
 It may seem redundant to have a name for it,
-but we'll see some uses for it soon.
+but we'll see some uses for it soon.  
+Finally, we now see a file called `.bash_profile`. This file usually contains settings to customize the shell (terminal). It may also be called `.bashrc` or `.bash_login`. For this lesson material it does not contain any settings.
+
+
 
 > ## Orthogonality {.callout}
 > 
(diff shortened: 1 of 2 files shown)


01-filedir.md
```diff
@@ -336,7 +336,10 @@ As you can see,
 it also displays another special directory that's just called `.`,
 which means "the current working directory".
 It may seem redundant to have a name for it,
-but we'll see some uses for it soon.
+but we'll see some uses for it soon.  
+Finally, we now see a file called `.bash_profile`. This file usually contains settings to customize the shell (terminal). It may also be called `.bashrc` or `.bash_login`. For this lesson material it does not contain any settings.
```
ChristinaLK: language pedantry: can you change "now" to something else, like "Finally, we also see..." or something along those lines?  "now" to me implies immediacy, and this is the last thing being discussed after the command/output is given.

wking:
On Wed, Feb 04, 2015 at 11:40:41PM -0800, Lex Nederbragt wrote:
>     A filesystem/users/nelle/.bash_profile (3)

I like this example, and most of the content in .bash_profile.
However, I'd replace “It may also be called .bashrc or .bash_login”
with “There may also be similar files called .bashrc or .bash_login”,
because the three are not quite synonymous.  From the INVOCATION
section of bash(1):

  When bash is invoked as an interactive login shell, or as a
  non-inter‐ active shell with the --login option, it first reads and
  executes com‐ mands from the file /etc/profile, if that file exists.
  After reading that file, it looks for ~/.bash_profile,
  ~/.bash_login, and ~/.profile, in that order, and reads and executes
  commands from the first one that exists and is readable.  The
  --noprofile option may be used when the shell is started to inhibit
  this behavior.

  …

  When an interactive shell that is not a login shell is started, bash
  reads and executes commands from ~/.bashrc, if that file exists.
  This may be inhibited by using the --norc option.  The --rcfile file
  option will force bash to read and execute commands from file
  instead of ~/.bashrc.

And from the FILES section:

  ~/.bash_profile
    The personal initialization file, executed for login shells
  ~/.bashrc
    The individual per-interactive-shell startup file

So .bash_profile and .bash_login are roughly synonymous, with
.bash_login being a fallback for .bash_profile, but .bashrc is
executed in different (non-login) circumstances.  In fact, since your
example lacks something like:

  [[ -f ~/.bashrc ]] && . ~/.bashrc

nelle's .bashrc won't be sourced at all when invoking a login shell.

Anyhow, new users will certainly not care about all of that, but I'd
still prefer my vague wording to avoid giving the wrong impression.

01-filedir.md
lexnederbragt: Thanks, adjusted accordingly.

lexnederbragt:
Thanks, this I didn't know. Adjusted the text accordingly.

wking:
With 0bef8e0 (Addresses language comments to first commit), this looks
good to me.

01-filedir.md
```diff
@@ -336,7 +336,10 @@ As you can see,
 it also displays another special directory that's just called `.`,
 which means "the current working directory".
 It may seem redundant to have a name for it,
-but we'll see some uses for it soon.
+but we'll see some uses for it soon.  
+Finally, we also see a file called `.bash_profile`. This file usually contains settings to customize the shell (terminal). There may also be similar files called `.bashrc` or `.bash_login`. For this lesson material it does not contain any settings.
```
willingc: Looks good. I suggest adding to: For this lesson material the dotfile, `.bash_profile`, does not contain any workshop settings. The dotfile, `.bash_profile` would be hidden when executing an `ls` command without the `-a` "show all" option.

lexnederbragt: Looks good. I suggest adding to: For this lesson material the dotfile, .bash_profile, does not contain any workshop settings.

I am afraid 'workshop settings' may be confusing...

The dotfile, .bash_profile would be hidden when executing an ls command without the -a "show all" option.

The whole point of adding is to show that adding '-a' to 'ls' shows this file. I feel adding this sentence is redundant and unnecessary.

willingc: I agree with your point on 'workshop settings'.

I do think that it is helpful to 'new' and 'novice' users to explicitly state that without the '-a' option the dotfile will not be displayed. Although an experienced user may find it redundant, many new learners will need the cognitive reinforcement.

ChristinaLK: I agree with @lexnederbragt here, the earlier paragraph explicitly states that `-a` reveals files with a leading dot, so I'm fine with what you had.

PR #67, opened 2015-02-05
fixes #49 (?)

### LINKED ISSUE #49: Issues at 06-find.md
From [check.py](https://github.com/swcarpentry/lesson-template/blob/gh-pages/tools/check.py):

- [ ] Heading at line 383 should be level 2
- [ ] Document contains heading not specified in the template: Conclusion
- [ ] The topic page should not have sub-headings outside of special blocks. If a topic needs sub-headings, it should be broken into multiple topics.

stephenturner:
Still not completely clear on the layout instructions, but I think this helps?

### DIFF
--- a/06-find.md
+++ b/06-find.md
@@ -380,18 +380,18 @@ $ grep FE $(find .. -name '*.pdb')
 > modern programming languages, Python included, have borrowed a lot of
 > ideas from it, and imitation is also the sincerest form of praise.
 
-### Conclusion
-
-The Unix shell is older than most of the people who use it. It has
-survived so long because it is one of the most productive programming
-environments ever created --- maybe even *the* most productive. Its syntax
-may be cryptic, but people who have mastered it can experiment with
-different commands interactively, then use what they have learned to
-automate their work. Graphical user interfaces may be better at the
-first, but the shell is still unbeaten at the second. And as Alfred
-North Whitehead wrote in 1911, "Civilization advances by extending the
-number of important operations which we can perform without thinking
-about them."
+> ## Conclusion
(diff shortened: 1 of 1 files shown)


r-gaia-cs:
@stephenturner Thanks for this pull request. I don't think that the conclusions fit in a call out box. What about move it to `07-conclusions.md`? Or remove it?

ChristinaLK:
This change makes sense to me - it will be formatted the same way as the objectives, which is more appropriate than either the call outs or exercises.

06-find.md
```diff
@@ -380,18 +380,18 @@ $ grep FE $(find .. -name '*.pdb')
...
-survived so long because it is one of the most productive programming
-environments ever created --- maybe even *the* most productive. Its syntax
-may be cryptic, but people who have mastered it can experiment with
-different commands interactively, then use what they have learned to
-automate their work. Graphical user interfaces may be better at the
-first, but the shell is still unbeaten at the second. And as Alfred
-North Whitehead wrote in 1911, "Civilization advances by extending the
-number of important operations which we can perform without thinking
-about them."
+> ## Conclusion
```
r-gaia-cs: For be a call out box this line should be

~~~
> ## Conclusion {.callout}
~~~

ChristinaLK:
@r-gaia-cs , do you think it should be formatted as a callout box?  Using the "objectives" format seems like a better choice to me, in which case it doesn't need the {.callout} tag.  Let me know what you think.

r-gaia-cs:
@ChristinaLK I'm OK with the "objectives" format.

PR #28, opened 2014-12-14
Suggested small fixes for the Unix Shell Loops lesson 

ttimbers:
For my final lesson to become a SWC instructor, I reviewed the unix shell loops lesson and am suggesting two small fixes. The first, is simply to clarify the explanation of the example of why loops are sometimes more useful than wildcards for repeating commands. The second is an addition to the first excercise at the end of the lesson. I think the point of the lesson was to reinforce the difference between using wildcards within for loops versus using $variable_name. Only an example of what happens if you use a wild-card is given. I added an example of using $variable_name so that the learners can clearly examine the differences between these two outputs.

PR #69, opened 2015-02-06
Re: PR #28 - which I accidentally pushed to master instead of gh-pages.

ttimbers:
This commit contains new multiple choice questions tested at a recent SWC workshop (Feb 2015) for 02-create, 03-pipefilter, 04-loop, 05-script and 06-find. This commit also has some suggested changes for the wording of the setup for the first example of why you would want to use loops (creatures directory example). And finally, this commit also has a suggested change to correct the reason for the error output when you try mv to rename files using wild-cards instead of a loop.

### DIFF
--- a/04-loop.md
+++ b/04-loop.md
@@ -19,8 +19,8 @@ Suppose we have several hundred genome data files named `basilisk.dat`, `unicorn
 In this example,
 we'll use the `creatures` directory which only has two example files,
 but the principles can be applied to many many more files at once.
-When new files arrive,
-we'd like to rename the existing ones to `original-basilisk.dat` and `original-unicorn.dat`.
+We would like to modify these files, but also save a version of the original files and rename them 
+as `original-basilisk.dat` and `original-unicorn.dat`.
 We can't use:
 
 ~~~ {.input}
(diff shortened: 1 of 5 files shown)


04-loop.md
```diff
@@ -39,9 +39,10 @@ This wouldn't back up our files, instead we get an error
 mv: target `original-*.dat' is not a directory
 ~~~
 
-This is because there are no files matching the wildcard `original-*.dat`.
-In this case,
-Bash passes the unexpanded wildcard as a parameter to the `mv` command.
+This is a problem arises when mv receives more than two inputs. When this happens, it
```
r-gaia-cs: Wrap "mv" with backtick quotes:

~~~
This is a problem arises when `mv` receives more than two inputs. When this happens, it
~~~

06-find.md
```diff
@@ -393,6 +393,34 @@ North Whitehead wrote in 1911, "Civilization advances by extending the
...
+> With searching comes loss
+> and the presence of absence:
+> "My Thesis" not found.
+>
+> Yesterday it worked
+> Today it is not working
+> Software is like that.
+> ~~~
+>
+> From the above text, contained in the file haiku.txt, which command would result in the
```
r-gaia-cs: Wrap "haiku.txt" with backtick quotes:

~~~
From the above text, contained in the file `haiku.txt`, which command would result in the
~~~

r-gaia-cs:
@ttimbers Awesome work. :sparkles:

Could you make the two changes that I suggested? After you change the files you can

~~~
$ git add -u
$ git commit --amend
$ git push -f origin gh-pages
~~~

to update this pull request. After you update this pull request, please leave a note so we can merge this change. =)

04-loop.md
```diff
@@ -39,9 +39,10 @@ This wouldn't back up our files, instead we get an error
 mv: target `original-*.dat' is not a directory
 ~~~
 
-This is because there are no files matching the wildcard `original-*.dat`.
-In this case,
-Bash passes the unexpanded wildcard as a parameter to the `mv` command.
+This is a problem arises when `mv` receives more than two inputs. When this happens, it
```
ChristinaLK: last little comment - "This problem" not "This is a problem"

ChristinaLK: I'll fix this via github.

Issue #65, opened 2015-02-05
Avoid HTML img tag


r-gaia-cs:
Will be nice to avoid HTML `img` tag. For example, instead of

~~~
<img src="fig/foo.svg" alt="Foo" />
~~~

we can use

~~~
![Foo](fig/foo.svg)
~~~

The list of `img` elements to replace is

~~~
$ grep img *.md      
01-filedir.md:<img src="fig/filesystem.svg" alt="The Filesystem" />
01-filedir.md:<img src="fig/home-directories.svg" alt="Home Directories" />
01-filedir.md:<img src="fig/homedir.svg" alt="Nelle's Home Directory" />
01-filedir.md:<img src="fig/filesystem-challenge.svg" alt="Filesystem for Challenge Questions" />
02-create.md:<img src="fig/nano-screenshot.png" alt="Nano in Action" />
06-find.md:<img src="fig/find-file-tree.svg" alt="File Tree for Find Example" />
~~~

gvwilson:
+1.

PR #70, opened 2015-02-11
Changing <img> tags to markdown

noamross:
This closes issue #65 

Only *.md files are changed, as I'm having trouble getting the build dependencies installed.


wking:
On Wed, Feb 11, 2015 at 10:23:45AM -0800, Noam Ross wrote:
> This closes issue #65 

Looks good to me :).

> Only *.md files are changed, …

As it should be (see [1] and swcarpentry/lesson-template#159 ;).

[1]: http://lists.software-carpentry.org/pipermail/discuss_lists.software-carpentry.org/2015-February/002578.html

ChristinaLK:
Hi @noamross , apparently one of our recently merged changes created conflicts with this.  The only thing you should need to edit/update are the image tags...can you/are you willing to do this?  You just need to change the .md files, we'll update the html.

noamross:
I think this should do the trick.

PR #71, opened 2015-02-12
added wc explanation

amandamiotto:
Trying to add an easier way of remembering wc by explaining the word count origin

### DIFF
--- a/05-script.md
+++ b/05-script.md
@@ -167,6 +167,7 @@ $ wc -l *.pdb | sort -n
 ~~~
 
 because `wc -l` lists the number of lines in the files
+(wc stands for 'word count', adding the -l flag means 'count lines' instead)
 and `sort -n` sorts things numerically.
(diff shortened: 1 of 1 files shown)


wking:
On Thu, Feb 12, 2015 at 02:48:11PM -0800, amandamiotto wrote:
> Trying to add an easier way of remembering wc by explaining the word
> count origin

I'm +1 to mentioning this when we introduce wc.  I haven't checked to
see if this is the first time we mention wc, but if it is then I think
this is great.

ChristinaLK:
`wc` is first mentioned in the pipes lesson, but I think this could stand if you add a "remember" or "recall" at the beginning of your parenthetical statement.

gvwilson:
@amandamiotto I'm +1 to merge after @ChristinaLK's suggested wording change.

amandamiotto:
No problems, added the recall bit

PR #72, opened 2015-02-12
Including .txt extension on output filenames

jmadin:
When running over this lesson during instructor training, I thought that redirecting command output to a file without an extension might be a little confusing for beginners.

### DIFF
--- a/03-pipefilter.md
+++ b/03-pipefilter.md
@@ -103,7 +103,7 @@ but what if there were 6000?
 Our first step toward a solution is to run the command:
 
 ~~~ {.input}
-$ wc -l *.pdb > lengths
+$ wc -l *.pdb > lengths.txt
 ~~~
 
 The `>` tells the shell to **redirect** the command's output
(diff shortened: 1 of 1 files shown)


ChristinaLK:
Hi @jmadin , thanks for the PR!  

I'd like at least one other person to chime in on this, because I'm torn.  I like the .txt addition because it makes the fact that `lengths` is a file more explicit, esp. to beginners.  On the other hand, one of the cool things about using command line is that file extensions are less relevant, so I like demonstrating that you can just dump things into unspecified files.  Other thoughts?  @gdevenyi ?

jduckles:
I'm :-1: I'd prefer a short reminder to point out that there is no extension needed and that extensions are optional, nut not necessary. When I explain redirection I always talk about how the thing on the other side of the `>` is a file, so maybe reinforce that. Files *can* have extensions, but they don't *need* them.

r-gaia-cs:
I'm :+1: because it will make easy for people with Windows machines. If Windows users double click `lengths.txt` they will get Notepad (or something similar) but if they double click `lengths` they will be prompt to select the software to open the file. Since this is the novice lesson I would vote to make it easy for students follow it.

Since I agree with @jduckles that is important that "Files can have extensions, but they don't need them." I would like to request that @jmadin add a note at `instructors.md` saying that instructors could use `lengths` to explain that extensions aren't required for files.

wking:
On Thu, Feb 12, 2015 at 06:51:40PM -0800, Raniere Silva wrote:
> I'm :+1: because it will make easy for people with Windows
> machines. If Windows users double click `lengths.txt` they will get
> Notepad (or something similar) but if they double click `lengths`
> they will be prompt to select the software to open the file. Since
> this is the novice lesson I would vote to make it easy for students
> follow it.

I haven't reviewed the diff and lessons to make sure it hits all the
changes we need, but I agree with Raniere.  POSIX-compliant OSes
generally don't care about extensions (except as a hint to human
users), and OS X (I think) keeps that information in the resource fork
[1,2], but Windows (I think) still trusts the extension completely
when deciding which program to use for opening the file.  Since we
still support folks using Windows, and this change doesn't *break* the
other OSes, I think it's a necessary evil to use the
Windows-accessible-but-otherwise-unconventional .txt extension.

[1]: http://en.wikipedia.org/wiki/OSType
[2]: http://en.wikipedia.org/wiki/Resource_fork

ChristinaLK:
Merge the change and then open an issue about attending a note about needing/not needing file extensions, either in instructors.md or the text of the lessons itself?

wking:
On Fri, Feb 13, 2015 at 11:35:57AM -0800, Christina Koch wrote:
> Merge the change…

Sounds good to me, although I still haven't reviewed the patch for
completeness.  Someone should check for any missing cases before we
merge it.

> …and then open an issue about attending a note about needing/not
> needing file extensions, either in instructors.md or the text of the
> lessons itself?

I'd keep the note in instructors.md.  The shell lesson has enough
going on without distracting students with information about MS
Windows warts ;).

ChristinaLK:
Patch looks good to me!  Thanks for everyone's input.

PR #73, opened 2015-02-12
Naming the > symbol as right arrow.

gshattow:
It makes it easier to refer to it later on and often students appreciate having names for symbols.

### DIFF
--- a/03-pipefilter.md
+++ b/03-pipefilter.md
@@ -106,7 +106,7 @@ Our first step toward a solution is to run the command:
 $ wc -l *.pdb > lengths
 ~~~
 
-The right arrow, `>`, tells the shell to **redirect** the command's output
+The greater than symbol, `>`, tells the shell to **redirect** the command's output
 to a file instead of printing it to the screen.
 The shell will create the file if it doesn't exist,
 or overwrite the contents of that file if it does.
--- a/03-pipefilter.md
+++ b/03-pipefilter.md
@@ -106,7 +106,7 @@ Our first step toward a solution is to run the command:
 $ wc -l *.pdb > lengths
 ~~~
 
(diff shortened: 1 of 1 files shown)


wking:
On Thu, Feb 12, 2015 at 02:58:54PM -0800, gshattow wrote:
>   * Naming the > symbol as right arrow.

+1 for naming, but > is "U+003E GREATER-THAN SIGN" in Unicode.
Right-arrow is something different (→ U+2192 RIGHTWARDS ARROW).  I'd
just say “The greater-than sign, `>`, …”.

gshattow:
Good call. How do I change that?

wking:
On Thu, Feb 12, 2015 at 03:07:03PM -0800, gshattow wrote:
> Good call. How do I change that?

Using GitHub's web UI, you should be able to browse through your
branch [1,2], edit the file [2,3], and make a new commit [3].  That's
going to create two commits though (your original, and then the follow
up fix).  Not a big deal, but for more complicated changes, it can get
hard to review.

Using Git from the command line, you can massage the history to make a
single, revised commit:

  $ git checkout patch-1
  $ emacs 03-pipefilter.md  # or whatever your favorite editor is
	...fix the line...
	$ git commit -a --amend
  $ git push -f ssh://git@github.com/gshattow/shell-novice patch-1

In both cases, GitHub will update this pull request automatically
after you've updated your patch-1 branch.

[1]: https://github.com/gshattow/shell-novice/tree/patch-1
[2]: https://github.com/gshattow/shell-novice/blob/patch-1/03-pipefilter.md
[3]: https://help.github.com/articles/editing-files-in-your-repository/

ChristinaLK:
Looks good!  I'll merge this in.

PR #21, opened 2014-11-21
 Correcting images to match text description in 01-filedir.md file #873 

shlake:
Per email and issue:

 Correcting images to match text description in 01-filedir.md file #873.

I'll close this issue now.
https://github.com/swcarpentry/bc/pull/873

PR #74, opened 2015-02-12
Make text dir paths resemble the images of those dirs

RobTheOceanographer:
This is part of the instructor training class in Melbourne.
I have made the text paths represent the image representations of those paths so that there is more consistency and flow to this part of the lesson.

### DIFF
--- a/01-filedir.md
+++ b/01-filedir.md
@@ -109,8 +109,8 @@ because its name begins with `/`.
 
 Underneath `/users`,
 we find one directory for each user with an account on this machine.
-The Mummy's files are stored in `/users/imhotep`,
-Wolfman's in `/users/larry`,
+The Mummy's files are stored in `/users/dru`,
+Wolfman's in `/users/gorgon`,
 and ours in `/users/nelle`,
 which is why `nelle` is the last part of the directory's name.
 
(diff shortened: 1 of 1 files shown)


ChristinaLK:
Haha, this changes the text, this old PR #21 (unmerged) changes the image. Obviously a long-standing consistency problem!  I feel like the text should stay the way it is (b/c in the  next lesson, learners create `thesis` using `mkdir`), but I'll let others chime in on this.

RobTheOceanographer:
Hi Christina, I don't mind which one is used but i think it would be good to merge one of these PRs so there is consistency in the lesson - maybe the image PR #21 change because, as you pointed out, the learners are asked to create the thesis when they get to mkdir.

ChristinaLK:
Would you be able to copy/save that image and resubmit it as a new PR?  Don't feel obligated! - if you don't want to, I'll file it as an issue and someone will get round to it sooner or later.

gvwilson:
@ChristinaLK your call whether to take this one or the older one that changes the image.

ChristinaLK:
Closed -> changing imgs instead from PR #21, resubmitted in PR #105

PR #76, opened 2015-02-12
Update 04-loop.md

DangarStu:
I feel this is an opportunity to reinforce how the shell goes through a pre-processing phase before calling any executables.

### DIFF
--- a/04-loop.md
+++ b/04-loop.md
@@ -144,9 +144,10 @@ prints:
 hello there
 ~~~
 
-In this case,
-since the shell expands `$filename` to be the name of a file,
-`echo $filename` just prints the name of the file.
+The shell expands all special tokens before calling program specified. In this case,
+the shell expands `$filename` to be the name of a file before calling `echo`, therefore
+printing the name of the file rather than the name of the variable specified in your script.
+In fact `echo` has no way of knowing the variable used, only its contents!
 Note that we can't write this as:
 
 ~~~ {.input}


wking:
On Thu, Feb 12, 2015 at 03:34:08PM -0800, DangarStu wrote:
> I feel this is an opportunity to reinforce how the shell goes
> through a pre-processing phase before calling any executables.

We need to walk the line between technical accuracy/detail and
accessibility.  For example, you're introducing "token" here.  Still,
I think explaining that this happens in the shell before calling the
command is worth the trouble.  How about “The shell expands variables
before…”?

DangarStu:
Sounds good to me. I agree that you can go too far and make it more
confusing for the student. :)

*PARTING LINES*: "A novel of human foibles and near redemption." – *Kirkus
Reviews*

[image: Support independent publishing: Buy this book on Lulu.]
<http://www.lulu.com/spotlight/dangarstu>

http://dangarstu.com

04-loop.md
```diff
@@ -144,9 +144,10 @@ prints:
 hello there
 ~~~
 
-In this case,
-since the shell expands `$filename` to be the name of a file,
-`echo $filename` just prints the name of the file.
+The shell expands all special tokens before calling program specified. In this case,
+the shell expands `$filename` to be the name of a file before calling `echo`, therefore
+printing the name of the file rather than the name of the variable specified in your script.
+In fact `echo` has no way of knowing the variable used, only its contents!
```
gvwilson: The shell expands variable references like `$filename` *before* calling commands like `echo`,
so `echo` prints the name of the file rather than the name of the variable ---
in fact, `echo` has no way of knowing a variable was used.

gdevenyi:
@DangarStu If you can remove the mentions of "tokens" which we don't describe or cover anywhere I'm happy with the other changes expanding upon what "echo" sees vs how you describe the loop.

gdevenyi:
@DangarStu  If you can remove the mentions of "tokens" which we don't describe or cover anywhere I'm happy with the other changes expanding upon what "echo" sees vs how you describe the loop.

PR #78, opened 2015-02-13
Fixing special-blockquotes

sritchie73:
I noticed that many of the challenge blocks on the rendered pages were not showing up correctly. After forking the repository and fixing a few blocks, it became apparent this was primarily due to the HTML files being out of date. I fixed up the remaining blocks, and ran `make preview` so that the HTML reflects what's in the markdown.


ChristinaLK:
Hi @sritchie73 !  This may be addressed by a previous (still unmerged) PR, but if not, we will merge this in soon.

PR #64, opened 2015-02-05
Update due changes on lessons template

r-gaia-cs:
**This include a merge from `core` branch of https://github.com/swcarpentry/lesson-template/.**

## Commits to review

- [ ] 9808e9c
- [ ] d9e4899

PR #79, opened 2015-02-14
Small fix to CONTRIBUTING

r-gaia-cs:
Address https://github.com/swcarpentry/shell-novice/pull/64#issuecomment-74338121

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,14 +36,9 @@ you are agreeing that Software Carpentry may redistribute your work under
 
 2.  For a list of helpful commands run `make` in this directory.
 
-3.  The [list of maintainers][swc-maintainers]
-    on the [Software Carpentry website][swc-website]
-    lists the people currently responsible for managing this repository.
-    Feel free to contact them if you have any questions or languishing pull requests.
-
 [christina]: https://github.com/ChristinaLK
 [gabriel]: https://github.com/gdevenyi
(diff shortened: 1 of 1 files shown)


wking:
On Fri, Feb 13, 2015 at 04:52:42PM -0800, Raniere Silva wrote:
> Assigned #79 to @wking.

If we want this change, I'd also recommend dropping the two refrences
we no longer use:

  [swc-maintainers]: http://software-carpentry.org/lessons.html#maintainers
  [swc-website]: http://software-carpentry.org

Other than that, this change looks good to me.  However:

* I'm not a maintainer for this repository, so it's not my call.
* I'm not sure if the upstream link to swc-maintainers [1] is a
  suggestion or an official policy.  Ideally (in my mind), the
  lesson-template example would list *its* maintainers in its
  CONTRIBUTING.md, and we'd adjust them in downstream lessons.
  Maintaining maintainer information in two places (local
  CONTRIBUTING.md, which I like, and a central listing, which doesn't
  make sense to me) isn't very DRY [2].

[1]: https://github.com/swcarpentry/lesson-template/commit/7547e803f292b48cca57185961c70b73e163a0e6
[2]: http://en.wikipedia.org/wiki/Don%27t_repeat_yourself

ChristinaLK:
Looks good to me - I'd like to bring back the "welcome, thank you for your interest", but I can do that in my own PR.

PR #80, opened 2015-02-15
Added small box with command.

areejalsheikh:
A small part was missing the command so I added it up.

### DIFF
--- a/02-create.md
+++ b/02-create.md
@@ -319,9 +319,10 @@ quotes.txt   thesis/quotations.txt
 To prove that we made a copy,
 let's delete the `quotes.txt` file in the current directory
 and then run that same `ls` again.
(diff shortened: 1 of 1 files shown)


ChristinaLK:
Hi @areejalsheikh , that's a great catch!  My very picky comment - can you put both of the "input" commands in the same box (as is done previously when there are two input commands in a row), for consistency?  If you commit that change to this same branch locally and push to your fork, the PR should update automatically and then I can merge it in.

areejalsheikh:
Done. Thanks for the comment. Though, though I have done this on the R Markdown file instead?

ChristinaLK:
You made the change in the right place - this lesson uses plain markdown (md) files as webpage "source", so that's what should be changed.

PR #81, opened 2015-02-15
Update HPC discussion

TomKellyGenetics:
Access to servers and HPC infrastructure was a powerful way the command-line improved my own research. As a biology researcher it was key motivating factor to learn the shell. I think emphasis on the broad range of applications of the shell, esp. HPC would appeal to many novice learners.

### DIFF
--- a/00-intro.md
+++ b/00-intro.md
@@ -118,6 +118,6 @@ On the other hand,
 the shell allows us to combine existing tools in powerful ways with only a few keystrokes
 and to set up pipelines to handle large volumes of data automatically.
 In addition,
-the command line is often the easiest way to interact with remote machines.
-As clusters and cloud computing become more popular for scientific data crunching,
-being able to drive them is becoming a necessary skill.
+the command line is often the easiest way to interact with remote machines and supercomputers. 
+Familiarity with the shell is near essential to run a variety of specialised tools and resources including high-peformance computing systems. As clusters and cloud computing become more popular for scientific data crunching,
+being able to drive them is becoming a necessary skill. We can build on the command-line skills covered here to tackle a wide range of scientific questions and computational challenges.


gvwilson:
Let's ask @adina what she thinks...

adina:
@gvwilson @TomKellyGenetics  Totally agree that shell on HPC or on the cloud is super useful!   I think the problem here is that it is hard to come up with a standard teaching module that is widely applicable and does not require everyone to sign up (and pay).  As an instructor, I usually discuss HPC and Cloud generally but don't do any specific examples within it.  I'm always open to suggestions though!

gvwilson:
Should a couple of sentences be put right up front in the `index.md` 
file, or in the "Legend" section of `instructors.md` ?

adina:
I think changes in 00-intro.md are good and useful.  I think adding it to the Legends section makes sense as well.  @TomKellyGenetics want to suggest some changes?

TomKellyGenetics:
I intended this simply to appeal to a broader range of new users (particularly scientists from other disciplines) as motivation to use learn to use the shell. New to SWcarpentry and not familiar with legend or index files. Good to see it's getting considered though. I'm busy running a SWC workshop in Christchurch at the moment. May revisit this topic later, I'll leave it in our capable hands until then.

ChristinaLK:
@TomKellyGenetics , I think this is a useful point to make in the intro and I'll make a line note for minor wording changes.

ChristinaLK:
Actually, just kidding on the changes, it looks good the way it is!  Did anyone (@adina?) want to add anything now?  Otherwise I'll just merge this in.

TomKellyGenetics:
I'm happy with the changes I've submitted as they are but I'm open to suggestions if anyone has more ideas. I think the take home message here is that this shell lesson isn't a stand-alone course for basic computing tools. It's the beginning of an exciting journey which culminates in using a wide range of powerful tools researchers otherwise wouldn't have access to (incl. HPC).

TomKellyGenetics:
@gvwilson looked into the index.md and instructors.md files. I agree, it would make sense to make corresponding changes in these files, esp index.md (although I like it's current clear and concise nature). Most instructors are likely aware of many applications of the shell not covered in this material (incl. HPC). Is it necessary to remind them of it? Would a general "third answer: enables use of many domain-specific tools and compute resources" line in Legend of instructors.md be appropriate or a specific example (e.g., HPC). It's worth noting that HPC infrastructure in Aus/NZ (NECTAR, NeSI) are recommending SWcarpentry to their new users too.

These changes are getting out of hand so I've created separate pull requests for the other files:
https://github.com/swcarpentry/shell-novice/pull/96
https://github.com/swcarpentry/shell-novice/pull/97

PR #82, opened 2015-02-15
Add Home Dir

TomKellyGenetics:
Added use of  '~' for home directory


TomKellyGenetics:
May not be necessary in hindsight, the "Another Useful Abbreviation" section in page-02 covers this well,

ChristinaLK:
I can see why this would be a logical place to explain the `~` shortcut, but since it's already explained later, I'll close this for now.

PR #86, opened 2015-02-17
Change a cd to make it 'correct' for those that follow all steps in 03-pipefilter and 04-loop

lexnederbragt:
When strictly following the flow, Nelle is still in the `molecules` folder at this point, and needs to do `cd ../north-pacific-gyre/2012-07-03` instead.

### DIFF
--- a/03-pipefilter.md
+++ b/03-pipefilter.md
@@ -289,10 +289,10 @@ so that you and other people can put those programs into pipes to multiply their
 
 Nelle has run her samples through the assay machines
 and created 1520 files in the `north-pacific-gyre/2012-07-03` directory described earlier.
-As a quick sanity check, she types:
+As a quick sanity check, starting from her home directory, Nelle types:
 
 ~~~ {.input}
-$ cd ../north-pacific-gyre/2012-07-03
+$ cd north-pacific-gyre/2012-07-03
 $ wc -l *.txt
 ~~~
 
--- a/03-pipefilter.md
+++ b/03-pipefilter.md
@@ -292,7 +292,7 @@ and created 1520 files in the `north-pacific-gyre/2012-07-03` directory describe
 As a quick sanity check, she types:
 
 ~~~ {.input}
-$ cd north-pacific-gyre/2012-07-03
+$ cd ../north-pacific-gyre/2012-07-03
 $ wc -l *.txt
 ~~~
 
--- a/04-loop.md
+++ b/04-loop.md
@@ -277,7 +277,7 @@ Her first step is to make sure that she can select the right files --- remember,
 these are ones whose names end in 'A' or 'B', rather than 'Z':
 
 ~~~ {.input}
-$ cd north-pacific-gyre/2012-07-03
+$ cd ../north-pacific-gyre/2012-07-03
 $ for datafile in *[AB].txt
 > do
 >     echo $datafile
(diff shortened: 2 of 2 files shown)


ChristinaLK:
I think I'd rather see a note added before the input box along the lines of "starting from her home directory, Nelle types..." and then have a `cd` without the `..`.  That's mostly because `cd ../north-pacific-gyre/` doesn't look as nice to me, so I'd welcome other yays/nays!

wking:
On Fri, Feb 20, 2015 at 08:14:29PM -0800, Christina Koch wrote:
> I think I'd rather see a note added before the input box along the
> lines of "starting from her home directory, Nelle types..." and then
> have a `cd` without the `..`.  That's mostly because `cd
> ../north-pacific-gyre/` doesn't look as nice to me, so I'd welcome
> other yays/nays!

Maybe ‘cd ~/north-pacific-gyre’?  Running two back-to-back ‘cd’
commands doesn't make sense to me.

Personally, I prefer @lexnederbragt's relative links, since many
(most?) users won't have the example filesystem unpacked to *their*
home directory.  The relative links allow to to avoid figuring out the
right absolute path more than once.  While live-coding the lesson,
it's probably a good idea to use a prompt that shows your full path.
In a POSIX shell you can set PS1 [1] to do that.  Something like:

  $ export PS1='\[\033[01;32m\]\u@\h\[\033[01;34m\] \w \$\[\033[00m\] '

should work (and it's Gentoo's default for non-root users).  See the
“PROMPTING” section in bash(1) for other ideas.  To get the real nelle
experience, you could do something like:

  $ export HOME=~/src/swc/shell-novice/filesystem/users/nelle
  $ export PS1='\[\033[01;32m\]nelle@\h\[\033[01;34m\] \w \$\[\033[00m\] '

It's probably also a good idea to make liberal use of ‘tree -d’ or
similar while students are still getting used to mentally tracking
their position in the filesystem.

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_05_03

lexnederbragt:
As for the prompt, at the most recent workshop I taught, the students asked for a shorter one to make the command written stand out more clearly :-)

wking:
On Fri, Feb 27, 2015 at 02:15:41AM -0800, Lex Nederbragt wrote:
> As for the prompt, at the most recent workshop I taught, the
> students asked for a shorter one to make the command written stand
> out more clearly :-)

But the command is white-on-black (or black-on-white, I guess,
depending on your terminal color preference), and the prompt is all
colored.  In any case, ‘tree -d’ and ‘pwd’ should be uncontroversial
;).

ChristinaLK:
Hey @gdevenyi, can you weigh in on this?

gdevenyi:
I agree that the learners are unlikely to have the example filesystem unpacked into their own home directory, as such "~" won't work.

While I understand the want to allow learners to to verbatim type the commands in the lessons, I believe that providing exact commands in all cases facilitates copy+paste following, which reduces understanding.

I would prefer prefacing the discussion with the "from Nelle's home directory"

Should this be a close of the PR and opening of a new one or just modify the existing?

lexnederbragt:
+1 for @gdevenyi's take. Adjusted PR accordingly. (should I change the PR title?)

ChristinaLK:
This looks good - what @gdevenyi said was what I was trying to suggest.  :)  Merging now.

PR #87, opened 2015-02-20
Update 06-find.md

jaredwinton:
An extra multiple choice question that highlights what I think would be a common next thing to try for a lot of people, searching for something with a space in the pattern. It might also be worth putting in a callout explaining how to use ' 's to search for a pattern that includes a space. But I'm not too sure what the convention is for callouts for questions.

### DIFF
--- a/06-find.md
+++ b/06-find.md
@@ -456,3 +456,42 @@ $ grep FE $(find .. -name '*.pdb')
 > of the four sisters is mentioned?  Hint: one solution might employ
 > the commands `grep` and `wc` and a `|`, while another might utilize
 > `grep` options.
+>
+> ## Grep, being more specific with your search {.challenge}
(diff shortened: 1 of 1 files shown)


06-find.md
```diff
@@ -456,3 +456,42 @@ $ grep FE $(find .. -name '*.pdb')
...
+> ~~~
+>
+> d) Prints all lines that contain the pattern "the" .
+>
+> ~~~ {.input}
+> Is not the true Tao, until
+> and the presence of absence:
+> ~~~
+>
+> e)  The second and third argument are treated as files, an error is printed because the file `presence` does not exist, followed by all lines that contain the pattern "the", prefaced with the file name
```
ChristinaLK: I would take out "the second and third argument are treated as files" - that's something to be discussed or for people to try to figure out on their own, once they know the right answer.

PR #90, opened 2015-03-01
`goostats` and `goodiff` need to be run with `bash`

lexnederbragt:
In 04-loop, `goostat` is run with `bash` before it and IRL that is needed as well. This PR adds `bash` to these scripts in 05-script

### DIFF
--- a/05-script.md
+++ b/05-script.md
@@ -260,8 +260,8 @@ $ history | tail -4 > redo-figure-3.sh
 The file `redo-figure-3.sh` now contains:
 
 ~~~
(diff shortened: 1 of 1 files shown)


wking:
On Sun, Mar 01, 2015 at 11:32:54AM -0800, Lex Nederbragt wrote:
> In 04-loop, `goostat` is run with `bash` before it and IRL that is
> needed as well. This PR adds `bash` to these scripts in 05-script

Looks good to me :).

BernhardKonrad:
Seconded, thank you.

PR #94, opened 2015-03-03
Fixing complaints about 00-intro.md

### LINKED ISSUE #54: Issues at 00-intro.md
From [check.py](https://github.com/swcarpentry/lesson-template/blob/gh-pages/tools/check.py):

- [ ] Heading at line 55 should be level 2
- [ ] Document contains heading not specified in the template: What and Why
- [ ] The topic page should not have sub-headings outside of special blocks. If a topic needs sub-headings, it should be broken into multiple topics.

gvwilson:
Fixes #54.

### DIFF
--- a/00-intro.md
+++ b/00-intro.md
@@ -4,55 +4,10 @@ title: The Unix Shell
 subtitle: Introducing the Shell
 minutes: 5
 ---
-> ## Learning Objectives
+> ## Learning Objectives {.objectives}
 >
 > *   Explain how the shell relates to the keyboard, the screen, the operating system, and users' programs.
 > *   Explain when and why command-line interfaces should be used instead of graphical interfaces.
-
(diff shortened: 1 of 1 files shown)


ChristinaLK:
Request: move the paragraph about Nelle back to 00-intro.md, but after the "technical intro".  Moving it all to `instructors.md` means that when the following lessons talk about "nelle's pipeline," someone reading the lesson has no context for what's happening there.  We can even mirror the format of the following lessons with a level-2 heading "Nelle's Pipeline: An Applied Example" or similar.  ;)  

I could pare down the Nelle verbiage in a separate PR if we'd like the intro to be short + snappy.

gvwilson:
Good idea: change made.  I'll file a separate PR against the template to allow for a running example.

ChristinaLK:
:+1: Excellent.

PR #95, opened 2015-03-03
Fixing (most) problems in 04-loop.md

gvwilson:
Addresses most of #28.

"Nelle's Pipeline" is a good use case for a level-2 heading inside a lesson...

### DIFF
--- a/04-loop.md
+++ b/04-loop.md
@@ -4,7 +4,7 @@ title: The Unix Shell
 subtitle: Loops
 minutes: 15
 ---
-> ## Learning Objectives
+> ## Learning Objectives {.objectives}
 >
 > *   Write a loop that applies one or more commands separately to each file in a set of files.
 > *   Trace the values taken on by a loop variable during execution of the loop.
@@ -268,7 +268,7 @@ mv unicorn.dat original-unicorn.dat
 > isn't foolproof, but it's a handy way to see what's going to happen when
 > you're still learning how loops work.
 
-### Nelle's Pipeline: Processing Files
+## Nelle's Pipeline: Processing Files
 
 Nelle is now ready to process her data files.
 Since she's still learning how to use the shell,
@@ -498,7 +498,7 @@ so she decides to get some coffee and catch up on her reading.
 > done
 > ~~~
 
-> ## FIXME {.challenge}
+> ## Command-Line Expressions {.challenge}
 >
 > The `expr` does simple arithmetic using command-line parameters:
 > 
@@ -521,7 +521,7 @@ so she decides to get some coffee and catch up on her reading.
 > done
 > ~~~
 
-> ## FIXME {.challenge}
+> ## Explain This Loop {.challenge}
 > 
 > Describe in words what the following loop does.
 > 
--- a/04-loop.md
+++ b/04-loop.md
@@ -394,7 +394,7 @@ so she decides to get some coffee and catch up on her reading.
 > then she can re-run `goostats` on `NENE01729B.txt` simply by typing
 > `!458`.
 
-> ## Variables in loops {.challenge}
+> ## Variables in Loops {.challenge}
 > 
 > Suppose that `ls` initially displays:
 > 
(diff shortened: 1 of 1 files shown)


wking:
On Tue, Mar 03, 2015 at 07:34:52AM -0800, Greg Wilson wrote:
>   * Fixing (most) problems in 04-loop.md

This replaces some of the challenge titles, which I'd tried to address
in #33.  I've never been clear on why #33 was closed [1], but I prefer
my:

  Nested loop reading comprehension

and:

  Leading variable reading comprehension

to this PR's:

  Command-Line Expressions

and:

  Explain This Loop

[1]: https://github.com/swcarpentry/shell-novice/pull/33#issuecomment-72404111

ChristinaLK:
Is there a protocol for level 2 vs level 1 (or 3) headings?  

@wking, including "reading comprehensions" in the title of a challenge seems a bit daunting to me, mostly because I think you mean it quite literally (as in, can you read + comprehend this), but I would conflate it w/ more technical ideas about "reading" (files) and comprehensions (lists in python).  Not sure a novice would have those associations, but I think short/simple words are best for titles in this situation.  

@gvwilson My pedantic nature would suggest including "nested loops" or "multiple loops" in the title of the "command line expressions" challenge (because that seems to be the real focus of the challenge).  Thoughts?

Otherwise looks good.

wking:
On Tue, Mar 03, 2015 at 07:14:46PM -0800, Christina Koch wrote:
> Is there a protocol for level 2 vs level 1 (or 3) headings?  

The lesson-template docs [1] use level 2 headings for blockquotes
(which we use for learning objectives, challenges, and callouts).
They also say: “There are no sub-headings inside a topic other than
the ones shown. (If a topic needs sub-headings, it should be broken
into two or more topics.)”

> @wking, including "reading comprehensions" in the title of a
> challenge seems a bit daunting to me, mostly because I think you
> mean it quite literally (as in, can you read + comprehend this), …

Yeah.

> … but I would conflate it w/ more technical ideas about "reading"
> (files) and comprehensions (lists in python).

Yikes ;).  Maybe I just took too many reading comprehension tests
growing up ;).

> Not sure a novice would have those associations, but I think
> short/simple words are best for titles in this situation.

We can probably find something that works for everyone ;).  How about
we just drop the troublesome phrase and have:

  Nested loops

and:

  Leading variables

?

As I pointed out in #33, I'd rather just drop the last challenge,
because I don't think it belongs in the loop section.  If we keep it,
it would be nice to at least have some notes in instructors.md
explaining what students are supposed to be learning, and why the
challenge is lurking in the loop section.

[1]: https://github.com/swcarpentry/lesson-template/blob/966a0322695852676884f8c094b83d31cc4c728c/LAYOUT.md#topics

gvwilson:
Changes made based on feedback.

wking:
On Wed, Mar 04, 2015 at 05:07:15AM -0800, Greg Wilson wrote:
> Changes made based on feedback.

If you want to highlight `expr`, maybe:

  Nested Loops and Command-Line Expressions

could be:

  Nested Loops and `expr`

bash(1) talks about expressions for ((…)) (arithmetic), [[…]]
(conditionals), {x..y[..incr]} (sequences), ….  Wikipedia's definition
for “expression” matches pretty much everything listed in bash(1)'s
SHELL GRAMMAR section (simple commands, pipelines, lists, compound
commands, …).  I'd rather be explicit and just mention `expr`
directly.

I'm also still not clear on the purpose of the last challenge [2].
Should I open a new PR for that discussion?

[1]: http://en.wikipedia.org/wiki/Expression_%28computer_science%29
[2]: https://github.com/swcarpentry/shell-novice/pull/95#issuecomment-77091277

ChristinaLK:
@wking I think a new PR (or issue) about last challenge is appropriate.

wking:
On Wed, Mar 04, 2015 at 08:27:17PM -0800, Christina Koch wrote:
> I think a new PR (or issue) about last challenge is appropriate.  

Done in #99.

PR #98, opened 2015-03-04
Add in a diagram showing redirects and pipes

jpallen:


jpallen:
This is a resubmission of #5

03-pipefilter.md
```diff
@@ -254,6 +254,8 @@ we get three processes with data flowing from the files,
 through `wc` to `sort`,
 and from `sort` through `head` to the screen.
 
+<img src="img/redirects-and-pipes.svg" alt="Redirects and Pipes" />
```
ChristinaLK: For the .md files, we're using markdown image syntax: 
Redirects and Pipes

ChristinaLK: Also, based on previous use, the image should probably be in figs/ not img/.

ChristinaLK:
I love the colors.  :)  This is going to be really helpful in this lesson!  

In the original, I liked having the little text blocks under the arrows, showing what was being passed/printed at each step.  Would it be a lot of trouble to add those back in?  If that ends up making a huge image, you could split it, with the normal output/simple redirect after the "Here's what happens..." paragraph and then the full piped expression after "If we run..."  

If that's all too much bother then I'll just merge the image as is and we can tweak later.

ChristinaLK:
Tagging @jpallen to make sure he sees comments - can you make the noted changes?

PR #104, opened 2015-03-13
fixing path to nano-screenshot

ChristinaLK:


wking:
On Thu, Mar 12, 2015 at 07:20:08PM -0700, Christina Koch wrote:
>   * fixing path to nano-screenshot

Looks good to me.

PR #106, opened 2015-03-13
resizing homedir img

ChristinaLK:


ChristinaLK:
quick fix of background img size.

ChristinaLK:
I only changed the .svg, do I need to also have an .odg copy or can I delete?

wking:
On Fri, Mar 13, 2015 at 07:27:48AM -0700, Christina Koch wrote:
> I only changed the .svg, do I need to also have an .odg copy or can
> I delete?

I don't mind.  It looks like you used Inkscape to edit the SVG, which
is fine with me, and in that case you should drop the (now
out-of-date) ODG.  If you want folks to use LibreOffice to edit the
graphic, you should keep the ODG, make your changes there, and just
export from LibreOffice to get the SVG.  Whatever route you take, it's
probably best to be consistent amoung the existing fig/ entries.

PR #107, opened 2015-03-17
Add motivation.md file

anelda:


wking:
I don't think you need the <section class="slide"> bits.  For example,
see [1].

* “unnecesarily” → “unnecessarily”
* “not to be reproducible” → “to not be reproducible”

I'd drop the semi-duplicate “manual … manipulation” wording with
something like:

  - Running the same workflow on several samples can be unnecessarily
    labour intensive
  - Manual manipulation of data files:
      - is often not captured in documentation
      - is hard to reproduce
      - is hard to troubleshoot, review, or improve

Other than those minor tweaks, this looks good to me.

[1]: https://raw.githubusercontent.com/swcarpentry/lesson-template/gh-pages/motivation.md

anelda:
Thanks @wking , I've incorporated your suggestions

wking:
On Tue, Mar 17, 2015 at 12:41:08PM -0700, Anelda van der Walt wrote:
> I've incorporated your suggestions

Thanks, and it's looking good.

Reviewing through 8edd4728, the only remaining thing I see is the
double period in “Let's see how the shell can be used in research data
analysis..”  I expect you wanted an ellipsis (…).  For previous
comments on a Unicode … (U+2026 HORIZONTAL ELLIPSIS), vs. triple
period, see swcarpentry/bc#89.  I'm not sure we have an official
policy on ASCII vs. Unicode though.  In any case, you probably don't
want ‘..’ ;).

anelda:
OK, I took the easy way out and added a third period. I have SO much to learn ;) Will this work?

wking:
On Tue, Mar 17, 2015 at 01:30:35PM -0700, Anelda van der Walt wrote:
> OK, I took the easy way out and added a third period. I have SO much
> to learn ;) Will this work?

Looks good to me :).

gdevenyi:
Looks good, merging!

PR #77, opened 2015-02-13
removing backup dirs, renaming filesystem to data

ChristinaLK:
See also #62.  This at least cleans up the directory.

PR #110, opened 2015-03-19
Filesystem revert

ChristinaLK:
Reverts some of the changes from #77 in order to re-match lesson text.  Also adds some blank directories to match the lesson diagrams.  Could add files to those directories upon request, but I wanted to keep things simple.


gdevenyi:
Thanks for reverting this, sorry you had to do all this work twice!

ChristinaLK:
I wouldn't call a `mv` a lot of work.  :wink:

PR #117, opened 2015-04-08
Update readme instructor

ChristinaLK:
Hi all - with the removal of `master`, I deleted the note about it on README.md and added some links to the live site and instructors.html.  

Also added some notes to instructors.md.  I can split this into two pull requests if the discussion on my changes there prove lengthy.  @gvwilson, would especially like your feedback on whether this is the kind of material you'd like to see in the instructors file.

### DIFF
--- a/instructors.md
+++ b/instructors.md
@@ -74,13 +74,20 @@ as long as learners using Windows do not run into roadblocks such as:
 
 *   Time estimates:
     *   @gvwilson: 3 hours
-
+	
+*	How to use the materials in the shell-novice repository (or, lesson planning)
+	* Use the `filesystem` directory for in-workshop exercises and live coding examples.  
+	The easiest way to get can get these files is by cloning the entire shell-novice directory.  Alternatives include creating a zipped copy of the `filesystem` directory and making that available to learners.  
+	* Website: Can give links to learners so they can follow along and see exercises (particularly if you're following the lesson content without many changes).  Make sure to point to website as a post-workshop reference.  
+	* Content: Unless you have a truly generous amount of time (4+ hours), it is likely that you will not cover ALL the material in this lesson in a single half-day session.  Plan ahead on what you might skip, what you really want to emphasize, etc.  
+	* Exercises: Think in advance about how you might want to handle exercises during the lesson.  Do you want everyone to try it and then you show the solution?  Have a learner show the solution?  Have groups each do a different exercise and present their solutions.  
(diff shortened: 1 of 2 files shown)


instructors.md
```diff
@@ -74,13 +74,20 @@ as long as learners using Windows do not run into roadblocks such as:
 
 *   Time estimates:
     *   @gvwilson: 3 hours
-
+	
+*	How to use the materials in the shell-novice repository (or, lesson planning)
+	* Use the `filesystem` directory for in-workshop exercises and live coding examples.  
+	The easiest way to get can get these files is by cloning the entire shell-novice directory.  Alternatives include creating a zipped copy of the `filesystem` directory and making that available to learners.
```
wking: No need to create anything, since GitHub provides that zip for you under the [Download ZIP][1] button.  Although I personally think it's worth creating a new repository for a workshop that collects all the data learners will need in one repository (see efran/2015-04-09-UW#16, although we didn't end up using that approach for the UW workshop).

[1]: https://github.com/swcarpentry/shell-novice/archive/gh-pages.zip

instructors.md
```diff
@@ -74,13 +74,20 @@ as long as learners using Windows do not run into roadblocks such as:
 
 *   Time estimates:
     *   @gvwilson: 3 hours
-
+	
+*	How to use the materials in the shell-novice repository (or, lesson planning)
+	* Use the `filesystem` directory for in-workshop exercises and live coding examples.  
+	The easiest way to get can get these files is by cloning the entire shell-novice directory.  Alternatives include creating a zipped copy of the `filesystem` directory and making that available to learners.  
+	* Website: Can give links to learners so they can follow along and see exercises (particularly if you're following the lesson content without many changes).  Make sure to point to website as a post-workshop reference.
```
wking: I think Greg says explicitly to *not* show learners the HTML version of the lessons until after the workshop, but this (show them the HTML version of the excercises) is the approach the Python room is taking at the current UW workshop.

ChristinaLK: I personally don't like having the lessons up while I teach, but I have consistent anecdotal evidence indicating that participants like having them up as a way of being able to catch up and/or figure out where we are in the lesson.  So I have mixed feelings but wanted to be fair to the input of other people.

wking: On Thu, Apr 09, 2015 at 01:12:49PM -0700, Christina Koch wrote:
> > + * Website: Can give links to learners so they can follow along
> >     and see exercises (particularly if you're following the lesson
> >     content without many changes).  Make sure to point to website
> >     as a post-workshop reference.
> 
> … So I have mixed feelings but wanted to be fair to the input of
> other people.

That sounds reasonable to me.  If we want to put forward both
positions and not make a firm suggestion, I think we should be laying
out the arguments in more detail.  I'm not sure if Greg's suggestions
are echoed somewhere in the repository's notes, but the “don't show
these live” recommendation is in the “What the Lesson Notes Are For”
of [1].

[1]: http://software-carpentry.org/blog/2015/03/teaching-tips.html

PR #118, opened 2015-04-14
[01-filedir] Claim six sub-directories in /users/nelle

dwinston:
I count six, not seven, sub-directories (in the lesson text as well as in the repo filesystem).


ChristinaLK:
you are correct!  thanks for the fix.  :)

PR #125, opened 2015-04-20
Remove touch.

ocefpaf:
I think that teaching `touch` to create files is unnecessary.  Every time I tried to explained it to others they ask back:  "Why can't I just type `nano draft.txt` instead."

### DIFF
--- a/02-create.md
+++ b/02-create.md
@@ -61,26 +61,6 @@ However, there's nothing in it yet:
 $ ls -F thesis
 ~~~
 
-> ## Touching Files {.callout}
->
-> One of the simplest ways to create an empty file is via the `touch` command.
-> Change the working directory to `thesis` using `cd`,
-> then touch an empty file called `draft.txt`:
->
-> ~~~ {.input}
-> $ cd thesis
-> $ touch draft.txt
-> ~~~
->
-> If we check the directory contents now,
->
-> ~~~ {.input}
-> $ ls -F .
-> ~~~
-> ~~~ {.output}
-> draft.txt
-> ~~~
-
 Let's change our working directory to `thesis` using `cd`,
 then run a text editor called Nano to create a file called `draft.txt`:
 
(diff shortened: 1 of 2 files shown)


iglpdc:
Hmm, I agree only in part. `touch` is a really fast, uncomplicated way to create files in this part of the lesson, when you just create and delete the file a couple of times. I'd keep it to save time and because it's likely that the instructors uses it at some point of the workshop.

ocefpaf:
> really fast, uncomplicated way to create files

OK, I agree with that.

> when you just create and delete the file a couple of times

Here is where I disagree.  I don't think an instructor should be creating and deleting files with `touch` just to teach creation/deletion.  Teaching "real" file creation and deletion is slow, but less confusing.

ChristinaLK:
I agree that `touch` is less valuable than file creation w/ a text editor - however, given that it's already in a callout, I think it's already been designated as an aside (which is appropriate).  If other folks weigh in and think it's worth deleting as an aside then I'll merge this.  

@gdevenyi , @wking , @r-gaia-cs , thoughts?

wking:
On Mon, Apr 20, 2015 at 04:24:15PM -0700, Christina Koch wrote:
> I agree that `touch` is less valuable than file creation w/ a text
> editor - however, given that it's already in a callout, I think it's
> already been designated as an aside (which is appropriate).  If
> other folks weigh in and think it's worth deleting as an aside then
> I'll merge this.
> 
> @gdevenyi , @wking , @r-gaia-cs , thoughts?  

No preference.  A callout is already fairly peripheral, so it's not
very intrusive.  But creating empty files isn't something that novice
shell users will need to do frequently, so I'm fine cutting it too.

r-gaia-cs:
~~~
$ grep touch *.md
02-create.md:> One of the simplest ways to create an empty file is via the `touch` command.
02-create.md:> then touch an empty file called `draft.txt`:
02-create.md:> $ touch draft.txt
reference.md:*   `touch path` creates an empty file if it doesn't already exist.
~~~

+1 to drop `touch`.

@ocefpaf Can you remove the description of `touch` at `reference.md`?

ocefpaf:
@r-gaia-cs done.  Do you usually squash the commits or leave them be?

gvwilson:
Leave them be - we have no objection to squashing, but don't see a 
particular need for it.

r-gaia-cs:
@ocefpaf as @gvwilson you don't need to squash the commits.

Issue #124, opened 2015-04-20
Consider changing /users to /home


ocefpaf:
Users using modern Linux distributions in their laptops while following the lesson will recognize `/home` immediately, while `/users`  might bring some confusing.

r-gaia-cs:
We use `/Users` because is what Mac OS X uses. I already complains once about this but leave that way.

ocefpaf:
Ah makes sense... or not :stuck_out_tongue_winking_eye:

ocefpaf:
@r-gaia-cs 
BTW: here http://swcarpentry.github.io/shell-novice/02-create.html  (in Another Useful Abbreviation) the lesson uses `/home` instead of `/users`.  Maybe that is on purpose to show the two different standards?  Although, I would vote to keep the lesson consistent.

gvwilson:
+1 to consistency; +1 to showing the Mac usage (since we have more Mac 
users in workshops than Linux users); +1 to a small callout box 
explaining what the path is likely to look like on Linux, and on Windows.

ChristinaLK:
addressed in PRs

PR #128, opened 2015-04-20
Changing /home to / users.

ocefpaf:
Partially solves #124.

PS: I am unsure if Mac uses `/Users` or `/users`.

### DIFF
--- a/01-filedir.md
+++ b/01-filedir.md
@@ -62,18 +62,18 @@ i.e.,
 the directory that the computer assumes we want to run commands in
 unless we explicitly specify something else.
 Here,
-the computer's response is `/users/nelle`,
+the computer's response is `/Users/nelle`,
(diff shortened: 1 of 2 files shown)


r-gaia-cs:
> PS: I am unsure if Mac uses `/Users` or `/users`.

AFAIR is `/Users`.

iglpdc:
Yep, /Users.

ocefpaf:
Thanks @iglpdc and @r-gaia-cs I think I got all `/users`.

r-gaia-cs:
@ocefpaf Can you squash this pull request? You asked me early today if squash the commits are required and I said no. It isn't but I think is something good in this case, i.e. having only one commit changing from `/home` to `/Users` instead of two commits.

ocefpaf:
@r-gaia-cs done.

PR #129, opened 2015-04-20
Added home directory callout.

### LINKED ISSUE #124: Consider changing /users to /home

ocefpaf:
Together with #128 this PR closes #124.

### DIFF
--- a/01-filedir.md
+++ b/01-filedir.md
@@ -72,8 +72,16 @@ $ pwd
 /users/nelle
 ~~~
 
+> ## Home directory {.callout}
+>
+> The home directory path will look different on different operating systems.
+> On Linux it would look like `/home/nelle`,
+> and on Windows it will be similar to `C:\Documents and Settings\nelle`.
+> Note that it may look slightly different for different versions of Windows.
+
+
 > ## Alphabet Soup {.callout}
-> 
+>
 > If the command to find out who we are is `whoami`, the command to find
 > out where we are ought to be called `whereami`, so why is it `pwd`
 > instead? The usual answer is that in the early 1970s, when Unix was
@@ -160,7 +168,7 @@ the shell thinks we're trying to run a command called `ls-F`,
 which doesn't exist.
 
 > ## What's In A Name? {.callout}
-> 
+>
 > You may have noticed that all of Nelle's files' names are "something dot
 > something". This is just a convention: we can call a file `mythesis` or
 > almost anything else we want. However, most people use two-part names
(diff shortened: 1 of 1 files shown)


ChristinaLK:
I would actually move the home directory callout to before the alphabet soup one.  

Also, my experience on windows is that home is not My Documents, but something like: 
`c/Users/Name`.  Maybe someone else who has used windows could weigh in...?

ocefpaf:
> Also, my experience on windows is that home is not My Documents

To be honest it is not.  It is something like `C:\Documents and Settings\username`.  However, I am unfamiliar how this directory manifest itself in the various Windows versions (XP, Vista, 7, 8, etc).

I chose `My Documents` because it is present in most of the Windows versions I tried.  Also, I think that the concept of that directory is a similar to the `home` directory.  But I am not a Windows expert!

It is important to notice that a shell class on Windows will use git-bash or cygwin.  I am unsure how those programs handle the `home` directory.  Maybe they create a whole new tree inside the installation directory?  Or maybe do nothing?  I don't know...

r-gaia-cs:
I would suggest to do a big reword of this.

> Next, let's find out where we are by running a command called `pwd` (which stands for "print working directory").
>
> > Alphabet Soup
> >
> > ...
>
> ~~~{.in}
> $ pwd
> ~~~
> ~~~{.out}
> /users/nelle
> ~~~
>
> The computer's response is `/users/nelle`, which is Nelle's home directory.
>
> > Home directory
> >
> > ...
>
> When you open the shell (or any other program) it will be running at your home directory. If you want to run a program in a directory other than your home directory you need to make it explicitly.

ocefpaf:
Looking at your message and the Markdown file I am unsure how to proceed.  Maybe, if the `Home Directory callout` is worth merging, we should merge it as is and re-word in another PR.

ChristinaLK:
I agree @ocefpaf, I think we can merge the callout and then someone can submit the reordering in another PR if they like.  

re: this PR: I would like the callout here to be more specific, something like: "the home directory path will look different on different operating systems. On linux…etc.  "
And for Windows, rather than saying "think of it like My Documents", I would give an actual example of a Windows home directory and then just say that it may look slightly different for different versions of Windows.  
Sorry I wasn't more specific about this before!

ocefpaf:
@ChristinaLK How about now?

ChristinaLK:
that's great, thank you!

PR #92, opened 2015-03-02
Update from core 2015 03 02

gvwilson:

PR #130, opened 2015-04-20
Update the Makefile with the latest template.

### LINKED ISSUE #127: Question regarding the us of 3 minus sign in the text
There are some `---` used in the text at some places and, I believe, that should represent a &mdash;.  Is that correct?  If so, should we use some "Markdown flavor/plugin" to take care of that?  Or should we fix it using the HTML syntax (`&mdash;`)?

ocefpaf:
Closes #127

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,6 @@
+PANDOC ?= pandoc
+PANDOC_FLAGS = --smart
+
 # R Markdown files.
 SRC_RMD = $(wildcard ??-*.Rmd)
 DST_RMD = $(patsubst %.Rmd,%.md,$(SRC_RMD))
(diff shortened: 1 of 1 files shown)


wking:
On Mon, Apr 20, 2015 at 12:07:18PM -0700, Filipe wrote:
>   * Update the Makefile with the latest template.

This is supposed to happen by merging the template branch in Git
(which preserves the history and picks up all the changes).  If you
cherry-pick or copy/paste across, it gets hard to keep track of what's
been brought over and what hasn't.  See previous discussion in #92.

ocefpaf:
Ah.  Makes sense.  Closing.

PR #132, opened 2015-04-26
Example of using command substitution in loop

mahdi-b:
Added an example that showcases the power of the command substitution `$()` and  for loops, as seen in section 4.

### DIFF
--- a/06-find.md
+++ b/06-find.md
@@ -383,6 +383,20 @@ $ grep FE $(find .. -name '*.pdb')
 ../data/pdb/heme.pdb:ATOM     25 FE           1      -0.924   0.535  -0.518
 ~~~
 
+Note that `$()` can be used in conjunction with other commands, or in
+loops. For example, the loop below finds both "FE" or "CA" (Calcium)
+in all the `.pdb` files located in the current directory and separates
+the occurrences in each file by a line of dashes.
+
+~~~ {.input}
+$ for filename in $(find . -name '*.pdb')
+> do
+>  grep FE $filename
+>  grep CA $filename
+>  echo "---------------------------"
+> done
+~~~
+
 > ## Binary Files {.callout}
 >
 > We have focused exclusively on finding things in text files. What if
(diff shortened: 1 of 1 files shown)


wking:
On Sat, Apr 25, 2015 at 05:26:54PM -0700, bingo11 wrote:
> Added an example that showcases the power of the command
> substitution `$()` and for loops, as seen in section 4.

This looks reasonable to me.  In fact, I'd consider using it *instead
of* (and not after) the current grep example, because:

  $ find .. -name '*.pdb' -exec grep FE {} \+

is much more idomatic than:

  $ grep FE $(find .. -name '*.pdb')

The former also avoids the “Argument list too long” that the latter
will throw if there are lots of matches.

bingo11:
I agree! I removed the nested command expansion `$()` and replaced it with a `for` loop.
I also added the https://github.com/swcarpentry/shell-novice/issues/133 as a suggestion to add an example showcasing the use of `-exec` with find.

wking:
On Mon, Apr 27, 2015 at 07:32:18PM -0700, bingo11 wrote:
> I agree! I removed the nested command expansion `$()` and replaced
> it with a `for` loop.

This works, but it might be better to start off with something other
than ‘find’ for introducing command substitution.  Loops and variable
substitution are also new, and “look we're using new syntax in
conjunction with two new constructs and a command you just learned
about in the last half hour” seems potentially confusing ;).  This is
pretty close to the end of the topic, so maybe split command
substitution out into a very small topic and use:

  $ wc -l $(grep --recursive --files-with-matches FE .)

for “how many files mention FE?”

ChristinaLK:
Hi @bingo11 , sorry to be chiming in so late on this.  I like the additions, but would like to see the deleted pieces stay as well.  In particular, I think it's valuable to have the following pieces, in this order: 
* introduce the idea of `$()` without a for loop with the original `grep`/`wc` example.   
* segue to, "hey, this `$()` thing would work really well in a for loop, esp. if you want to run multiple commands!" (this would be your `head`/`tail` example)
* I'm okay with leaving out the grep/find at the end.  

My main concern is that we introduce concepts one at a time, so `grep` -> `find` -> `$()`, then add the for loop.  I almost wonder if the part of the lesson on `$()` should be in the for loops lesson, but that's a discussion for another time.  :)

What do you think?

PR #135, opened 2015-05-03
Clarified use of . prefix for configuration files

keipertk:
I think the use of `.` and `..` as relative pathname symbols is well explain, but it is not clear why `.` is used as a prefix for filenames. This change clarifies the usage of the `.` prefix for configuration files.

(This pull request is part of my instructor training)


ChristinaLK:
looks good to me.

PR #136, opened 2015-05-04
removing redundant bullet point

jevbelikov:
my first pull request with a correction for the shell lesson (as part of instructior training)


ChristinaLK:
merged!

PR #141, opened 2015-05-07
Add getready callout

iglpdc:
[This PR in lesson-example](https://github.com/swcarpentry/lesson-example/pull/3) add a "getready" callout box in the `index.md` that can be used to provide the learners with instructions on "what they should do to follow the lessons".

The PR here add the callout box. I've also changed the name of the folder containing Nelle's filesystem from "filesystem" to "data" and make a zip file of it linked in the callout box.


iglpdc:
@ChristinaLK @r-gaia-cs this is how the "getready" callout would work for a lesson that needs some preparation steps.

index.md
```diff
@@ -24,6 +24,19 @@ These lessons will start you on a path towards using these resources effectively
...
+>
+> You need to download some files to follow this lesson:
+> 
+> 1. Make a new folder in your Desktop called `shell-novice`.
+> 2. Download [shell-novice-data.zip](./shell-novice-data.zip) and move the file to this folder.
+> 3. If it's not unzipped yet, double-click on it to unzip it. You should end up with a new folder called `data`.
+> 4. You can access this folder from the Unix shell with:
+>
+> ~~~ {.input}
+> $ cd && cd Desktop/shell-novice/data
```
r-gaia-cs: This will work for GNU/Linux distributions but I'm not sure if it will also work for Mac OS X and Windows with Git Bash.

r-gaia-cs:
@iglpdc Thanks a lot for your amazing job with the Git and Shell lesson.

index.md
iglpdc: You mean the cd && cd Desktop?

If so, yes, it works in the 3 OS that we used, and AFAIK is the only consistent way to do it.

r-gaia-cs: I mean if the directory `Desktop` exists at that location.

iglpdc: > > +> 4. You can access this folder from the Unix shell with:
> > +>
> > +> ~~~ {.input}
> > +> $ cd && cd Desktop/shell-novice/data
> I mean if the directory Desktop exists at that location.
> 

It does for MacOS and git-bash on Win. Never tried for Linux, but you say so.

ChristinaLK:
I made a previous PR to change the `filesystem` directory name and then reverted (see #77).  I do like that using `data` makes the naming consistent, so as long as there isn't strong outcry that `filesystem` is a more indicative name, I'm ok w/ this change.  :)

Question for @iglpdc : if we ever change the data files (at some point), I assume the zip will have to be updated manually - it's not automatically created via the makefile?

iglpdc:
@ChristinaLK I think the best place to take care of this at the level of lesson-template, because pretty much all our lessons require extra files.

I said in [this issue in the python lesson](https://github.com/swcarpentry/python-novice-inflammation/issues/94) that we could have a post-commit hook or something fancy, but maybe just an extra line to the Makefile would do it. So when we generate the html with `make preview`, the file is rezipped.

Issue #138, opened 2015-05-05
Wonky formatting for a challenge item


chungkky:
Re. page on Finding Things file <06-finding.md>, lines 457-471, The formatting of the section on "Matching ose.dat but not temp" is out of wack with the other challenges (the titles is too large, it is not in a box with a pencil icon like other challenge, etc.) Looks like the code has been interpreted as "block indent" and "heading font size" rather than "challenge." I looked at the Markdown file and can't find where the error is but the resulting HTML page is definitely wrong. Can anyone tell me where the problem is in the code?
![formatting-error-06-find](https://cloud.githubusercontent.com/assets/7913902/7467552/f53ef1ce-f2c8-11e4-9d93-5dc5cc00d066.jpeg)

gvwilson:
The problem is the use of back quotes for code formatting in the level-2 header of the challenge: if `ose.dat` and `temp` are in plain font (no backquotes), the formatting is correct.  Is this a bug in Pandoc?

r-gaia-cs:
> I looked at the Markdown file and can't find where the error is but the resulting HTML page is definitely wrong. Can anyone tell me where the problem is in the code?

~~~
$ pandoc <<EOF
# \`foo\` {.bar}
EOF
<h1 id="foo"><code class="bar">foo</code></h1>
$ pandoc <<EOF
# \`foo\` {}{.bar}
EOF
<h1 id="foo" class="bar"><code>foo</code></h1>
~~~

>  Is this a bug in Pandoc?

Nope. This is intended. See my example above. I couldn't find the issue at https://github.com/jgm/pandoc/ that cover this.

I already send a pull request to fix this: https://github.com/swcarpentry/shell-novice/pull/142.

PR #142, opened 2015-05-08
Fix formatting for a challenge item

r-gaia-cs:
Related to #138.

screen shot 2015-05-07 at 21 53 25

### DIFF
--- a/06-find.md
+++ b/06-find.md
@@ -454,8 +454,8 @@ about them."
 > find . -name '*.dat' | wc -l | sort -n
 > ~~~
 
-> ## Matching `ose.dat` but not `temp` {.challenge}
-> 
+> ## Matching `ose.dat` but not `temp` {}{.challenge}
+>
 > The `-v` flag to `grep` inverts pattern matching, so that only lines
 > which do *not* match the pattern are printed. Given that, which of
 > the following commands will find all files in `/data` whose names


06-find.md
```diff
@@ -454,8 +454,8 @@ about them."
 > find . -name '*.dat' | wc -l | sort -n
 > ~~~
 
-> ## Matching `ose.dat` but not `temp` {.challenge}
-> 
+> ## Matching `ose.dat` but not `temp` {}{.challenge}
```
gvwilson: Whoa - why does this fix things? It looks like magic to me.

r-gaia-cs: Normally Pandoc processes curly bracket as normal text:

~~~
$ pandoc <<EOF
{.foo}
EOF
<p>{.foo}</p>
~~~

But in some cases it uses curly bracket for HTML elements attributes. E.g.

    $ pandoc <<EOF
    ~~~ {.bar}
    foo
    ~~~
    EOF
    <pre class="bar"><code>foo</code></pre>
    $ pandoc <<EOF
    \`foo\`{.bar}
    EOF
    <p><code class="bar">foo</code></p>
    $ pandoc <<EOF
    # foo {.bar}  
    EOF
    <h1 id="foo" class="bar">foo</h1>


So far Pandoc allows you to do that for only a few elements (I think that only the three cases above). The trick part is when you have a heading ending with a code. Pandoc rules is that the curly bracket will refer to the near HTML element. So, in

~~~
> ## Matching `ose.dat` but not `temp` {}{.challenge}
~~~

the `{}` refers to `temp` and `{.challenge}` to the heading.

**Extra:** you can escape this behavior using

~~~
$ pandoc <<EOF
# foo \{.bar\}
EOF
<h1 id="foo-.bar">foo {.bar}</h1>
~~~

or

~~~
$ pandoc -f markdown-header_attributes <<EOF
# foo {.bar}
EOF
<h1 id="foo-.bar">foo {.bar}</h1>
~~~

> Whoa - why does this fix things? It looks like magic to me.

@gvwilson This makes sense now?

abought: I learned something new today- thanks, @r-gaia-cs !

ChristinaLK:
looks good to me.

Issue #139, opened 2015-05-05
Unspecific link in 06-find.md


jevbelikov:
There is a side comment in a callout 'Wildcards' promising a lesson (or at least an example for complex searches: "if you want to do complex searches, please look at the lesson on our website") but the link ("our website") only links to the top-level http://software-carpentry.org. Is there a lesson that shows complex find-grep queries so that the link could be updated? In case such example doesn't yet exist, should it be added (e.g. to discussion.md?) or the link removed?

ChristinaLK:
I believe it should be referencing this: http://software-carpentry.org/v4/regexp/index.html

PR #146, opened 2015-05-09
adding correct link to regex lesson in wildcard callout

ChristinaLK:
resolving issue #139 (I think?)


jevbelikov:
Cheers, I've closed the issue (only looked in v 5.3 lessons myself).

PR #150, opened 2015-05-11
Suggest deleting What's in a Name?

daivanielsen:
I am suggesting we delete this section (What's in a Name?) because I don't find it to be informative. I think most people know what a filename extension is. This lesson is already lengthy (took me more than 15 minutes in total to read through/code in my shell/wrap my head around topics...), so cutting out parts that are not directly relevant to the learning objectives may be worthwhile.

### DIFF
--- a/01-filedir.md
+++ b/01-filedir.md
@@ -167,26 +167,6 @@ without it,
 the shell thinks we're trying to run a command called `ls-F`,
 which doesn't exist.
 
-> ## What's In A Name? {.callout}
->
-> You may have noticed that all of Nelle's files' names are "something dot
-> something". This is just a convention: we can call a file `mythesis` or
-> almost anything else we want. However, most people use two-part names
-> most of the time to help them (and their programs) tell different kinds
-> of files apart. The second part of such a name is called the
-> **filename extension**, and indicates
-> what type of data the file holds: `.txt` signals a plain text file, `.pdf`
-> indicates a PDF document, `.cfg` is a configuration file full of parameters
-> for some program or other, and so on.
->
-> This is just a convention, albeit an important one. Files contain
-> bytes: it's up to us and our programs to interpret those bytes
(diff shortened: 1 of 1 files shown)


iglpdc:
I'm -1 for this. Actually, most novices don't know how the extension work. In some operating systems, the extensions are not shown anymore, but in the Unix world (and thru our lessons) extensions are used heavily. Although it's true that people "knows" what the extension is, they often don't know that has nothing to do with the type of the file (i.e. `my_file.pdf` doesn't have to be a pdf file), which is the main point of this block.

jttkim:
-1 from me too. I've just received a bunch of sequences in *.fasta.docx files, and upon saving them as plain text, LibreOffice's word wrapping "intelligence" had, in some cases, appended the first line of sequence data to the header line (containing the sequence ID). This would have resulted in these lines of sequence data to be ignored if I hadn't checked for it, illustrating the importance of clarifying the distinction between file content and file name.

daivanielsen:
Valid points made. This was just a suggestion to shorten the lesson, but I don't mind if this stays. One other idea I had for SWC lessons was to have a Glossary of Terms section for each lesson where key terms that are relevant to that topic are defined. For Shell Novice some examples would be "command line", "argument", "relative path", "absolute path",... and "extension" could be fit in as well. Basically things that are currently in bold in the lessons. This would take some time to create though. Might be something to think about for the future.

iglpdc:
The Glossary is at the end of `reference.md` (see [displayed here](http://swcarpentry.github.io/shell-novice/reference.html)). You can link to the glossary using this syntax in the markdown file:

`I'm linking [directory](reference.html#directory) to its glossary entry.`

gdevenyi:
I'm -1 for this as well. I think its quite important in a unix environment to distinguish between the name of a file and its contents. Students have found things quite informative to rename a text file into a jpeg and think about the implications of whether it became a picture when that happened.

PR #151, opened 2015-05-11
Tweak to Tab Completion section

daivanielsen:
Tab completion is so useful, but I think that it is not highlighted enough in this section. I have proposed some minor tweaks to illuminate the concept a bit further and clarify exactly how it is done. Again, total newbies to the shell (like myself) might not really understand what it means when the lesson says "press tab...", so I think clarifying at least one time to "press the tab key on keyboard" is useful.

### DIFF
--- a/01-filedir.md
+++ b/01-filedir.md
@@ -403,14 +403,14 @@ $ ls north-pacific-gyre/2012-07-03/
 ~~~
 
 This is a lot to type,
-but she can let the shell do most of the work.
+but she can let the shell do most of the work through what is called **tab completion**.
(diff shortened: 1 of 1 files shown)


iglpdc:
That's a good change. It's true that people get confused when they hear "press tab". Maybe it's a good idea to link to [this Wikipedia page](http://en.wikipedia.org/wiki/Tab_key). Macs don't have "Tab" written in the key, so sometimes you have to explain which one it is.

ChristinaLK:
I like this addition!  Merging it in.

PR #153, opened 2015-05-13
Change "data/users" to "data/Users"

iglpdc:
So it fits the rest of the lesson. Updates also the zip file.


ChristinaLK:
thank you!

PR #158, opened 2015-05-22
Create pipe-mcq.md

hpourreza:
An MCQ about piping in Unix shell.


r-gaia-cs:
@hpourreza Thanks for the pull request. This is something for instructor training?

PR #159, opened 2015-05-22
Create sort-mcq.md

hpourreza:
An MCQ about sort command in the Unix shell.

### DIFF
--- a/sort-mcq.md
+++ b/sort-mcq.md
@@ -1,0 +1,47 @@
+The file `input.txt` contains the following lines:
+~~~{.output}
+1234
+2345
+1
+45
+~~~
+
+In which order will the lines be returned by the following command:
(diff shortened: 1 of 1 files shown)


r-gaia-cs:
@hpourreza Thanks for the pull request. This is something for instructor training?

hpourreza:
Hi
One of them was but I was interested in doing it so I did some extra ones. 

Hossein

r-gaia-cs:
> One of them was but I was interested in doing it so I did some extra ones.

Great. Could you add the challenge inside the lesson (and not as a separate file)? For example,
check the last lines of https://raw.githubusercontent.com/swcarpentry/lesson-example/gh-pages/01-one.md for how we type the challenges. If you need help, just let us know.

PR #160, opened 2015-05-24
Added missing files from the data directory

DamienIrving:
Looks like these files were forgotten in the move across from the bc repo. At one point during `01-filedir.md` the `ls -F /data` command is executed, so it's important that they are there.


iglpdc:
Remember that at some point, maybe when the html is generated or maybe now, the zip file with all data files should be updated with `zip -r python-inflammation-data.zip data`.

ChristinaLK:
ty!  will merge.

PR #162, opened 2015-05-28
Convert blockquote to callout

r-gaia-cs:


ChristinaLK:
I'm going to merge this for now, but it might be worth having a further discussion (lesson-wide) about the purpose of callouts - are they an tangential discussion to the main material (so, optional reading) or an aside that probably should be read in the lesson.

Issue #155, opened 2015-05-21
.tmp and .my.log hidden files extraneous?


jttkim:
There's an emtpy file called `.tmp` in `data/Users/larry`, and there are `.tmp` files in other directories as well. Likewise, there are empty files `.my.log` in some directories. `find data -name '.*' -print` prints a full list of dotfiles.

My impression is that only `data/Users/nelle/.bash_profile` could serve any didactic purpose (other than perhaps some "easter egg hunt" type challenge). Is this correct, and if so, should we get rid of these `.tmp` and `.my.log` files?

gvwilson:
Good catch - yes, I believe the .tmp files should go.  We should add an 
entry to .gitignore to exclude them in future.

ChristinaLK:
I put them there because I didn't think I could commit the directories if they were empty.  If that's not true, then they can be removed.

drj11:
@ChristinaLK you're write that git can't commit empty directories. If you need that, you add a dummy file as you did.

There is a convention of calling these files `.gitkeep` but it is entirely an unenforced social convention.  See http://stackoverflow.com/questions/7229885/what-are-the-differences-between-gitignore-and-gitkeep

r-gaia-cs:
#165 tries to address this issue.

PR #165, opened 2015-05-28
Use .gitkeep

r-gaia-cs:
Address #155.


ChristinaLK:
hi @r-gaia-cs , sorry for the late follow up with this.  I don't think we need the callout explaining the gitkeep files - they'll be mostly invisible anyway and there's enough random files in these directories that these don't need a special mention.  Can you remove the callout + then I'll merge the PR?

r-gaia-cs:
> I don't think we need the callout explaining the gitkeep files - they'll be mostly invisible anyway and there's enough random files in these directories that these don't need a special mention.
> Can you remove the callout?

Done.

ChristinaLK:
thanks!

PR #187, opened 2015-06-02
Change 'users' to 'Users' in filesystem-challenge.svg

lexnederbragt:
Make it consistent with capitalised 'Users' in the challenge question in 01-filedir.md and with the rest of the material


ChristinaLK:
Sorry this was so late!  thanks for this change.

PR #188, opened 2015-06-02
Fix Saving to a File in a Loop - Part One

huguesfontenelle:
The phrasing "copies ... to create ..." was incorrect.
Explanation: it is not the file that is copied, but its content!
Rephrase to "copies the text content of ... in ... "
This is now consistent with the next exercise.

### DIFF
--- a/04-loop.md
+++ b/04-loop.md
@@ -435,11 +435,11 @@ so she decides to get some coffee and catch up on her reading.
 > ~~~
 > 
 > 1.  Prints `fructose.dat`, `glucose.dat`, and `sucrose.dat`, and
->     copies `sucrose.dat` to create `xylose.dat`.
+>     copies the text content of `sucrose.dat` in `xylose.dat`.
 > 2.  Prints `fructose.dat`, `glucose.dat`, and `sucrose.dat`, and
->     concatenates all three files to create `xylose.dat`.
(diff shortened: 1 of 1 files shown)


04-loop.md
```diff
@@ -435,11 +435,11 @@ so she decides to get some coffee and catch up on her reading.
 > ~~~
 > 
 > 1.  Prints `fructose.dat`, `glucose.dat`, and `sucrose.dat`, and
->     copies `sucrose.dat` to create `xylose.dat`.
+>     copies the text content of `sucrose.dat` in `xylose.dat`.
```
jiffyclub: I think using "into" instead of "in" will be more clear.

jiffyclub:
One problem with this change is that if `xylose.dat` does not exist it is created. Your change makes it seem as though `xylose.dat` is pre-existing. It *can* be pre-existing, but if it does not exist it will be created.

huguesfontenelle:
Then let's use the text inspired from the next exercise (`Saving to a File in a Loop - Part Two`)
>The text from sucrose.dat will be saved to a file called xylose.dat

The conditional creation is still rather implicit.

jiffyclub:
Works for me.

ChristinaLK:
looks good for me - thanks for the PR @huguesfontenelle and review @jiffyclub

PR #189, opened 2015-06-06
presence of absence typo

marwahaha:
quick fix, got pointed out in the SWC workshop we just ran in Berkeley


marwahaha:
also, an ls vs ls -F typo

marwahaha:
see #196

Issue #174, opened 2015-05-28
Add example of expression with more than one word for grep


rgaiacs:
In `06-find.md', there are no example of search for a expression with more than one word.
What about add "is not" in the begin?

pschloss:
This is my first PR with SWC. :)

One thing I wonder in reference to this issue is why grep is taught without quotes in the first place. It would certainly seem a lot safer in the long run to teach quotes from the get go.

ChristinaLK:
I agree - I'll leave some ideas on your pull request

gdevenyi:
Changes merged

PR #201, opened 2015-06-17
Add example of grep'ing for a phrase

pschloss:
This is in response to issue #174.

### DIFF
--- a/06-find.md
+++ b/06-find.md
@@ -98,10 +98,12 @@ $ grep -w "is not" haiku.txt
 Today it is not working
 ~~~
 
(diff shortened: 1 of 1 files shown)


06-find.md
```diff
@@ -87,7 +87,16 @@ This will limit matches to word boundaries.
 $ grep -w day haiku.txt
 ~~~
 
-In this case, there aren't any, so `grep`'s output is empty.
+In this case, there aren't any, so `grep`'s output is empty. Sometimes we don't
```
ChristinaLK: move new text to line 67 (between first example + introduction of flags)

ChristinaLK:
I might add some comment after your example like "we've now seen that you don't have to have quotes around single words, but it is useful to do so, esp. w/ multiple words.  We will use them from now on" and then put quotes around the remaining grep examples.  What do you think?

pschloss:
Sounds good - thanks. The updated PR incorporates the quotes in the remaining grep commands.

Issue #173, opened 2015-05-28
Clarify exercise "Listing Directories and Files"


rgaiacs:
I think the exercise "Listing Directories and Files" in `02-create.md` isn't very clear.

ChristinaLK:
agree.

pschloss:
Sorry for all of the nuisance references above, still getting the hang of submitting multiple pull requests to the same repository. As I indicated in my commit message, IMHO this exercise has little to do with "Listing Files and Directories" and more to do with organization of files and directories. That is the path I took with my pull request.

ChristinaLK:
Haha, no problem.  I'll review the PR.

PR #202, opened 2015-06-17
Clarify Listing Directories and Files exercise

pschloss:
If the goal of the exercise was to test the learner's ability to list
directories and files, then the `ls` commands would need to be modified to get
the desired output. The modifications are far beyond what is covered in the
content. Instead, this exercise is really about organizing a directory using the
`mv` command. I changed the title of the exercise to Organizing Directories and
Files. And provided some additional background on what the learner needs to do.
This is in response to issue #173.

### DIFF
--- a/02-create.md
+++ b/02-create.md
@@ -352,16 +352,19 @@ but it does find the copy in `thesis` that we didn't delete.
 > 3.   `proteins.dat recombine`
 > 4.   `proteins-saved.dat`
 
-> ## Listing Directories and Files {.challenge}
+> ## Organizing Directories and Files {.challenge}
 >
-> Suppose that:
+> Jamie is working on a project and she sees that her files aren't very well
+> organized:
 >
 > ~~~
 > $ ls -F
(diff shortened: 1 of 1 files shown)


02-create.md
```diff
@@ -352,16 +352,19 @@ but it does find the copy in `thesis` that we didn't delete.
...
+> organized:
 >
 > ~~~
 > $ ls -F
 > analyzed/  fructose.dat    raw/   sucrose.dat
 > ~~~
 >
-> What command(s) could you run so that the commands below will produce the output shown?
+> The `fructose.dat` and `sucrose.dat` files contain output from her data
+> analysis. What command(s) does she need to run so that the commands below will
```
ChristinaLK: maybe add "covered in this lesson" after "command(s)", just to make it extra clear?

ChristinaLK:
That's much better already, thank you.  let me know what you think about my one inline comment.

pschloss:
Sounds great, I've updated the PR - thanks!

PR #205, opened 2015-06-23
Added a small amount of extra detail to discussion of ls file sorting

dikaiosune:
One sentence addition to clarify use of ISO dates in folder names.


gdevenyi:
Looks good. Thanks!

PR #209, opened 2015-06-28
Fixed callout style in topic 1

sritchie73:
Added level 2 heading markdown so that callout box renders the same as other callout boxes in the lesson.


ChristinaLK:
good catch - thank you!

PR #210, opened 2015-06-29
Added MCQ to 01-filedir.md as part of instructor training

anelda:
Used MCQ from http://swcarpentry.github.io/recycling/2014/10/assessment-2/index.html to bring across the idea that there are many ways (including relative and absolute paths) to navigate around the filesystem


ChristinaLK:
Finally catching up with PRs - this is a great question @anelda, will merge!

PR #211, opened 2015-07-07
Small changes to '*' wildcard description.

ryanmid:
Changed the text at the end of two sentences to explicitly state which groups of files the wildcard expressions are matching.


iglpdc:
+1 Looks much clearer to me.

r-gaia-cs:
Thanks @ryanmid. Another +1.

PR #212, opened 2015-07-08
Fix small typo in editor callout

Cadair:
Just a very small grammatical tweak I found when reading through the lesson.


ChristinaLK:
I think this is already correct the way it is.

Cadair:
ah yes, I see.

PR #213, opened 2015-07-10
Update 04-loop.md

mboisson:
If Nelle wants to create a backup of the files, it is more likely she would want to copy the files than move them. I replaced mv by cp everywhere.

### DIFF
--- a/04-loop.md
+++ b/04-loop.md
@@ -24,22 +24,22 @@ as `original-basilisk.dat` and `original-unicorn.dat`.
 We can't use:
 
 ~~~ {.bash}
-$ mv *.dat original-*.dat
+$ cp *.dat original-*.dat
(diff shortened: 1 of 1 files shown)


ChristinaLK:
I agree that `cp` makes more sense than `mv` in the context of making backup files.  However, making this change requires more than just changing the commands - the error message from using `mv` as well as its explanation would also be changed.  If you're willing to make those changes, please add them to this PR for review, otherwise I'll leave the lesson as is.

mboisson:
I modified the explanation as well. The error message from using cp is the same as the one from using mv for me, i.e. that the target is not a directory.

gdevenyi:
This is a excellent change, moving files is definitely not for backing up.

PR #216, opened 2015-08-17
Show which shell commands are used in a lesson

briandk:
## Major Changes

- At the top of a lesson, include a list of which commands will be used

I took Lesson 1 (directories) as an example

## Comparing Before/After

2015-08-17 - improving software carpentry unix shell lesson 08-17-15 2 49 08 pm

## Justification

As I was working through the tutorial, I found it useful to know which lessons would cover which commands. I often found myself asking "wait, did this tutorial cover `ls`, or was that in another tutorial? Also, I think that if people come to these lessons looking for reference beyond just the manpage, it's helpful to know up front that a lesson will cover/use a specific command they're interested in.

## Notes/Caveats

When I ran `make preview`, it also seems to have introduced some incidental changes to the other chapters, including adding `id` attributes to some sections. I’m guessing this is because of updates to the CommonMark spec?

### DIFF
--- a/00-intro.html
+++ b/00-intro.html
@@ -30,7 +30,7 @@
           <h2 class="subtitle">Introducing the Shell</h2>
           <section class="objectives panel panel-warning">
 <div class="panel-heading">
-<h2 id="learning-objectives"><span class="glyphicon glyphicon-certificate"></span>Learning Objectives</h2>
+<h2><span class="glyphicon glyphicon-certificate"></span>Learning Objectives</h2>
 </div>
 <div class="panel-body">
 <ul>
--- a/00-intro.html
+++ b/00-intro.html
@@ -30,7 +30,7 @@
           <h2 class="subtitle">Introducing the Shell</h2>
           <section class="objectives panel panel-warning">
 <div class="panel-heading">
-<h2><span class="glyphicon glyphicon-certificate"></span>Learning Objectives</h2>
+<h2 id="learning-objectives"><span class="glyphicon glyphicon-certificate"></span>Learning Objectives</h2>
 </div>
 <div class="panel-body">
 <ul>
--- a/01-filedir.html
+++ b/01-filedir.html
@@ -30,7 +30,7 @@
           <h2 class="subtitle">Files and Directories</h2>
           <section class="objectives panel panel-warning">
 <div class="panel-heading">
-<h2 id="learning-objectives"><span class="glyphicon glyphicon-certificate"></span>Learning Objectives</h2>
+<h2><span class="glyphicon glyphicon-certificate"></span>Learning Objectives</h2>
 </div>
 <div class="panel-body">
 <ul>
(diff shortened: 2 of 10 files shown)


rgaiacs:
Thanks for the pull request.

> At the top of a lesson, include a list of which commands will be used

I prefer the list of commands at the end of the lesson **but you can convince me** that the top is better.

> When I ran `make preview`, it also seems to have introduced some incidental changes to the other chapters, including adding id attributes to some sections. I’m guessing this is because of updates to the CommonMark spec?

This is probably because the use of different versions of Pandoc. **We are requesting that pull request only cover changes on Markdown files** and lesson maintainers will commit the changes on HTML files. If you need help to let us know.

briandk:
> I prefer the list of commands at the end of the lesson but you can convince me that the top is better.

I completely see your point. My only argument for putting them at the top (versus the bottom) is that they feel like they belong near the learning objectives. So, a person scanning the top of the lesson would see "oh, here's what I'll be able to do by the end of this tutorial, and here are the commands I'll learn/see used in this tutorial." My further rationale is that design-wise, if it's at the top a reader would see it without doing any scrolling. If it's at the bottom, a reader would have to think to go all the way down to find this info, which they might be less likely to do.

> **We are requesting that pull request only cover changes on Markdown files** and lesson maintainers will commit the changes on HTML files. If you need help to let us know.

Sorry about that! I'd even read the contributor guidelines and knew that when I commited :cry: 

With the latest pushes to this branch, I've reverted the HTML changes, so the net patch change affects only the markdown file. 

Let me know how else I can help!

rgaiacs:
> Sorry about that!

No problem at all. We love all pull request.

> Let me know how else I can help!

You will need to wait a few days to give time for other contributors make comment about your proposal.

tomwright01:
I think this is a great idea. Personally I think having the list at the top of the page is more logical.

iglpdc:
I don't really see the need for this. I think this is very useful for the instructors (as in your justification above), but no so much for the learners. We are assuming that before starting the lesson, learners know nothing about the commands. As @rgaiacs, I like it better at the end of the lesson, as a cheatsheet or a summary they can look up in the rest of the workshop.

Also:

- it adds more stuff to maintain, duplicating the glossary and somehow the learning objectives,
- it complicates the lesson template with an extra box,
- it is hard to do in other lessons with more complex commands, like python or R,
- it is not useful when teaching the lesson in a workshop (I don't want to spend time going quickly thru the summary of the commands when I'll be demonstrating them in a few minutes.)

So, my opinion is that this is good, but better to put at the end of the lesson (to reinforce what we have just seen), the instructor's guide (so instructors can quickly map the lesson in their minds), and/or the glossary (so learners can use it to as a cheatsheet during the rest of the workshop).

briandk:
@iglpdc I'd be willing to compromise by putting it at the bottom

WilDoane:
@iglpdc One question to consider WRT placement of the commands list is the apparent difference between [how novices approach a learning task and how experts structure content knowledge](http://www.nap.edu/openbook.php?record_id=9853&page=31). Novices lack the content knowledge to recognize new patterns reliably as they're revealed; experts use their prior knowledge to ascribe meaning to patterns. Making overall patterns manifest for the learner before introducing new content provides [a basis upon which learners can construct their understanding](http://www.nap.edu/openbook.php?record_id=9853&page=243).

If you're writing this material primarily for experts, teachers, etc, then placing the command overview at the bottom is reasonable. If you're writing this material primarily for novices, then placing the command overview at the top (or at both the top and bottom) makes more sense.

ChristinaLK:
My vote is for the top - it seems to fit conceptually with the idea of objectives, and for the not-quite-completely-novice learner, it might be a helpful memory jog.  My only worry is that it takes up page space and could be redundant to maintain.  I'm willing to merge it in and get feedback from instructors - if no one likes it and/or it's distracting, we can always take it out or move it to the bottom of the page.  

Any strong objections to this?

briandk:
@ChristinaLK has got me thinking about the space issue: what if we included it as a kind of collapsible `<div>`, so it would be much thinner (vertically) and users could click to disclose the list of commands if they wish. The compromise there is that it wouldn't add much visual bulk to the top of the page, and people who don't want to see the commands listed won't have to.

rgaiacs:
> what if we included it as a kind of collapsible `<div>`

Collapsible `<div>` are troubles. I want to use https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details but Firefox doesn't support it. =(

ChristinaLK:
A collapsible div sounds great, if all browsers supported it.  :(  Alas.  

What's the largest number of commands introduced in a lesson?  As long as it's not greater than 5-6, I'd be okay with trying out a non-collapsed box and see how it goes.

gdevenyi:
This is a good idea, but I have concerns about the maintainability of this as the lessons drift and get edited...

Mahdisadjadi:
What's the status of this PR? I vote for the bottom, like a conclusion box to sum up what was covered in the lesson!

shwina:
FWIW, I am -1 to introducing commands out of context in the beginning. In my opinion, it's more useful "build up" to the use of a unix command rather than learn the command first and then have a look at an example.

gdevenyi:
I think this feature would be better suited to generalize for the template and be somehow auto generated.

@gvwilson where did the new template go?

PR #218, opened 2015-08-21
Update 01-filedir.html

andypohl:
misspelling, I think.


ChristinaLK:
Hi @andypohl , sorry for the late reply.  We typically request PRs on the markdown, not html files - can you resubmit with this change?

PR #230, opened 2015-09-10
Remove mention of "terminal"

phineas42:
Introducing the term "terminal" is unnecessary load for this lesson. Though I do think it's useful jargon, I don't know where it fits.

### DIFF
--- a/01-filedir.md
+++ b/01-filedir.md
@@ -13,6 +13,13 @@ minutes: 15
 > *   Identify the actual command, flags, and filenames in a command-line call.
 > *   Demonstrate the use of tab completion, and explain its advantages.
 
+
+> ## Terminal-ogy {.callout}
+> This lesson is about the **Unix Shell**. You may hear people say the terms 'terminal,'
+> 'bash,' 'console,' 'CLI,' or sometimes 'prompt.'
+> Often, "shell" is what is meant by those terms, even in the context of this lesson, and
+> even though the terms may also refer to things other than the shell.
+
 The part of the operating system responsible for managing files and directories
 is called the **file system**.
 It organizes our data into files,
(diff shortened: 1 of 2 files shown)


tomwright01:
Since on many computers (mac & nix) 'terminal' is the command that is run to access the shell I think maybe both are important.

phineas42:
Absolutely, that is true. In fact, I think everything is important, but nobody wants to hear me, impassioned though I may be, ramble about everything. As an instructor, I run the risk of following a tangential lesson path too deeply. I think there could be a time-sink of a tangent here because "terminal" is overloaded and can refer to the terminal program (as you suggest), the terminal emulation settings (which are addressed in the .bash_profile), and the [virtual] terminals known to the O/S. Additionally, while I recognize that there is no useful distinction to a novice about the difference between "shell" and "terminal", they are nonetheless not equivalent terms.

If making the connection between the "terminal program" and the shell is important in shell-novice, I think it would be good to do so in the intro lesson rather than in the filedir lesson, as it's in the intro lesson that we have to make that connection for the first time.

iglpdc:
What about a small callout box at the beginning of the lesson explaining
the many names for "terminal" (in a loose sense) and maybe their
differences? I can think about shell, terminal, CLI, "bash"... Once this is
explained, we could remove all of them but one in the rest of the lesson.

phineas42:
Going down that road, one thing I'd consider saying when I'm teaching this is something like: "This lesson is about the shell. I may say (and you may hear from others) the terms 'terminal,' 'bash,' 'console,' or sometimes 'prompt.' Often, "shell" is what is meant by those terms, even though they may also refer to other things."

ChristinaLK:
I like the idea of a "terminal-ogy" callout box that lists all the different words that might be used.  The one thing I would avoid is a super technical definitiion of what all of them mean.  @phineas42 , what you wrote in your last comment sounds good - mind closing this PR and opening a new one with that addition?

iglpdc:
Love "Terminal-ogy" as title for the box!

jttkim:
+1 for "Terminal-ogy" and +1 for removing "(terminal)" after "shell" -- `~/.bash_profile` is run by the shell, regardless of the terminal (i.e. even if there is none).

The second occurrence of "terminal" in the context of preventing `ls` from "cluttering the terminal" is much better, as it points to an orthogonality where the terminal is the same, regardless of whether it's used by the shell, `ls`, or any other process to output into.

This is quite relevant to the "Pipes & Filters" section -- the "Terminal-ogy" box could be written with this in mind, and once that's done the wording "screen" (e.g. in "whatever the process sends to standard output to the screen") could be changed to "terminal".

PR #231, opened 2015-09-12
Fix grammatical error: 'and this another example'

Tjips:
This is just a first pull request to complete the instructor training. Just a little mistake I noticed while reading through the material.


ChristinaLK:
Thanks!  merging in...

PR #238, opened 2015-10-04
Added example files to shell-novice-data for use with challenge Qs

emilydolson:
In Pipes and Filters (and again in Shell Scripts), some of the challenge questions involve the example files salmon.txt and animals.txt. Currently, these questions are framed as examples that learners are supposed to work through in their heads/through discussion. While this is valuable, I think there's a lot of benefit to having learners actually writing their own pipes and scripts too. I've tried this out for the "Pipe Creation" challenge problem at the past two workshops I've taught, and it seems like it's gone well. It forces less confident learners to engage, and gives them a nice confidence boost when they can see that they got it right!

This modification adds salmon.txt and animals.txt to the shell-novice-data.zip simulated file system in the /Users/nelle/data directory, so that instructors have the option of presenting these challenge problems in a more hands-on way.


ChristinaLK:
I've had fellow instructors who thought this would be a good idea - thanks for the addition!  One request: can you go into the `01-filedir.md` file and add these two files when relevant (the example folder for `ls` is the data one).

emilydolson:
Good idea. Done!

ChristinaLK:
Parfait.  :)

PR #240, opened 2015-10-10
Added information on cd -

tpoisot:
I think that `cd -` is incredibly useful.

### DIFF
--- a/01-filedir.md
+++ b/01-filedir.md
@@ -348,6 +348,45 @@ it also displays another special directory that's just called `.`,
 which means "the current working directory".
 It may seem redundant to have a name for it,
 but we'll see some uses for it soon.
+
+`cd` also allows you switch back and forth between places in a rapid way,
+using `cd -`.
+Much like `..`, `-` is a special instruction that `cd` translates to
+*the previous directory I was in*.
+For example, if we are currently in Nelle's `data` directory,
+
+~~~ {.bash}
+$ pwd
+~~~
+~~~ {.output}
+/Users/nelle/data
+~~~
+
+we can move to Nell's Desktop with
+
+~~~ {.bash}
+$ cd /Users/nelle/Desktop
+$ pwd
+~~~
+~~~ {.output}
+/Users/nelle/Desktop
+~~~
+
+If we want to go back to the `data` directory,
+we just need to type `cd -`:
+
+~~~ {.bash}
+$ cd -
+$ pwd
+~~~
+~~~ {.output}
+/Users/nelle/data
+~~~
+
+This is a *very* efficient way of moving back and forth between directories. 
+The difference between `cd ..` and `cd -` is that the former brings you *up*,
+while the later brings you *back*.
+
 Finally, we also see a file called `.bash_profile`. This file usually contains settings to customize the shell (terminal). For this lesson material it does not contain any settings. There may also be similar files called `.bashrc` or `.bash_login`. The `.` prefix is used to prevent these configuration files from cluttering the terminal when a standard `ls` command is used.
 
 > ## Orthogonality {.callout}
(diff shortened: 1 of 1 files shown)


01-filedir.md
```diff
@@ -348,6 +348,45 @@ it also displays another special directory that's just called `.`,
 which means "the current working directory".
 It may seem redundant to have a name for it,
 but we'll see some uses for it soon.
+
+`cd` also allows you switch back and forth between places in a rapid way,
+using `cd -`.
+Much like `..`, `-` is a special instruction that `cd` translates to
```
rgaiacs: Could we drop "Much like `..`" because "if `-` is like `..` I should be allow to run `cd -/some-dir` as the same way of `cd ../some-dir`".

01-filedir.md
```diff
@@ -348,6 +348,45 @@ it also displays another special directory that's just called `.`,
...
+~~~ {.bash}
+$ cd -
+$ pwd
+~~~
+~~~ {.output}
+/Users/nelle/data
+~~~
+
+This is a *very* efficient way of moving back and forth between directories. 
+The difference between `cd ..` and `cd -` is that the former brings you *up*,
```
rgaiacs: I think that we could drop this.

01-filedir.md
```diff
@@ -441,6 +480,21 @@ and we will see it in many other tools as we go on.
 > 3.  `2012-12-01/ 2013-01-08/ 2013-01-27/`
 > 4.  `original pnas_final pnas_sub`
 
+> ## Moving back and forth {.challenge}
+>
+> For a hypothetical filesystem location of /home/amanda/data/,
```
rgaiacs: "/home/amanda/data/" -> "`/home/amanda/data/`"

rgaiacs: I thought that we agree to use `/User/` instead of `/home` in all the lesson.

rgaiacs:
+1 to include `cd -`.

01-filedir.md
tpoisot: Why? I think this is the core of what `cd - ` does. Unless I don't understand your comment?

01-filedir.md
tpoisot: Good call.

01-filedir.md
rgaiacs: "the previous directory I was in" at line 355 is very explanatory to me so I don't see the need to explain it again. But I'm OK leaving this as is.

tpoisot:
I changed `/home/` into `/Users/` for two questions. It *should* be homogeneous across the lesson now.

rgaiacs:
@tpoisot Thanks for the pull request. +1 to merge.

01-filedir.md
```diff
@@ -348,6 +348,46 @@ it also displays another special directory that's just called `.`,
 which means "the current working directory".
 It may seem redundant to have a name for it,
 but we'll see some uses for it soon.
+
+`cd` also allows you switch back and forth between places in a rapid way,
+using `cd -`. `cd` will translate `-` into
+*the previous directory I was in*,
+which is faster than having to remmber, then type,
```
ChristinaLK: remmber -> remember

01-filedir.md
```diff
@@ -348,6 +348,46 @@ it also displays another special directory that's just called `.`,
...
+For example, if we are currently in Nelle's `data` directory,
+
+~~~ {.bash}
+$ pwd
+~~~
+~~~ {.output}
+/Users/nelle/data
+~~~
+
+we can move to Nell's Desktop with
```
ChristinaLK: Nell's -> Nelle's

ChristinaLK:
I'm okay with this, although I'd like at least one other person to look at it (@gdevenyi ?).  I would take out the exercise - I think the following "home dir" exercise is more effective at testing comprehension.

jevbelikov:
At a recent workshop I mentioned 'cd -' and participants found it useful and mentioned in the feedback, especially when changing to a directory within a deep hierarchy (otherwise more typing).

ChristinaLK:
addressed in #241

Issue #194, opened 2015-06-12
What changes should we make for our 5.4 release?


gvwilson:
Please use this issue for discussion of changes we should make by mid-August 2015.

ChristinaLK:
These can each spawn more specific issues to be worked on: 

[ ]: read through learning objectives + confirm that lesson content matches.  Do we need new objectives?  To remove others?  
[ ]: Look at callouts + what purpose they serve.  Are there points where callout text should be re-integrated into main text or moved to `discussion.md`, `reference.md`?  
[ ]: In general, having more a list of interactive exercises - either for instructor demo or student participation, not just multiple choice questions.  Pull from what people have actually used in workshops!

ChristinaLK:
Will tag other already open issues that we'd like to complete w/ the milestone

ChristinaLK:
Issues I would really like to see addressed are: 
- moving homedir callout ("another useful abbrevation") from 02-create to 01-filedir (#171)
- better handling of man pages / getting help (#55 #176)

ChristinaLK:
Also, after train the trainers today, a "solution" page for the exercises, with solutions and - more importantly - what concept the question is trying to test and what misconceptions are attached to the wrong answers.

gvwilson:
*  Maxime Boissoneault asked about the use of '/' with the downloaded files in [this message on the discussion list](http://lists.software-carpentry.org/pipermail/discuss_lists.software-carpentry.org/2015-July/003160.html).
*  Emily Davenport had [the same question](http://lists.software-carpentry.org/pipermail/discuss_lists.software-carpentry.org/2015-July/003161.html).
*  Raniere Silva [explained what he does](http://lists.software-carpentry.org/pipermail/discuss_lists.software-carpentry.org/2015-July/003162.html).
*  Amanda Charbonneau explained why [she finds the whole 'Files and Directories' page difficult to teach](http://lists.software-carpentry.org/pipermail/discuss_lists.software-carpentry.org/2015-July/003163.html).

spthm:
Slightly off topic(?), but I'd like to second Raniere's [most recent message](http://lists.software-carpentry.org/pipermail/discuss_lists.software-carpentry.org/2015-July/003167.html). Pretty much the first thing I get students to do after showing them `pwd` is
```bash
export PS1='\u@\h:\w$ '
```
and then, if they want it to be permanent,
```bash
echo "export PS1='\u@\h:\w$ '" >> ~/.bashrc
```
[...]

PR #241, opened 2015-10-19
Filedir rewrite

ChristinaLK:
Massive(?) rewrite of 01-filedir, to reflect what people will have on their own computers.  Based in part on comments in #194 and feedback from @lmichael107

* reduces "data" zip file from full filesystem model to just data used in exercises
* explicitly explains where example commands will deviate from what's written in the text
* reorders presentation of information

I don't necessarily want to do away with the full filesystem examples + text as it was - please share your ideas for potentially creating two streams of this section of the lesson.


emilydolson:
+1 for making this whole explanation much more intuitive! 

My only suggestion is that as long as we're providing an instruction to change the command prompt, it might be helpful to make it more informative (as suggested by Raniere (http://lists.software-carpentry.org/pipermail/discuss_lists.software-carpentry.org/2015-July/003167.html). Many people's prompts will include their current working directory by default, and it seems unfortunate to suggest that they take that information away. On the other hand, that would probably require changing all of the example input to reflect the new suggested prompt, which may or may not be something people think is worth the effort.

profgiuseppe:
Hello, just adding a few comments about (possible) modifications:
* "on Windows it will be similar to C:\Documents and Settings\nelle.": since Windows Vista, it is "C:\Users\nelle". The next sentence could be kept in case someone still has XP (that uses "Documents and Settings"). Also, while "C" is the default disk for user directories, it should be not taken for granted
* "Relative path resolution": answer 1 states an error that was not explicitly presented in the previous examples (unless it was visualized while explaining "ls-F")

ChristinaLK:
@emilydolson thanks for the feedback!  I agree with the prompt thing - it would probably be better to have the working directory too.  We'll see if I feel up to changing the syntax...

@profgiuseppe thanks for the windows tips - good to get that perspective and I'll edit that section to reflect more recent versions of Windows.

01-filedir.md
```diff
@@ -73,29 +91,26 @@ $ pwd
 /Users/nelle
 ~~~
 
-> ## Home directory {.callout}
+> ## Home Directory Variation {.callout}
 >
 > The home directory path will look different on different operating systems.
 > On Linux it will look like `/home/nelle`,
```
lexnederbragt: --> On Linux it *may* look like `/home/nelle` 

The linux systems I am on don't usually have `/home/username`

01-filedir.md
```diff
@@ -108,45 +123,47 @@ Inside that directory are several other directories:
 `tmp` (for temporary files that don't need to be stored long-term),
 and so on:
```
lexnederbragt: Why is there a colon here, is there supposed to be another image?

01-filedir.md
```diff
@@ -155,135 +172,102 @@ which tells `ls` to add a trailing `/` to the names of directories:
...
 ~~~ {.output}
-amino-acids.txt   elements/     pdb/	        salmon.txt
-animals.txt       morse.txt     planets.txt     sunspot.txt
+shell-novice/
 ~~~
 
-The output shows us that there are four text files and two sub-sub-directories.
+Your output should be a list of all the files and sub-directories on your 
+Desktop, including the `shell-novice` directory you downloaded at 
+the start of the lesson.  Take a look at your Desktop to confirm that
```
lexnederbragt: Maybe add 'If you have many files the first ones listed may disappear out of view. You can use your mouse to scroll up to see them'

lexnederbragt:
Since the lesson now starts from the home area, in `index.md`, this should be removed:

```
> 4. You can access this folder from the Unix shell with:
>
> ~~~ {.input}
> $ cd && cd Desktop/shell-novice/data
> ~~~
```

lexnederbragt:
The pull request mentions 200 changed files, many of which are new datafiles, but it isn't clear to me where these come from or what they are used for? For example, `shell-novice/data/elements/*.xml` and `data/Users/nelle/data/elements/*.xml`?

01-filedir.md
```diff
@@ -73,29 +91,26 @@ $ pwd
...
+> (Note that it may look slightly different for different versions of Windows.)
+> In future examples, we've used Mac output as the default - Linux and Windows 
+> output may differ slightly, but should be generally similar.  
 
+To understand what a "home directory" is,
+let's have a look at how the file system as a whole is organized.  For the 
+sake of example, we'll be 
+illustrating the filesystem on our scientist Nelle's computer.  After this
+illustration, you'll be learning a command to explore your own filesystem, 
+which will be constructed in a similar way, but not be exactly identical.
```
sstevens2: I really like pointing out that their filesystem will be similar but not identical and the commands they will learn next will show them what their filesystem looks like.

01-filedir.md
```diff
@@ -73,29 +91,26 @@ $ pwd
...
-> Note that it may look slightly different for different versions of Windows.
+> (Note that it may look slightly different for different versions of Windows.)
+> In future examples, we've used Mac output as the default - Linux and Windows 
+> output may differ slightly, but should be generally similar.  
 
+To understand what a "home directory" is,
+let's have a look at how the file system as a whole is organized.  For the 
+sake of example, we'll be 
+illustrating the filesystem on our scientist Nelle's computer.  After this
+illustration, you'll be learning a command to explore your own filesystem,
```
sstevens2: I'd say commands since they are learning pwd, cd and ls, which all give info about the filesystem but not a high priority change.

01-filedir.md
```diff
@@ -108,45 +123,47 @@ Inside that directory are several other directories:
...
+Underneath `/Users`,
+we find one directory for each user with an account on Nelle's machine, 
+her colleagues the Mummy and Wolfman.  
+
+![Home Directories](fig/home-directories.svg)
+
+The Mummy's files are stored in `/Users/imhotep`,
+Wolfman's in `/Users/larry`,
+and Nelle's in `/Users/nelle`.  Because Nelle is the user in our 
+examples here, this is why we get `/Users/nelle` as our home directory.
```
sstevens2: Maybe a note that this is generally the directory where you start when you open a shell?

01-filedir.md
```diff
@@ -155,135 +172,102 @@ which tells `ls` to add a trailing `/` to the names of directories:
...
+> In practice,
+> however,
+> most people use them interchangeably or inconsistently,
+> so we will too.
 
-Now let's take a look at what's in Nelle's `data` directory by running `ls -F data`,
+We can also use `ls` to see the contents of a different directory.  Let's take a 
+look at our `Desktop` directory by running `ls -F Desktop`,
 i.e.,
 the command `ls` with the **arguments** `-F` and `data`.
```
sstevens2: Needs to be updated for the change in line 199
Should be:
the command `ls` with the **arguments** `-F` and `Desktop`.

01-filedir.md
```diff
@@ -342,13 +330,22 @@ Desktop/            north-pacific-gyre/ writing/
...
 which means "the current working directory".
 It may seem redundant to have a name for it,
 but we'll see some uses for it soon.
-Finally, we also see a file called `.bash_profile`. This file usually contains settings to customize the shell (terminal). For this lesson material it does not contain any settings. There may also be similar files called `.bashrc` or `.bash_login`. The `.` prefix is used to prevent these configuration files from cluttering the terminal when a standard `ls` command is used.
+
+> ## Other Hidden Files {.callout}
+> 
+> In addition to the hidden directories `..` and `.`, you'll also see a file 
+> called `.bash_profile`. This file usually 
+> contains settings to customize the shell. For this lesson material
```
sstevens2: In the old data, there was a `.bash_profile` file in `/Users/nelle` that only contained comments.  Their `.bash_profile` might be different.  I think we could remove the "For this lesson material..." line.

02-create.md
```diff
@@ -316,6 +317,28 @@ thesis/quotations.txt
 This time it tells us that it can't find `quotes.txt` in the current directory,
 but it does find the copy in `thesis` that we didn't delete.
 
+> ## What's In A Name? {.callout}
```
sstevens2: Like the new placement for this section.  Makes sense to have it when you are learning to create files.

01-filedir.md
```diff
@@ -24,6 +24,12 @@ Several commands are frequently used to create, inspect, rename, and delete file
 To start exploring them,
 let's open a shell window:
 
+> ## Preparation Magic {.callout}
```
sstevens2: I have mixed feelings about having them change their prompt.  It does make what the learner sees and the guide shows consistent.  However, we generally try to avoid magic that we don't explain and I'm not sure it is necessary here either.  I propose explaining it similarly to how the filesystem and username differences are now explained, that they should expect this might be different depending on how their computer is set up but that the $ will likely be consistent.

sstevens2:
I really like it overall.  I think it makes everything clearer, both for learning it on your own and teaching.

I made several inline comments but here were a few lines I noticed that weren't changed but might need to be with the current proposed changes:
new 501: only about 19 possibilities with the new data
new 515 and 518: answers need to be changed to be consistent with the /Users change in 509 and 511

01-filedir.md
ChristinaLK: @lexnederbragt what do they have instead?

lexnederbragt: This is determined by the HPC IT staff. For example I have:

`/usit/abel/u1/username`

This is basically "abbreviation of the IT staff departments name"/"name of the university's HPC cluster"/"some disk name"/username.

01-filedir.md
```diff
@@ -155,135 +175,102 @@ which tells `ls` to add a trailing `/` to the names of directories:
...
 ~~~
 
 Here,
-we can see that `/Users/nelle` contains six **sub-directories**.
-The names that don't have trailing slashes,
-like `notes.txt`, `pizza.cfg`, and `solar.pdf`,
-are plain old files.
+we can see that our home directory contains mostly **sub-directories**.
+Any names in your output that don't have trailing slashes,
+are plain old **files**.
```
BernhardKonrad: This doesn't seem up-to-date: The current home directory only contains sub-directories.

01-filedir.md
```diff
@@ -342,13 +333,21 @@ Desktop/            north-pacific-gyre/ writing/
 
 `-a` stands for "show all";
 it forces `ls` to show us file and directory names that begin with `.`,
```
BernhardKonrad: Add a half-sentence that files and folders starting with '.' are hidden by default (in shell and Finder)?

cmacdonell:
I really like it.  Great work.

Add-on suggestion: for the multiple choice questions at the end can we include a link for instructors to an existing Socrative quiz?  I don't think all will use it, but it will be helpful for those who want it.

PR #254, opened 2016-01-01
Update index.html

fu9ar:
using the && operator, right at the beginning, might be confusing. I propose that using a semicolon, the standard bash end of line character is better. && does work, and might be "safer" in many cases because it will stop processing the line when a failed command happens, but since cd with no arguments should never fail unless there are serious problems beyond the scope of the exercise, using that functionality is superfluous at this point.

### DIFF
--- a/index.html
+++ b/index.html
@@ -49,7 +49,7 @@
 <li>If it’s not unzipped yet, double-click on it to unzip it. You should end up with a new folder called <code>data</code>.</li>
 <li>You can access this folder from the Unix shell with:</li>
 </ol>
-<pre class="input"><code>$ cd &amp;&amp; cd Desktop/shell-novice/data</code></pre>
+<pre class="input"><code>$ cd ; cd Desktop/shell-novice/data</code></pre>
 </div>
 </section>
 <h2 id="topics">Topics</h2>


index.html
```diff
@@ -49,7 +49,7 @@ <h1 class="title">The Unix Shell</h1>
 <li>If it’s not unzipped yet, double-click on it to unzip it. You should end up with a new folder called <code>data</code>.</li>
 <li>You can access this folder from the Unix shell with:</li>
 </ol>
-<pre class="input"><code>$ cd &amp;&amp; cd Desktop/shell-novice/data</code></pre>
+<pre class="input"><code>$ cd ; cd Desktop/shell-novice/data</code></pre>
```
gvwilson: Why not put the two commands on separate lines to avoid the problem entirely?

fu9ar: Sure, why not?

On Fri, Jan 1, 2016 at 4:35 PM, Greg Wilson <notifications@github.com>
wrote:

> In index.html
> <https://github.com/swcarpentry/shell-novice/pull/254#discussion_r48680862>
> :
>
> > @@ -49,7 +49,7 @@ <h1 class="title">The Unix Shell</h1>
> >  <li>If it’s not unzipped yet, double-click on it to unzip it. You should end up with a new folder called <code>data</code>.</li>
> >  <li>You can access this folder from the Unix shell with:</li>
> >  </ol>
> > -<pre class="input"><code>$ cd &amp;&amp; cd Desktop/shell-novice/data</code></pre>
> > +<pre class="input"><code>$ cd ; cd Desktop/shell-novice/data</code></pre>
>
> Why not put the two commands on separate lines to avoid the problem
> entirely?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/swcarpentry/shell-novice/pull/254/files#r48680862>.
>



-- 
-Kevin at fu9ar.com
214-708-6649

PR #255, opened 2016-01-02
added RNA-seq reads exercise

tmogrady:
Exercise for Fall 2015 instructor training checkout


gdevenyi:
Can you please fix this to be a change to the markdown instead of the html. Markdown is our working format and the html is generated from it.

tmogrady:
Ah, sorry. Done.

PR #256, opened 2016-01-03
exercice about grep

camillemarini:
exercice for instructor training

### DIFF
--- a/06-find.md
+++ b/06-find.md
@@ -505,7 +505,7 @@ about them."
 > 2013-11-07,18
 > ~~~
 >
-> Help her by putting in the right reorder the following commands and pipes :  
+> Help her by putting in the right order the following commands and pipes :  
 > 
 > `cut -d : -f 2`  
 > `>`  
--- a/06-find.md
+++ b/06-find.md
@@ -481,6 +481,43 @@ about them."
 >
 > 4. None of the above.
 
+> ## Track a species {.challenge}
+> 
+> Do you remember Leah from the previous lesson? She has several hundred 
+> data files saved in one directory, each of which is formatted like this:
+> 
+> ~~~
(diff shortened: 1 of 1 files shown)


06-find.md
```diff
@@ -481,6 +481,43 @@ about them."
...
+> containing a list of dates and associated number of this species, such as 
+> for rabbits:
+> 
+> ~~~
+> 2013-11-05,22
+> 2013-11-06,19
+> 2013-11-07,18
+> ~~~
+>
+> Help her by putting in the right reorder the following commands and pipes :
```
ashwinsrnth: Just a tiny bit of re-wording needed here: "right order" would sound better

camillemarini: thanks. I think it is clearer indeed :-)

06-find.md
```diff
@@ -481,6 +481,43 @@ about them."
 >
 > 4. None of the above.
 
+> ## Track a species {.challenge}
```
gvwilson: Species

06-find.md
```diff
@@ -481,6 +481,43 @@ about them."
...
+> 
+> Do you remember Leah from the previous lesson? She has several hundred
```
gvwilson: "Leah has several hundred..."

06-find.md
```diff
@@ -481,6 +481,43 @@ about them."
...
+> containing a list of dates and associated number of this species, such as 
+> for rabbits:
+> 
+> ~~~
+> 2013-11-05,22
+> 2013-11-06,19
+> 2013-11-07,18
+> ~~~
+>
+> Help her by putting in the right order the following commands and pipes :
```
gvwilson: Put these commands and pipes in the right order to achieve this:

PR #257, opened 2016-01-04
Exercise about grep and cut

fpsom:
Exercise for instructor training.

Sample input retrieved from BLAST output in tabular format.


ashwinsrnth:
Hi @fpsom - thanks for your example. It looks great, but it renders slightly incorrectly. You have a list of commands like this:

~~~
`cut -f 1,2`
`grep "100.00"`
`BLAST.sample.output`
~~~

This is rendered as:

`cut -f 1,2` `grep "100.00"` `BLAST.sample.output`

(See the rendered Mardown file on GitHub)

From the Mardown [documentation](https://daringfireball.net/projects/markdown/syntax#p), it looks like the fix is to end each line with two spaces, which forces a line break. Can you try this?

fpsom:
Hi @ashwinsrnth, thanks for the observation!
I've changed these lines by adding the extra two spaces. As far as I can see, it renders ok now.

gvwilson:
This is a good exercise, but we have others now that cover the same material - thanks for submitting it.

PR #261, opened 2016-01-05
AW added an exercise to 05-script.md

AnnaWilliford:
https://github.com/swcarpentry/shell-novice/commit/0c884ddb3ef4abef1e91cedaa4d5eb83b45bd153

### DIFF
--- a/05-script.md
+++ b/05-script.md
@@ -492,14 +492,14 @@ Of course, this introduces another tradeoff between flexibility and complexity.
 > A2. #Usage: script.sh $pathToInputFilesFolder $pathToResultsFolder
 >
 > Q3. How would you change this script to find out how much you spend in user-defined location/store?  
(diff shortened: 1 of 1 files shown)


ashwinsrnth:
Hi @AnnaWilliford - thanks for your contribution. This looks like a nice challenge that pulls together a lot of concepts and can really help reinforce ideas. Here are some comments:

1. The challenge introduces a bunch of files and directories, and although it's clear about what belongs where, it's still a little difficult to keep a clear mental picture of the directory structure. One thing that could definitely improve the challenge is a figure showing the directory structure. Alternatively, you might want to ask learners to draw out the directory structure before working on the problem.

2. The challenge is using `grep` - something that will be introduced shortly in the workshop - and this makes me think that a better place for it would be the next lesson. Although learners will not explicitly work with `grep` for this challenge, it's still an alien command and that can be intimidating.

gvwilson:
Thanks for the PR - I think this one is now redundant with others that have been merged.

Issue #180, opened 2015-05-28
Drop flags "-mindepth" and "-maxdepth" of "find"


rgaiacs:
I think that we can drop the flags `-mindepth` and `-maxdepth` of `find`. Students can always find them looking at the man page.

drj11:
I've always found the flags obnoxiously GNU specific.

+1

jevbelikov:
+1  "-exec" seems more useful

Ry4an:
Added a no-new-content PR for this.

gdevenyi:
Resolved.

PR #206, opened 2015-06-24
Improve find lesson

rgaiacs:
Cover

- [x] #178
- [x] #180

PR #262, opened 2016-01-06
remove mindepth and maxdepth

Ry4an:
Address #180  by removing mindepth and maxdepth.


ashwinsrnth:
@Ry4an - It looks like this was already being tackled (#206).

Ry4an:
Ah, yes, I missed it! No worries.

PR #265, opened 2016-01-06
Scaffolded Example for Section 4: Loops

fu9ar:
It might be good to pull the use of awk in my example and instead just count the carbon atoms in each molecule, but I thought that it would also be a good thing to show people, so I'm putting it in there anyway because it's pretty cool! I think my formatting is right, also, but I'm not totally sure.


gvwilson:
We don't introduce awk in our lessons, so most users won't be able to follow this - is it possible to use `cut` to get the columns instead?

gvwilson:
p.s. can you please contact Greg to help him map your GitHub ID back to a human name and email address? :-)

fu9ar:
I just cut (hehe) that whole example down into a simpler scaffolded example that should still illustrate the concept.

ashwinsrnth:
Thanks @fu9ar - this challenge introduces `grep`, something that will be taught later in the workshop. Could you consider a simpler example, or perhaps relocating this one?

gdevenyi:
Thanks for your contribution, the grep command isn't introduced until later in these lessons. Feel free to re-spin this exercise without grep.

PR #266, opened 2016-01-06
minor change in index

fu9ar:
removed the use of the unexplained and subtly complicated conditional operators in the introduction


gdevenyi:
Fixed by someone at some point, thanks.

PR #268, opened 2016-01-10
Add a new exercise about wildcard expressions to 03-pipefilter.md

mstimberg:
[This is part of the instructor training checkout procedure]

The idea behind this exercise is to show that there are often multiple ways to achieve the same thing in the shell (in this case with wildcard expressions).
Here, `*[AB].txt` was used in the lesson to select all files ending in `A.txt` or `B.txt` but an alternative is to use `*A.txt *.B.txt` (expected response to first question) which will match the same files in this example, although not in the same order (response to second question). The final question is aimed to make learners aware of the subtle differences between the two expressions, most importantly if there are no files matching `*A.txt`(or no files matching `*B.txt`) -- the more verbose expression will give you an error message while the one used in the lesson will not (response to third question).

I am not sure whether the exercise is not a bit too wordy -- there's a lot of text and not much to answer for the learner. Suggestions for improvements welcome :)

### DIFF
--- a/03-pipefilter.md
+++ b/03-pipefilter.md
@@ -552,3 +552,26 @@ so this matches all the valid data files she has.
 > What other command(s) could be added to this in a pipeline to find
 > out what animals the file contains (without any duplicates in their
 > names)?
+
+> ## Wildcard expressions {.challenge}
+>
+> Wildcard expressions can be very complex and this lesson has only scratched
+> the surface of what they allow you to do. However, the syntax details of the
+> more advanced features can be difficult to remember, especially if you use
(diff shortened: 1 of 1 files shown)


gdevenyi:
Thanks for the contribution, can you be a bit more concrete in hinting at the alternative solution, perhaps by hinting that you need more than one expression to do the equivalent?

mstimberg:
> Thanks for the contribution, can you be a bit more concrete in hinting at the alternative solution, perhaps by hinting that you need more than one expression to do the equivalent?

Good point -- I added an explicit hint and also took the opportunity to rebase this on latest upstream `gh-pages`.

PR #269, opened 2016-01-11
Update 03-pipefilter.md

juanu:
Added an exercise related to counting entries in a Fasta file, combining the commands grep, uniq, and cat.
Any comments and suggestions are more than welcome. I tried to be clear on the explanation, but maybe I'm missing things.

### DIFF
--- a/03-pipefilter.md
+++ b/03-pipefilter.md
@@ -516,3 +516,27 @@ so this matches all the valid data files she has.
 > What other command(s) could be added to this in a pipeline to find
 > out what animals the file contains (without any duplicates in their
 > names)?
+
+> ## Looking for repeated entries in a Fasta file {.challenge}
+>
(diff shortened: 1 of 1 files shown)


shwina:
Hi @juanu - thanks for your PR.

Remember that the shell lesson is taught to learners from varied academic backgrounds. As a mechanical engineer, the description of the FASTA file that you provide is not very useful for me:

~~~
> UniqueID
> AMINO ACID SEQUENCE
~~~

For example, what does a UniqueID look like? Is it a fixed length? Does it have only numbers? Also, what does an amino acid sequence look like?

I think a more concrete example of an entry would be very useful :)

gdevenyi:
@juanu can you update your PR with a concrete example as per @shwina 's request? Thanks.

gvwilson:
Thank you for the PR - I think this is now redundant with others that have been merged.

PR #270, opened 2016-01-11
Fix data directory instructions

kblin:
In the 'Getting read' box on the index page, the shell command was surrounded by a block of triple
tildes. This causes the html to display a subscript tilde before and after the command.
I assume that a triple backtick was desired instead, causing a `<pre><code>` box in HTML.

Signed-off-by: Kai Blin <kblin@biosustain.dtu.dk>

### DIFF
--- a/01-filedir.md
+++ b/01-filedir.md
@@ -211,11 +211,11 @@ we want a listing of something other than our current working directory:
 $ ls -F Desktop
 ~~~
 ~~~ {.output}
-shell-novice/
+data-shell/
 ~~~
 
 Your output should be a list of all the files and sub-directories on your 
-Desktop, including the `shell-novice` directory you downloaded at 
+Desktop, including the `data-shell` directory you downloaded at
 the start of the lesson.  Take a look at your Desktop to confirm that 
 your output is accurate.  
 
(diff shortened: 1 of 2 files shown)


kblin:
Hm, sorry. Looks like github futzes around with my commit message, trying to interpret it as markdown instead of plain text. Let me try and fix that.

kblin:
Also, it looks like a813e253 actually removed the part where the user changes into the correct directory. I've added two patches to fix this. I'll rename the pull request to match the new content.

kblin:
Also fixed the references in the file & directory lesson.

kblin:
I just realized that the ~~~ does different markdown magic depending on there being a blank line in front of it or not. Let me know if you want to keep the ~~~ in the first patch, but add a blank line in front of it.

ChristinaLK:
Hi @kblin - I beat you to the fix!  Thanks for sending this in anyway - it's good to know someone else caught this besides me.

kblin:
Great, thanks for fixing it. :)

kblin:
Hey @ChristinaLK, it's not quite correct yet. index.md still says only `cd` while it should say `cd && cd Desktop/data-shell` in the instructions on how to enter the example directory.

kblin:
01-filedir also still has the wrong command to actually set `$PS1`.

kblin:
Ok, scratch the last comment, that's my shell customisation misleading me. Sorry for the noise.

PR #271, opened 2016-01-11
filedir: export the prompt changes to make them stick

kblin:
Fix a typo in the preparation magic callout that prevents the PS1 changes to actually persist.

Signed-off-by: Kai Blin <kblin@biosustain.dtu.dk>

### DIFF
--- a/01-filedir.md
+++ b/01-filedir.md
@@ -27,7 +27,7 @@ let's open a shell window:
 > ## Preparation Magic {.callout}
 >
 > If you type the command:
-> `PS1='$ '`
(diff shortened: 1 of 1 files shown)


gdevenyi:
What's your definition of persists? When I use the unchanged code, my prompt changes and remains changed.

kblin:
Ah, durn. I have PROMPT_COMMAND set to display the return code of my previous command. That overwrites any changes done to PS1. My bad, if I unset PROMPY_COMMAND, this works as it should.
Sorry for the noise.

PR #274, opened 2016-01-13
added exercise: 'Create thumbnails'  in 'Finding files'

btel:
as a part of requirements for the SWC Instructors training


gdevenyi:
Hi, thanks for your contribution, it's a very helpful example, however we can't add material which relies on tools that not all learners will have during a workshop, we don't check for imagemagick, so we can't expect them to have it.

PR #276, opened 2016-01-18
Replace `ls -s` with `ls -l`

### LINKED ISSUE #275: 01-filedir: "Exploring more ls arguments" challenge
Last year, I taught the shell and finished the first lesson with its last challenge:

> What does the command `ls` do when used with the `-s` and `-h` arguments?

While doing this I noticed two things that I had not anticipated:

1. `-s` does not simply display the size of the files but the "allocated size of each file, in blocks" which I think is a bit of an advanced topic and also will differ quite a bit between different file systems
2. `-h` does not change anything when used on OS X (i.e. with the BSD `ls` command), its documentation (on both Linux and OSX) states that it is for use "with -l", even though on Linux it *does* change the output for `-s`.

Given that the lessons generally stick to the basic commands and options that "work everywhere", I'd rather remove this challenge from the lesson (or maybe adapt it or add an explaining note -- but not sure the `-s` option is important enough for this...). Showing the use of `-l` would certainly be more interesting (since it shows the highly useful last modification date), but unfortunately this also prominently displays permissions and ownership information that are not covered anywhere in the novice lessons.

mstimberg:
As discussed in issue #275, this replaces `ls -s` with `ls -l` in the "Exploring more ls arguments" challenge. It also adds a little note about permissions/users/groups (which are not covered in the shell novice lesson) and hopefully encourages learners to explore (in line with the title of the challenge) :)

Closes #275

### DIFF
--- a/01-filedir.md
+++ b/01-filedir.md
@@ -707,8 +707,11 @@ what will `ls ../backup` display?
 
 > ## Exploring more `ls` arguments {.challenge}
 >
-> What does the command `ls` do when used with the `-s` and `-h`
-> arguments?
+> What does the command `ls` do when used with the `-l` and `-h` arguments?
+>
+> Some of its output is about properties that we do not cover in this lesson (such
+> as file permissions and ownership), but the rest should be useful
+> nevertheless.
 
 > ## Listing Recursively and By Time {.challenge}
 >
@@ -718,3 +721,4 @@ what will `ls ../backup` display?
 > most recently changed files or directories first.
 > In what order does ls -R -t display things? Hint: ls -l uses a long listing 
 > format to view timestamps.
+


gdevenyi:
I'm fine with the change but I think we can be a bit less patronizing to the learners about the output.

Can you refine your description to quickly describe the output and hint at documentation which explains it more fully?

mstimberg:
Thanks for having a look at this. The comment in the challenge was certainly not meant to be patronizing but I think I get what you mean. I struggle a bit to come up with a "hint at documentation" that works here (I don't think we want to link to an external website, do we?). The "shell-extras" lesson does explain permissions and such, but I got the impression that this is not a lesson we are referring to (it is not linked on the website, half of the topics are placeholders, etc.). Man pages have not been introduced at this part of the shell lesson (and the sole explanation it offers is "-l: use a long listing format"). Any suggestion for a documentation that we can refer to?

wking:
On Mon, Feb 01, 2016 at 01:54:56AM -0800, Marcel Stimberg wrote:
> … I don't think we want to link to an external website, do we? … Man
> pages have not been introduced at this part of the shell lesson (and
> the sole explanation it offers is "-l: use a long listing format") …

I'm not a maintainer, but would like to see lots of links to the POSIX
pages for each utility.  For example, [1] is POSIX on ‘ls’, and the
‘-l’ entry references the STDOUT section [2] which specifies the long
format field by field.

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/ls.html
[2]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/ls.html#tag_20_73_10

wking:
On Mon, Feb 01, 2016 at 09:17:49AM -0800, W. Trevor King wrote:
> I'm not a maintainer, but would like to see lots of links to the POSIX
> pages for each utility…

The POSIX utility docs are available on Linux [1] via ‘man 1p ls’,
etc. (where the ‘p’ section suffix gets you the POSIX page [2]).  But
I don't think that means we can't link to web version of the docs,
without needing to dig into the local ‘man’ command.

[1]: http://article.gmane.org/gmane.linux.man/4942
[2]: https://en.wikipedia.org/wiki/Man_page#Manual_sections

shwina:
@wking: Just a warning that the POSIX pages look a bit intimidating:

~~~
ls [-ikqrs] [-glno] [-A|-a] [-C|-m|-x|-1] \
       [-F|-p] [-H|-L] [-R|-d] [-S|-f|-t] [-c|-u] [file...]
~~~

If we're going to link to these, should we specifically point to the "options" section of the page?

mstimberg:
Sorry for not having followed up on this PR. I agree that the POSIX page is quite daunting for beginners. Also, it would be a bit odd to introduce a link here in an exercise, while the rest of the lesson does not contain any links to documentation. IMHO, the question whether to link to the POSIX documentation (or something more beginner-friendly) should be discussed in a wider context for this lesson.

For now, I just updated the text of the challenge to (hopefully) address @gdevenyi's concern about it being too patronizing.

shwina:
Thanks @mstimberg!

@wking or @gdevenyi: can you +1 this?

shwina:
@gdevenyi

mstimberg:
Hi, @gdevenyi /@shwina : this PR cannot be merged anymore without conflicts, so I'll update it. Do you prefer that I merge current `gh-pages` into this branch or should I rather re-base it and force-push (to get a "clean history")?

gdevenyi:
I prefer rebase and force push please :)

mstimberg:
> I prefer rebase and force push please :)

Ok, I updated it and squashed everything in a single commit -- feedback welcome!

gdevenyi:
Looks great, thanks!

PR #278, opened 2016-01-19
Added challenge exercise.

ericmjl:
SWC Exercise - Eric Ma.

Added a challenge exercise to show a different way of creating empty files.

### DIFF
--- a/02-create.md
+++ b/02-create.md
@@ -421,10 +421,12 @@ but it does find the copy in `thesis` that we didn't delete.
 
 > ## Creating Files in a Different Way {.challenge}
 > 
-> In this lesson, you learned how to create a text file using the nano command. Now, in the Unix (i.e. Mac OS/Linux) terminal, try the following command in your directory:
+> In this lesson, you learned how to create a text file using the `nano` text editor. Now, try the following command in your Desktop directory:
 > 
+> $ cd /path/to/Desktop  # change to your Desktop directory
 > $ touch my_file.txt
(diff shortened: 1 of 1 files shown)


02-create.md
```diff
@@ -418,3 +418,14 @@ but it does find the copy in `thesis` that we didn't delete.
 > The command `ls -t` lists things by time of last change,
 > with most recently changed files or directories first.
 > In what order does `ls -R -t` display things?
+
+> ## Creating Files in a Different Way {.challenge}
+> 
+> In this lesson, you learned how to create a text file using the nano command. Now, in the Unix (i.e. Mac OS/Linux) terminal, try the following command in your directory:
```
iglpdc: I suggest: 
- "the nano command" -> "`nano`", or maybe a text editor (some people use other editors), as everybody should be familair with it now.
- ~~"in the Unix (i.e. Mac OS/Linux) terminal"~~, this may be confusing for people using Git Bash and everyone knows now what the terminal is.
- "in your directory", which directory? :)

iglpdc:
@ericmjl I really like this exercise and it's something that I always do at some point during the lesson. One thing that you could do is to ask them to create the file in their Desktop while watching their Desktop with the GUI. Seeing the icon of the file showing up when they create the file reinforces the idea that the GUI and CLI are just two different ways of interacting with the (**same**) computer.

ericmjl:
@iglpdc: thanks for the comments!

PR #281, opened 2016-01-22
Adds challenge to check for correct number of arguments to the script.

rohitrawat:
Submitted by Rohit Rawat for instructor training.


gvwilson:
Thank you for your PR - we haven't introduced conditionals in the shell lecture because they have many traps for novices, but this would be a good addition to https://github.com/swcarpentry/shell-extras.  Can you please submit it there instead?

PR #282, opened 2016-01-22
SWC_IT_bash_exercise_04-loop_Natasha_Pavlovikj

npavlovikj:
Software Carpentry Instructor Training bash exercise using loops


gvwilson:
Thank you for the PR - I think this one covers the same material as others we now have.

PR #285, opened 2016-01-22
added Bonus exercise to one of the excercises

tomomis86:
This is a part of instructor training.


gvwilson:
Not substantial enough to incorporate.

PR #286, opened 2016-01-22
Added man pages challenge

miagao:
Adding a challenge: try man pages to show the documentation of the programs.


gdevenyi:
We can't use man pages because they're not installed for our windows learners unfortunately.

gvwilson:
Can we modify the challenge so that people on Windows (or other platforms) learn to use `man` via Google search or some other online resource?

PR #287, opened 2016-01-23
New exercise for instructor training pipefilter_bchapm5.md

bchapm5:
Sorry for the delay the existing exercises covered the material very well. It was difficult coming up with  something new and useful.


gdevenyi:
Thanks for the PR, you have provided an interesting example, but we need this placed within existing lessons and I think all these items have already been covered.

PR #288, opened 2016-01-23
Added challenge exercise for SWC Instructor Training requirement

darencard:
New challenge exercise and appropriate answers are included.


gvwilson:
Thank you for submitting this - it's a good exercise, but as @weaverbel points out, it's a bit long, and it covers some of the same material as others we have now merged.

PR #291, opened 2016-01-28
Changed the body of Nelle's loop to only select her original datafiles

gracecox:
The original *[AB].txt matches Nelle's original data files and any newly created stats-$filename files as well. This results in an infinite loop that creates stats-$filename, stats-stats-$filename stats-stats-stats-$filename etc.

### DIFF
--- a/04-loop.md
+++ b/04-loop.md
@@ -334,7 +334,7 @@ Using the left arrow key,
 Nelle backs up and changes the command `echo` to `goostats`:
 
 ~~~ {.bash}
-$ for datafile in *[AB].txt; do bash goostats $datafile stats-$datafile; done
+$ for datafile in N*[AB].txt; do bash goostats $datafile stats-$datafile; done
 ~~~
 
 When she presses enter,


gdevenyi:
Sorry, I'm not quite getting the problem here, the expansion of *[AB].txt is executed before goostats is run, so the list of files the loop runs over won't get larger.

If someone were to run the loop more than once that would be a problem.

Can you explain why you think this change is needed a bit more? Thanks.

gracecox:
The loop is run more than once during the SWC lessons; once in the loops lesson and then again in the following lesson when Nelle creates her do-stats.sh shell script. The premise is that she made a mistake in her initial analysis and so uses a shell script to repeat it (but she doesn't remove the original set of stats- files first and so would get a set of stats-stats- files instead of the desired output). 

Since we get this undesired output when following the lessons exactly, I suggested a modification... Also, when I taught this lesson, several learners did run the script several times (with different parameters) and got confused about the files produced.

In general, I think it is more robust to write scripts so that they can be rerun without having to do other things first (i.e. remove a bunch of files). Does this make more sense? Thanks

gdevenyi:
Okay, that sounds reasonable, however, all the loops before and afterwards look different. Can you update your PR to fix all instances, so the instructor doesn't have to explain why this one is different. Thanks.

gdevenyi:
No response to request to uniformize changes. Please resubmit if updated.

PR #292, opened 2016-01-29
Replace cat with nano and clarify

danwiththeplan:
In several places "cat" is called where a file edit using "nano" or some other editor is required. This replaces "cat" with "nano" where appropriate and provides some clarifying text.


gdevenyi:
Excellent catch!

PR #294, opened 2016-01-30
Add convert image challenge

asiaszmek:


asiaszmek:
I am sorry for doing the assignment so late.

gdevenyi:
While this is a good idea, its highly unlikely that our learners will have image/graphicsmagick installed on their systems during a workshop, so I don't think this an appropriate assignment.

Can you try out a challenge which relies on the standard programs we can guarantee learners will have? See the SWC install instructions.

PR #298, opened 2016-02-01
format fix  in 05-script.md

dbrunson:
Added a blank line between text and code block so it renders correctly in the html version.


shwina:
Thanks, @dbrunson

PR #300, opened 2016-02-03
added RNA-Seq exercise

tmogrady:
Exercise for Instructor Training Fall 2015

### DIFF
--- a/06-find.md
+++ b/06-find.md
@@ -492,3 +492,27 @@ about them."
 > of the four sisters is mentioned?  Hint: one solution might employ
 > the commands `grep` and `wc` and a `|`, while another might utilize
 > `grep` options.
+
+> ## Counting unmapped RNA-seq reads {.challenge}
+>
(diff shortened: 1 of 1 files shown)


06-find.md
```diff
@@ -492,3 +492,27 @@ about them."
...
+> Put the following in the correct order and add pipes to create a
+> command that will return the number of reads in the file 
+> `alignment.sam` that are unmapped:
+>
+> * `alignment.sam`
+> * `wc -l`
+> * `cut -f 3`
+> * `grep "*"`
+>
+> Bonus: Can you get the same information using fewer commands> What
```
gdevenyi: Can you please fix up the errant ">"

gvwilson:
Please also provide some sample data so that users can test their solution.

gdevenyi:
Thanks for your contribution. We try and keep our examples domain-agnostic. Without example data we are unable to use this with learners.

PR #301, opened 2016-02-03
shell script exercise addition

dbrunson:
This pull request is for instructor training.  It's about launching scripts with `./` instead of `bash`.  It requires chmod to make it work, but so many online tutorials use it that I thought it'd help to have a little exposure here.

### DIFF
--- a/05-script.md
+++ b/05-script.md
@@ -460,3 +460,27 @@ Of course, this introduces another tradeoff between flexibility and complexity.
 > # Script 3
 > echo $@.dat
 > ~~~
+
+
+> ## Making your script executable {.challenge}
+> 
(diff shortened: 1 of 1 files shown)


dbrunson:
Just saw this about not using shebang in novice lessons: https://github.com/swcarpentry/shell-novice/pull/307  
Many of the learners that I've been around see this on other tutorials and ask about it, which is what made me think of adding it.  But it sounds like it should remain one of those "answer only if asked" issues. :)

gvwilson:
Unfortunately, `#!` only works with `chmod`, and we've avoided talking about permissions until the shell-extras lesson https://github.com/swcarpentry/shell-extras - can you please submit this one there instead?

PR #302, opened 2016-02-04
Clarify pipe chaining

iamc:
Instructor training exercise.

Add an intermediate step in pipe concatenation so it is more explicit for the uninitiated.


gdevenyi:
Very nice clarification.

PR #305, opened 2016-02-08
revise command from 'pipeline reading' exercise (fixes #280)

### LINKED ISSUE #280: "find pipeline reading comprehension" exercise
This one-liner from exercise "`find` pipeline reading comprehension" in "Finding things" lesson:

`find . -name '*.dat' | wc -l | sort -n`

 will always print a single line (the number of .dat files in the current dir and its subdirs), so sort at the end of the pipeline is not necessary.  Since the goal of the exercise is to write an explanatory comment, it may be confusing.

Perhaps this command would make more sense:

`wc -l $(find . -name '*.dat')  | sort -n`

btel:
The original command always print a single line (the number of .dat files in the current dir and its subdirs), so sort at the end of the pipeline is not necessary. This command finds all *.dat files and prints them in the order of increasing number of lines. For more info see issue #280.

### DIFF
--- a/06-find.md
+++ b/06-find.md
@@ -462,7 +462,7 @@ about them."
 > Write a short explanatory comment for the following shell script:
 >
 > ~~~ {.bash}
-> find . -name '*.dat' | wc -l | sort -n
+> wc -l $(find . -name '*.dat') | sort -n
 > ~~~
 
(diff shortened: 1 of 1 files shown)


gdevenyi:
Looks good, thanks!

PR #306, opened 2016-02-09
Adding something about tar

adgdt:
Instructor training exercise.

I believe that tar is a useful and easy tool, and deserves to be in the basics.


shwina:
Thanks, @adgdt. @ChristinaLK, do you think that with a little work, this could belong in [shell-extras](http://swcarpentry.github.io/shell-extras/)?

adgdt:
Sure, it was just an idea. I agree than shell-extras could be a good place for it

gvwilson:
Agreed - let's move this to shell-extras.

PR #307, opened 2016-02-09
instructors.md: We really, really don't want to talk about #!

wking:
I found the Mascheck link via [this post][1].  For our
mostly-OS-agnostic lesson, I think we want to stay far, far away from
anything that uses kernel and/or shell magic to invoke the script with
a given interpreter.  `bash example.sh` is a bit more typing and/or
two auto-completes, but it is a lot easier to explain to folks hitting
the shell for the first time.

[1]: https://blogs.gentoo.org/mgorny/2016/02/08/a-quick-note-on-portable-shebangs/

### DIFF
--- a/instructors.md
+++ b/instructors.md
@@ -162,8 +162,10 @@ on the front page of the website for more details.
     this can be covered instead using the online lessons as guidelines.
     These limitations also have follow-on consequences:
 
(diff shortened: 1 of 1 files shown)


gvwilson:
+100

PR #312, opened 2016-02-17
Added challenge to list only directories

philrosenfield:
synthesize material from lesson 2 and this lesson.


gvwilson:
Thanks for the PR - it pulls the parts of the lesson together well, but I think it covers material that is also now covered by other exercises.

Issue #289, opened 2016-01-27
Home directory items inconsistent from L01 to L02


mjones01:
Caveat: I am going through the lesson on my own, not as part of a workshop.  I acknowledge this isn't how the lessons are designed to be used.  If this issue doesn't cause problems in workshops then please disregard and close the issue.  

In the second topic of The UNIX Shell lesson series, Files and Directories  (http://swcarpentry.github.io/shell-novice/01-filedir.html), Nelle's home directory (/Users/nelle/) contains subdirectories ~/Desktop/data-shell/  that contains other sub-directories and folders.  In topic 3, Creating Things (http://swcarpentry.github.io/shell-novice/02-create.html), the folders/files are in the home directory (/Users/nelle/) instead of in Desktop/data-shell.   I found this confusing when going from one lesson to the next.  However, it is arguably a minor issue.  

To make consistent, in L03 should start with /Users/nelle/Desktop/data-shell instead of just the home directory (/Users/nelle/Desktop/data-shell).

shwina:
Thanks @mjones01 - nice catch! Here's the directory structure (I have this placed in my Desktop):

~~~
/home/shwina/Desktop

└── data-shell
    ├── creatures
    ├── data
    │   ├── elements
    │   └── pdb
    ├── Desktop
    ├── molecules
    ├── north-pacific-gyre
    │   └── 2012-07-03
    └── writing
        ├── data
        ├── old
        ├── thesis
        └── tools
            └── old
~~~

ChristinaLK:
Once the above PR is merged, maybe open/create a new issue for removing the empty Desktop dir from data-shell?  That's a bigger project.

mjones01:
This Issue appears to have been fixed with PR #314.  Close the issue?

shwina:
Yes please. Thanks!

PR #314, opened 2016-02-17
fixing intro material to match new data directory

ChristinaLK:
addresses part of #289

### DIFF
--- a/02-create.md
+++ b/02-create.md
@@ -13,15 +13,14 @@ minutes: 15
 
 We now know how to explore files and directories,
 but how do we create them in the first place?
-Let's go back to Nelle's home directory,
-`/Users/nelle`,
+Let's go back to our `data-shell` directory on the Desktop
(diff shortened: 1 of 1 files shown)


shwina:
Thanks @ChristinaLK

JohnRMoreau:
I think this is the right issue for my comment. If not, please point me in the right direction. I think this may be related to #289.
On L01, under "Nelle’s Pipeline: Organizing Files", the text says that _from the home directory_ the user can run the following command to look at the files available:
`$ ls north-pacific-gyre/2012-07-03/`
which does not work from the home directory. Instead, if we start from
`/Users/nelle/Desktop/data-shell/data`
we can run the first command. Part of the confusion may be that the last cd command given takes the user to that data directory. I don't have time to submit a PR right now, so I'm leaving this note to remind myself to correct the error and submit the PR.

PR #318, opened 2016-02-20
Tilde expansion in the path

Mahdisadjadi:
One might think "~" expands literally in the path so `here/there/~/elsewhere` will be `here/there/Users/nelle/elsewhere`.

### DIFF
--- a/01-filedir.md
+++ b/01-filedir.md
@@ -432,7 +432,7 @@ Run `pwd` and `ls -F` to ensure that we're in the directory we expect.
 > mean "the current user's home directory". For example, if Nelle's home
 > directory is `/Users/nelle`, then `~/data` is equivalent to
 > `/Users/nelle/data`. This only works if it is the first character in the
-> path: `here/there/~/elsewhere` is *not* `/Users/nelle/elsewhere`. 
+> path: `here/there/~/elsewhere` is *not* `here/there//Users/nelle/elsewhere`. 
 > 
 > Another shortcut is the `-` (dash) character.  `cd` will translate `-` into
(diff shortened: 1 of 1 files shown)


01-filedir.md
```diff
@@ -432,7 +432,7 @@ Run `pwd` and `ls -F` to ensure that we're in the directory we expect.
 > mean "the current user's home directory". For example, if Nelle's home
 > directory is `/Users/nelle`, then `~/data` is equivalent to
 > `/Users/nelle/data`. This only works if it is the first character in the
-> path: `here/there/~/elsewhere` is *not* `/Users/nelle/elsewhere`. 
+> path: `here/there/~/elsewhere` is *not* `here/there//Users/nelle/elsewhere`.
```
shwina: Can we use the form `here/there/Users/nelle/elsewhere` instead? At this point, it's not clear to learners if `//` is a legal separator.

Mahdisadjadi: In the previous line, `~/data` expands to `/Users/nelle/data`, I think it's not consistent to drop `//`. Can we explain that this is one reason `~` does expand in the middle of path?

shwina: Yes - but I think it's OK to be a little sloppy here as we really just want to stress on the idea. We'd have to add a side note that `//` is a valid separator, but that distracts from the main point that `~` doesn't expand to anything in the middle of a path.

Mahdisadjadi: Yes that's right, I agree. I'll edit the change. Thanks.

shwina:
Thanks @Mahdisadjadi

Mahdisadjadi:
:thumbsup: @shwina

PR #319, opened 2016-02-21
Added challenge question to the find section

avishek-r-kumar:
Added a challenge question to loop through directories beginning with *t*. This tests the 
student's ability to use a `for loop`, `grep`, `find` and `pwd`.


gvwilson:
Thanks for this PR - it's a useful skill, and ties several ideas together, but there's no sense of *why* someone would want to do this.  Can you please re-submit with more motivation?

PR #320, opened 2016-02-21
Added challenge: using find to delete newly created files

markusuwe:
I added a challenge that introduces the new find options: -delete and -mmin. These are very handy for cleaning up folders; they are kind of a roll-back mechanism for the shell.

### DIFF
--- a/06-find.md
+++ b/06-find.md
@@ -496,5 +496,5 @@ about them."
 > ## Cleaning up {.challenge}
 >
 > During the course of this lesson you created a number of files.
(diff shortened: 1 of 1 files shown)


mstimberg:
I certainly agree that these options are useful! Introducing them in a challenge without further explanation how to use them or where to find further information will not work well in practice though, I am afraid.

gdevenyi:
I'll have to agree with @mstimberg, we should attempt to use existing features already taught for challenges.

PR #326, opened 2016-02-25
File changed to add exercises for SWC training

hoytpr:
May have additional comments or suggestions, but the exercises are the important part for the instructor training. 
02/25/2016
Peter Hoyt


hoytpr:
Used for instructor training purposes.

DanielBaird:
I think the section around line 154, that reads:

    The Mummy likes to store his files in a directory called `/Users/imhotep`,
    while the Wolfman just calls his file direcotry `/Users/larry`,

...needs rephrasing to avoid the implication that a user's home dir can be wherever the user likes, when it's tightly constrained by the OS and username (and fix the `direcotry` typo).

hoytpr:
I agree Daniel, this needs to be very clear that Nelle's colleagues have 
well... actual-organismal names, but their **usernames** (on the 
computer) are the same as the name of their home directory.

Peter

gvwilson:
Thanks for the PR - please start a fresh PR that is based on our recent changes, and that includes *only* the .md (Markdown) file, not the HTML.

PR #328, opened 2016-02-29
Changed typing to pressing

vfaghirh:
Typing can be misleading, because in order to kill a running job one does not type Control. I changed typing Control-C to pressing ^C (^ means Control)

### DIFF
--- a/03-pipefilter.md
+++ b/03-pipefilter.md
@@ -57,7 +57,8 @@ $ wc *.pdb
 > matches filenames that begin with the letter 'p'.
 >
 > `?` is also a wildcard, but it only matches a single character. This
-> means that `p?.pdb` matches `pi.pdb` or `p5.pdb`, but not `propane.pdb`.
+> means that `p?.pdb` matches `pi.pdb` or `p5.pdb` (if we had these files 
+in the molecules directory), but not `propane.pdb`.
 > We can use any number of wildcards at a time: for example, `p*.p?*`
 > matches anything that starts with a 'p' and ends with '.', 'p', and at
 > least one more character (since the '?' has to match one character, and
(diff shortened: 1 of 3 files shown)


btel:
Perhaps it would be better to say <Control>-C (where <Control> means pressing and holding the Control key).

btel:
Discussion actually lists all possible codes for Control key. Whatever you decide please not that there are more than one mentions of Control in the lesson. You need to change all for consistency.

vfaghirh:
@btel Yes I am aware of that, thank you.

gvwilson:
Thanks for the PR - I realize `^C` is common Unix convention for control characters, but I've found in the past that it's better to spell it out to avoid confusion with ASCIIfied accent characters.  I like replacing "typing" with "pressing" - do you want to resubmit with that, or should I go ahead and make the change?

vfaghirh:
@gvwilson : Greg, I agree that for a novice it would be less confusing to see the word Control spelled out and I think mentioning once or twice in the lessons that ^C is the Unix convention for control character would be sufficient. I also have noticed that there is not a consistency in our lessons for referring to control key. Sometimes it's referred to as Ctrl, sometimes as Control and sometimes as ^C. So I suggest that we make all of them consistent. I can submit a PR for all of them if you agree.

gvwilson:
Consistency would be great - how about we split the difference and call 
it `Ctrl` ?

vfaghirh:
@gvwilson : Sounds good. I am going to read through all the shell-novice lessons today and change all the ^C and Controls to 'Ctrl'. Could you change the Typing to Pressing?

btel:
@vfaghirh It seems to me this PR has been superseded by #400 and #401. Should we close it?

PR #331, opened 2016-03-02
New exercise to assist debugging scripts - trainer certification

ozej8y:


weaverbel:
I should have known this would be good - I liked it before I realised it was yours. I liked Daniel's too! Go JCU.

05-script.md
```diff
@@ -440,3 +440,34 @@ Of course, this introduces another tradeoff between flexibility and complexity.
 > # Script 3
 > echo $@.dat
 > ~~~
+
+> ## Ooops my large script has a bug. How can I easily find the problem? {.challenge}
+>
+> You have written a large script with many lines, and it's not working properly.
+> How can I easily identify the problem ?
```
mstimberg: A minor comment here: this uses "I" and "you" a bit inconsistently: "you" have written the script but "I" identify the problem. I think it is perfectly fine in the title, but here I'd stay with "you". Maybe also consider rephrasing it to "What can you do to help you identify the problem?" -- I am not sure interpreting the output with `bash -x` is necessarily "easy" for the novice learner.

ozej8y: I tweaked the text a bit noting your comments of 'You' and 'I'. I don't mind if the title changes. Interpreting the 'output' might not be 'easy' but at least someone can look at it, hopefully spot the error and they have something to provide them with an idea as to where they went wrong.

mstimberg: > Interpreting the 'output' might not be 'easy' but at least someone can look at it, hopefully spot the error and they have something to provide them with an idea as to where they went wrong.

Sorry if this came across wrong -- I do not argue with the usefulness of `bash -x`, it is certainly easier to debug the script with it! What I meant to say was that by saying "How can you easily identify the problem" you are implying that with `bash -x` the learner will "easily identify the problem". It might be frustrating for the learner if that does not quite work for them after they were told it was "easy".
Not a major point and others might disagree with my view here, so don't worry about it too much ;)

ozej8y: That's OK Marcel. I wasn't offended, the comments were aimed at improving the language. :)

PR #332, opened 2016-03-03
some text changed.

oiertwo:
might or might not seemed something like: your are stupid or maybe not, or who knows?... I think its better now.

### DIFF
--- a/02-create.md
+++ b/02-create.md
@@ -38,7 +38,7 @@ Let's create a new directory called `thesis` using the command `mkdir thesis`
 $ mkdir thesis
 ~~~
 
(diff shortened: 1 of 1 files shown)


DanielBaird:
The exclamation point `!` doesn't need a period `.` after it.

Not correct: ` easy to remember!.`
Correct: ` easy to remember!`

btel:
It's a good suggestion. How about showing where the acronym comes from, for example, by capitalising first letters: "MaKe DIRectory".

gvwilson:
Closing with 96cab70 instead.

PR #335, opened 2016-03-04
fix typo later -> latter

DanielBaird:
Just a typo fix.  I originally intended to expand a smidge on the use of "up" to mean toward-the-root-in-the-directory-tree, and "back" to mean back-in-time-to-the-dir-I-was-in-just-before, but I couldn't find a phrasing clear enough.

Later I'll try and add something about up meaning toward the root, which is so powerful a metaphor it's hard to even think of it as a metaphor -- even though it clashes with the tree/branching metaphor that gives you the word "root".  English is fun isn't it..


shwina:
I thing that "up" and "back" in this context are familiar to most users from working on some sort of file manager on their OS. However, I'm not sure if this is the case for non-English OS users. It's an interesting point.

PR #339, opened 2016-03-04
Small fixes to 05-script.md

nsoranzo:

PR #341, opened 2016-03-07
Corrects language in exercise: "ending in" > "matches"

DanielBaird:
Exercise referred to files "ending in" `*.pdb`; this changes both occurrences of that phrase to the more accurate term "matching".

I support colloquial idiom in lessons, but I think this is incorrect enough that it will mislead learners.  The term "matching" has been introduced in the previous lesson to describe files selected by wildcards.

### DIFF
--- a/05-script.md
+++ b/05-script.md
@@ -363,9 +363,9 @@ Of course, this introduces another tradeoff between flexibility and complexity.
 >
 > Which of the following outputs would you expect to see?
 >
-> 1. All of the lines between the first and the last lines of each file ending in `*.pdb`
+> 1. All of the lines between the first and the last lines of each file matching `*.pdb`
 >    in the molecules directory
-> 2. The first and the last line of each file ending in `*.pdb` in the molecules directory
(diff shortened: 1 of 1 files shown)


iglpdc:
Good catch! I agree that "ending in `*.pdb`" is incorrect and might lead to confusion. An alternative to keep a more intuitive language could just use "ending in `.pdb`".

DanielBaird:
#339 also fixes this problem, and uses just that technique @iglpdc :)  I considered going that way, because it shows the learner an nice example of "english-izing" a wildcard pattern, but for the written content, I ended up sticking to using the string that directly matches the command.

In actual teaching, I'd definitely recommend using the expression "ending in dot P D B" when talking aloud.

If people like the "ending in" construction better, we can close this pull-req and accept #339.

shwina:
I'm in favour of #339 (english-ising)! Shall we go ahead and close this PR?

DanielBaird:
Yep I like it too. Closing..

PR #343, opened 2016-03-07
Added exercise

milhans:
Added exercise to demonstrate removing files with wildcards and care that must be taken when using * with rm.


mstimberg:
Nice idea and an important lesson about `*` and incorrect spaces!
Two small comments:
> You want to delete your processed data files for archiving.

I get what you mean but "delete files for archiving" sounds a bit odd, maybe you can phrase this in a different way
> `rmdir *.txt`

The `rmdir` command has recently been removed from the lesson (#384) so it should probably no longer be used here.

Issue #344, opened 2016-03-08
Using expr in 04-loop.md ...


plijnzaad:
In exercise 04-loop.md, I came across the ancient `expr` command. Noone uses that anymore these days. Would it not be better to replace it with `$[ 3 + 5 ]` ? (I will submit a pull request to this effect later on)

shwina:
Thanks @plijnzaad 

I'm no expert on portable bash scripting, but based on the discussion [here](http://stackoverflow.com/questions/2415724/bash-arithmetic-expression-vs-arithmetic-expression), the `$((3+5))` form is preferable.

gvwilson:
I still use `expr` to avoid yet another layer of parentheses in my 
expressions - as with teaching people to run scripts with `bash 
myscript.sh` instead of changing permissions and adding a `#!`, it's not 
how proficient users do things, but it saves us having to explain 
another layer of shell-ism to novices.

jttkim:
Adding to @gvwilson's comment I'd also say that if a script does so much / so complex arithmetics that using `expr ...` vs. `$[ ... ]` vs. `$(( ... ))` makes any real difference, it quite certainly wants to be written in another language that provides proper numerical types.

Variable values in the shell are basically always strings, using bash-isms don't change that. bash-isms been brought in at a time when the overhead of starting an external process was much more of a concern that it is today.

plijnzaad:
@Ashwin: good call, and @Jan: I wholeheartedly agree. See also my pull request.

plijnzaad:
(This could be considered part of my instructor training/certification 'home work', see also https://github.com/swcarpentry/shell-novice/issues/349)

gdevenyi:
Looks like this has been resolved.

jan:
@plijnzaad Please don't use "@Jan" unless you want to talk to me. ;-)

PR #345, opened 2016-03-08
alternative nested loop example

plijnzaad:
Nested for-loop example that does not use expr, which I also think is a bit less contrived than the arithmetic example. The slight disadvantage is that it  unceremoniously strings two variables together. Then again, it's a challenge, and students can simply try it out. 
See https://github.com/swcarpentry/shell-novice/issues/344

### DIFF
--- a/04-loop.md
+++ b/04-loop.md
@@ -499,26 +499,19 @@ so she decides to get some coffee and catch up on her reading.
 > done
 > ~~~
 
-> ## Nested Loops and Command-Line Expressions {.challenge}
+> ## Nested Loops {.challenge}
 >
-> The `expr` does simple arithmetic using command-line parameters:
-> 
+> Suppose we want to set up up a directory structure to organize
+> some experiments measuring the growth rate under different sugar
+> types *and* different temperatures.  What would be the
+> result of the following code:
+>
 > ~~~
-> $ expr 3 + 5
-> 8
-> $ expr 30 / 5 - 2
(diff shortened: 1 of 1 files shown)


04-loop.md
```diff
@@ -499,26 +499,19 @@ so she decides to get some coffee and catch up on her reading.
...
-> 
 > ~~~
-> for left in 2 3
+> for sugar in fructose glucose sucrose
 > do
->     for right in $left
+>     for temperature in 25 30 37 40
 >     do
->         expr $left + $right
+>         mkdir $sugar$temperature
```
shwina: May I suggest $sugar-$temperature instead?

plijnzaad:
@Ashwin: I would be fine with $sugar-$temperature, but considered that novices might find it confusing as it would seem to imply arithmetic. The alternatives would be '_' (which requires the ${sugar} construct, '.' (which is slightly unusual for directory names) or '=' or '+' (which are also arithmetically clumsy). So I don't kno :-)

shwina:
That's a good point, but we haven't introduced shell arithmetic at this point, so I don't see this as a problem. I'd keep the hyphen as an example to learners of how to name their directories. If it's OK with you, I'll merge and make the change?

plijnzaad:
Sure, thanks!
Op 8 mrt. 2016 19:10 schreef "Ashwin Srinath" <notifications@github.com>:

PR #349, opened 2016-03-09
Backtick

plijnzaad:
I decided to create a lesson on the use of backticks (nowadays $(...) ). I think it is useful, especially in the context of loops. But it may be a bit too much for the complete novice. If that is the case, maybe it can be incorporated elsewhere, e.g. in the genomics/cloud computing area where people run into this kind of thing readily. 

During the writing of this lesson, I realized that the students have not been introduced to the concept of shell variables; they only see them in passing as loop variables. May be something to consider at some point?

Two more things: I would like to refer to previous lessons, is there syntax for that? I now simply say 'in the *Loops*' topic, which may be a bit brittle ...

Secondly, I provide the answers to the challenges in HTML comments (), I hope that is OK (at least for instructors, right)

Thirdly :-), I'd like to submit this as part of my swcarpentry/data carpentry instructor training homework
(actually just like https://github.com/swcarpentry/shell-novice/issues/344 and https://github.com/swcarpentry/shell-novice/pull/345). I hope that is OK. Cheers,

Philip


gvwilson:
I think this would be great to add to shell-extras (another swcarpentry 
repo with "extra" stuff).

shwina:
Can I close this in favour of the PR at `shell-extras`?

plijnzaad:
certainly. Cheers, P

PR #352, opened 2016-03-14
Incorrect file path in 01-filedir

hindsjj:
Nelle's home directory is /Users/nelle/, so to see her files she should type:
$  ls Desktop/data-shell/north-pacific-gyre/2012-07-03/

### DIFF
--- a/01-filedir.md
+++ b/01-filedir.md
@@ -480,7 +480,7 @@ If she is in her home directory,
 Nelle can see what files she has using the command:
 
 ~~~ {.bash}
-$ ls north-pacific-gyre/2012-07-03/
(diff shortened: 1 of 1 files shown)


gdevenyi:
Seems some inconsistencies have crept into the filesystem description, if you look at https://github.com/swcarpentry/shell-novice/tree/gh-pages/filesystem there is nothing in Desktop, which is consistent with most but not all of the lesson.

Can you do an audit of the current wording with reference to the current file layout? Your current change fixes the wording, but makes it inconsistent with the actual filesystem.

Please open a new PR to address this. Thanks.

PR #356, opened 2016-03-24
Removed negative comments about programmers and references to peoples' ages

mikej888:


iglpdc:
+1, it looks better than before.

ChristinaLK:
+1 agreed!  Good catch.

shwina:
I'll miss the joke about "wit among programmers", but I completely agree with getting rid of "crusty old Unix programmers". :)

PR #358, opened 2016-03-24
Added .git and .ssh as other . files users might see to Other Hidden Files in 01-filedir.md

mikej888:


01-filedir.md
```diff
@@ -344,11 +344,12 @@ but we'll see some uses for it soon.
 
 > ## Other Hidden Files {.callout}
 > 
-> In addition to the hidden directories `..` and `.`, you'll also see a file 
+> In addition to the hidden directories `..` and `.`, you may also see a file 
 > called `.bash_profile`. This file usually 
 > contains settings to customize the shell.  There may also be similar files called 
-> `.bashrc` or `.bash_login` in your own home directory. The `.` prefix is 
-> used to prevent these 
+> `.bashrc` or `.bash_login` in your own home directory.
```
shwina: Do we really want to list all of these? None of these things, i.e., "git", "ssh", or "customizing your shell" make any sense to learners at this point.

So we want to say something like, "You may also see other files and directories beginning with a dot. These are usually files and folders that are used to configure different programs on your computer. The `.` prefix ... "

shwina:
Thanks @mikej888

PR #359, opened 2016-03-25
Added 'Same symbols, different meanings' callout.

mikej888:
Explains distinction between > and $ used as prompts and used for output redirection or accessing variable values, depending on whether they're printed by the shell or typed by the user.

### DIFF
--- a/04-loop.md
+++ b/04-loop.md
@@ -100,9 +100,11 @@ so this loop prints out the first three lines of each data file in turn.
 > used to redirect output.
 > Similarly, `$` is used as a shell prompt, but, as we saw earler,
 > it is also used to ask the shell to get the value of a variable.
-> If the shell prints `>` or `$` then it expects you to type something,
+>
+> If the *shell* prints `>` or `$` then it expects you to type something,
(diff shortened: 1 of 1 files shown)


04-loop.md
```diff
@@ -94,6 +94,17 @@ so this loop prints out the first three lines of each data file in turn.
...
 
+> ## Same symbols, different meanings {.callout}
+>
+> Here we see `>` being used a shell prompt, whereas `>` is also
+> used to redirect output.
+> Similarly, `$` is used as a shell prompt, but, as we saw earler,
+> it is also used to ask the shell to get the value of a variable.
+> If the shell prints `>` or `$` then it expects you to type something,
+> and the symbol is a prompt.
+> If you type `>` or `$` yourself, it is an instruction from you that
```
shwina: Do you think emphasis on the words *shell* and *you* may help here?

> If the *shell* prints ... 
> If *you* type...

mikej888: Good idea. Commited to this branch as d233525385551c57b9a46d420dc70f297a097e3b.

PR #360, opened 2016-03-25
Added 'man and Git for Windows' callout recommending use of Google search since 'man' command isn't supported.

mikej888:


06-find.md
```diff
@@ -160,6 +160,14 @@ To find out what they are, we can type `man grep`.
 it prints a description of a command and its options,
 and (if you're lucky) provides a few examples of how to use it.
 
+> ## `man` and Git for Windows {.callout}
+>
+> The bash shell provided by Git for Windows does not
+> support the `man` command. Googling for
+> `unix man page COMMAND` (e.g. `unix man page grep`)
+> provides links to numerous copies of the Unix manual
+> pages online.
```
shwina: What about `--help` for quicker/offline reference? Works for most commands on Windows, but I haven't tested personally..

mikej888: I've updated this page and reference.md with information on the --help flag and how it is supported by many bash commands, and programs people have written.

shwina:
Can we point to a specific resource(s) instead of just saying "Google this"? For example, http://www.gnu.org/software/grep/manual/

gdevenyi:
Looks great thanks!

PR #365, opened 2016-03-30
00-changes: Add an example of automating large tasks with Bash.

andre-geldenhuis:
Sometimes when learning bash, students only really realize that is
is a really powerful tool towards the end of the lesson.  Adding an
initial example might help hook some of them earlier in the lesson.

### DIFF
--- a/00-intro.md
+++ b/00-intro.md
@@ -75,6 +75,22 @@ In addition, the command line is often the easiest way to interact with remote m
 Familiarity with the shell is near essential to run a variety of specialized tools and resources including high-performance computing systems. As clusters and cloud computing systems become more popular for scientific data crunching,
 being able to interact with them is becoming a necessary skill. We can build on the command-line skills covered here to tackle a wide range of scientific questions and computational challenges.
 
+## A quick example to demonstrate automation with shell scripting.
+
+Shell scripting is a really powerful tool to automate tasks.  Imagine you had 100 000 microscopy slides that you needed to resize to put on the web. We could use Imagemagick, an image tool with a command line interface, to resize the images.
+
+~~~ {.bash}
+$ convert input.jpg -resize 800×600 output.jpg
+~~~
+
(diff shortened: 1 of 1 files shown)


iamc:
Beware of the wrong commit message:
 
> 00-changes: Add an example of automating large tasks in **git**

andre-geldenhuis:
Ah! Sorry, had git on the brain, corrected.

shwina:
Thanks, @andre-geldenhuis. Note that ImageMagick is not installed by default on most systems - we should be careful about what we use as examples when teaching.

andre-geldenhuis:
@shwina  yeah, that is a really good point.  Perhaps a better place might be in the instructors guide as a suggestion?  Something instructors demo to show off automating something in bash before getting going with the lesson?

raynamharris:
Hmm. @andre-geldenhuis I'm not sure I agree that one needs to use a coding demonstration to convey the utility of bash. This two-lines of code example could be very intimidating to the novices in the room because it introduces many new pieces of code at once without an explanation of what they do or how they work. Additionally, there is the confound of the non-standard program which means that students cannot follow along.

In my opinion, this type of example would be better shown after/during the for loop lesson to show a practical example of for loops in action.

andre-geldenhuis:
@raynamharris yeah I definitely take the point on the non-standard program.  

I guess my attempt is to show a simple example of the power of automation. We've only run one swc here so far, but many of out attendees have never even heard of bash before, they seemed to be attending mostly for the R session.  I just want to try and engage them - maybe inspire them with the kinds of things you can do with bash, before getting into the lesson proper.

Like I said though, I'm not sure it is the best approach.  I would hope you could mitigate people feeling intimidated by the code by explaining they are not expected to understand it at the start of the lesson - it is just a quick demo to show the kinds of things you can do quite easily.

gdevenyi:
Thanks for your submission, but we are unable to integrate lessons requiring external software not already included in the existing SWC standard install.

PR #372, opened 2016-05-19
Two new challenges: Loops and Scripts (instructor training)

tibbben:
I have added two new challenges to the loops and scripts lessons that use the curly brackets {} to introduce the idea of substitution within variable contents. The challenge in the Loops lesson builds on using wildcards, and the challenge in the Scripts lesson introduces the idea of 'slicing' (similar to the R module).


gdevenyi:
Please break these up into seperate PRs for the loop and script lessons.

PR #373, opened 2016-05-19
new challenge in Loop lesson

tibbben:
A challenge has been added to the Loop lesson that uses curly brackets {} to make substitutions within loop that outputs modified filenames. This exercise builds on using wildcards with curly brackets on string variables.


gvwilson:
Filename substitution is very useful, but is a pretty advanced topic for our crowd - can you please re-submit this one against the https://github.com/swcarpentry/shell-extras repository? That seems like a better home for it.

PR #374, opened 2016-05-19
added challenge to Script lesson

tibbben:
Added a new challenge in the Scripts lesson that uses curly brackets {} to extract the last parameter in the parameter list. This builds upon the curly bracket challenge added to the Loop lesson (see my previous pull request) and helps set the stage for 'slicing' in the R module.

### DIFF
--- a/04-loop.md
+++ b/04-loop.md
@@ -538,3 +538,43 @@ so she decides to get some coffee and catch up on her reading.
 >     done
 > done
 > ~~~
+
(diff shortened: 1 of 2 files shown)


gdevenyi:
Hi,

Thanks for your contribution, the array slicing material is indeed very helpful in bash, but we don't teach it and adding it to an existing lesson as a challenge is a bit too much for learners.

Perhaps you could re-spin this against shell-extras?

PR #376, opened 2016-05-31
Marconi Sergio Challenge

MarconiS:
solution: 
1. cd ./writing # supposing to be in data-shell/
2. nano haiku.sh
3. for line in Yesterday Today Software; do grep $line haiku.txt >> haiku_software.txt; done
4. bash haiku.sh; cat haiku_software.txt
5. haiku_software.txt would contain 4 lines: the three of the selected haiku, plus Today is a beautiful day (placed in the line after "Today it is not working", since it is called by the second argument of the for cycle)
6. find . -name 'haiku.txt'


gvwilson:
Can @shwina please decide whether or not to merge?

gcapes:
Hi @MarconiS and thanks for the PR.

My feeling is that the example is rather too contrived, I'm afraid.
While of course you could achieve the outcome to this exercise using `grep`, I think an easier solution would be to use `head` to extract the three lines required, and I wouldn't be able to justify why the learner should use `grep` instead.

With 6 exercises in the episode already, I think any extra would need to have a strong motivation.

Thanks again for the contribution.

PR #378, opened 2016-06-06
Rename file system lessons

Ry4an:
Trivial PR for #199 that renames lessons one and two to reflect their focus on navigation and
manipulation, respectively.

Issue #199, opened 2015-06-14
change title of 02-create.md?


ChristinaLK:
from #140, have a more general title as this lesson covers deleting and copying/moving too

jevbelikov:
I find it surprisingly difficult to come up with a good name for this one that is not long (Creating, Moving, Copying, Removing Files and Directories') and not too abstract ('Manipulating Files and Directories'). I find 'Things' a little bit to informal and unspecific. How about: 'Operations on Files and Directories'?

Ry4an:
"File and Directory Manipulation"
"Working with Files and Directories"
?

jevbelikov:
@Ry4an: +1, your second suggestion is my favourite so far

datamat:
@Ry4an: +1 for your second suggestion!

shwina:
+1 for "Working with Files and Directories". To avoid confusion with the "Files and Directories" module, may I also propose changing the title of `01-filedir.md` to something like "Finding Your Way Around"? This section is really about filesystem navigation.

Ry4an:
Spun up a tiny PR ( #378 ) to with the suggestions thus far.

Ry4an:
The PR went through with the title change, but I mistakenly thought that was interpolated more places than it is. I'm going to follow up with a PR to correct all the references to those sections.

Ry4an:
Tacked on #379 to fix the internal link text. Sorry for missing that on the first go round.

Ry4an:
Looks like this got merged today.  I think we're good to go. @ChristinaLK assigning to you for closing I guess.

Ry4an:
n/m looks like I can't assign.  I'll go mention it in the etherpad.

gdevenyi:
Resolved!

PR #379, opened 2016-06-07
Correct internal link text for renamed chapters.

Ry4an:
URLs don't change (to avoid broken inbound links), but the chapter title
link text needed updating.  Addresses swcarpentry/shell-novice#199 continuing swcarpentry/shell-novice#378

### DIFF
--- a/index.md
+++ b/index.md
@@ -42,8 +42,8 @@ These lessons will start you on a path towards using these resources effectively
 ## Topics
 
 1.  [Introducing the Shell](00-intro.html)
-2.  [Files and Directories](01-filedir.html)
-3.  [Creating Things](02-create.html)
+2.  [Navigating Files and Directories](01-filedir.html)
+3.  [Working With Files and Directories](02-create.html)
 4.  [Pipes and Filters](03-pipefilter.html)
 5.  [Loops](04-loop.html)
 6.  [Shell Scripts](05-script.html)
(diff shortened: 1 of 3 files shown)


shwina:
Can we also change the filename "01-filedir" and "02-create" to something more reflective of the new titles? I would propose `01-navigating.md` and `02-filedir.md`. As an aside, I think that "Navigating the file system"/"Navigating the filesystem" may be more appropriate for section 1 - any thoughts?

shwina:
I see your point. This being the first section, it's probably not a great idea to use the term "filesystem". My initial suggestion was for something like "Finding Your Way Around". It's friendlier and more accurate than the current "Navigating Diles and Directories". I'm uncomfortable with the notion of "navigating files". Perhaps "Browsing Files and Directories" is also OK.

shwina:
Perhaps changing file names and breaking URLs should be a separate issue. Thanks so much for raising these important points.

Ry4an:
I like "finding your way around" a lot. 

I'll add renaming to the pr and we'll see if anyone objects. :) 

On June 7, 2016 5:29:37 PM EDT, Ashwin Srinath <notifications@github.com> wrote:
>I see your point. This being the first section, it's probably not a
>great idea to use the term "filesystem". My initial suggestion was for
>something like "Finding Your Way Around". It's friendlier and more
>accurate than the current "Navigating Diles and Directories". I'm
>uncomfortable with the notion of "navigating files". Perhaps "Browsing
>Files and Directories" is also OK. 
>
>---
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/swcarpentry/shell-novice/pull/379#issuecomment-224420358

-- 
tYoped on a tiny keyboard.

gdevenyi:
This is good as-is. We need to be quite careful adjusting filenames as to not break external links, many of which we may not have control of.

Issue #172, opened 2015-05-28
Improve exercise "Moving and Copying"


rgaiacs:
In `02-create.md`, the exercise "Moving and Copying" isn't in conformance with our example filesystem.

ChristinaLK:
I think that's okay - maybe as an intro to the exercise, learners should try drawing the contents of the directory/filesystem, since they don't actually have these files in front of them.

gcapes:
Closing stale issue. Please re-open if this issue still applies.

Issue #371, opened 2016-05-09
Audit filesystem/lesson examples for consistency


gdevenyi:
It seems we have had some drift between the two, need to resolve this

rgaiacs:
We have files duplicated in three places: `filesystem/Users/nelle/`, `data-shell` and `shell-novice-data.zip`.

Can we drop `data-shell` and generate `shell-novice-data.zip` based on `filesystem/Users/nelle/`?

Can we also drop `shell-novice-data.zip` and have it only in the release page?

tomwright01:
+1 for dropping the zip file. Following someone elses suggestion I recently started the lesson by getting participants to clone a repo containing the filesystem. This worked great. 

Probably too great a change for today but I'd like to see a version controlled 'filesystem' repo that could be embedded in lessons as a submodule.

gdevenyi:
We have one more set of duplications, the filesystem examples in the lessons themselves. We'll need to ensure those are synced with the filesystem as well.

@tomwright01 we can't go for an all-git solution for the filesystem, in case of SWC workshops where git isn't taught, they won't necessarily have a way to grab the filesystem

jevbelikov:
in the zip archive there is an empty data-shell/Desktop/ sub-directory. Not sure why it is there. In any case, some example outputs (in 01-filedir.md) mention it and some don't.

shwina:
@rgaiacs.

> We have files duplicated in three places: filesystem/Users/nelle/, data-shell and shell-novice-data.zip.

I don't understand. What is `filesystem/Users/nelle`? Also, what if some instructors want to use the bleeding-edge version of the lesson - wouldn't dropping the `.zip` file make things difficult?

rgaiacs:
@shwina Sorry for the long delay to reply your question.

> What is `filesystem/Users/nelle`?

Did you get the answer for this question at https://github.com/swcarpentry/shell-novice/issues/490?

> Also, what if some instructors want to use the bleeding-edge version of the lesson - wouldn't dropping the .zip file make things difficult?

Long long time ago, at the time that we used, the now deprecated, https://github.com/swcarpentry/DEPRECATED-bc all instructors need to clone the repo and use the bleeding-edge version. Now we have a release process, https://github.com/swcarpentry/swc-releases, and we should suggest to instructors to use the last release and not the bleeding-edge version. Instructors that want to use the bleeding-edge version probably have a fork of the lesson and can generate the `.zip` file in 5 minutes or less.

[...]

PR #382, opened 2016-06-13
Improve exercise "Moving and Copying"

rgaiacs:
Related with #172.


gdevenyi:
Needs to be checked re: #371

shwina:
@rgaiacs, I agree with Christina (#172) that this should be a "fictional filesystem" example. Would it be OK to close?

Issue #337, opened 2016-03-04
Don't teach "rmdir"


DanielBaird:
Immediately after teaching `rm`, we teach `rmdir`, along with the paragraph or two to explain that it doesn't work when there's content in the dir you're removing.

Then in an callout we teach `rm -r` and warn learners about its potential for catastrophe.

`rmdir` seems useful in tiny example-sized situations with few files and shallow directory trees, but in practice I think learners will / should always use `rm -r` (with a mandatory pause before hitting enter).

So here's an opportunity to remove some content.  Unless anyone disagrees, I'll have a go at rewriting the rm / rmdir / rm -r section to leave out rmdir, and submit a pullreq for comment...?

shwina:
I'm +1 for this - thanks.

plijnzaad:
I concur.

gdevenyi:
PRs welcome.

PR #384, opened 2016-06-13
Remove rmdir, since no one really uses it

gdevenyi:
Addresses #337

### DIFF
--- a/02-create.md
+++ b/02-create.md
@@ -183,47 +183,31 @@ $ rm thesis
 rm: cannot remove `thesis': Is a directory
 ~~~
 
-This happens because `rm` only works on files, not directories.
-The right command is `rmdir`,
-which is short for "remove directory".
-It doesn't work yet either, though,
-because the directory we're trying to remove isn't empty:
+This happens because `rm` by default only works on files, not directories.
+
+To really get rid of `thesis` we must also delete the file `draft.txt`.
+We can do this with the "recursive" option for `rm`:
 
 ~~~ {.bash}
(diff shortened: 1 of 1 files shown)


Mahdisadjadi:
This is very good now!

shwina:
Maybe link to the [recursion](https://en.wikipedia.org/wiki/Recursion) wiki page? Learners may not be familiar with this term..

vfaghirh:
I usually make and remove several directories during my computational work, but I usually use rm -r to remove a directory. In general I think it is good to tell the readers how they can remove a directory and also give them several options for doing that. So I do not agree with removing this from the lessons, I think it could be useful to some people.

wking:
On Mon, Jun 13, 2016 at 11:08:04AM -0700, vfaghirh wrote:
> So I do not agree with removing this from the lessons, I think it
> could be useful to some people.

While I agree with @vfaghirh that rmdir is useful to some people (I
use it frequently), I also agree with the folks in #337 that it's
probably not useful enough to be worth teaching an additional command
in the learner's first four hours of shell.  So I'm fine removing it
(although pointing novices at ‘rm -r’ makes me jumpy ;).

gdevenyi:
@wking exactly why the callout stayed and discussed interactive mode :)

gdevenyi:
Addressed @shwina ' s suggestion of a link to wikipedia.

shwina:
+1 from me.

gdevenyi:
Great, merging

Issue #169, opened 2015-05-28
Improve paragraph about control keys


rgaiacs:
In `02-create.md` we

- [x] should mention that `^` means the control key before using "Control-O".
- [x] should mention that in some keyboards the control key is the "Ctrl".
- [x] should use only "Control", not "control".
- [x] **on Mac OS X**, control really means control and not command.

ChristinaLK:
when you say that `^` means control from the beginning, where were you thinking?  
otherwise, +3

r-gaia-cs:
> when you say that ^ means control from the beginning, where were you thinking?

In replacing

> Let’s type in a few lines of text, then use Control-O to write our data to disk:

with

> Let’s type in a few lines of text, then press `^0` *that is the Unix way to say hold the Control key and press `O`) to write our data to disk:

gdevenyi:
I agree with these changes, I'll try and get them done.

bsmith89:
Claiming this issue for the Bug BBQ (we're demoing how to do a pull request).

Ry4an:
Merged fix.  Probably closeable.

gdevenyi:
Resolved!

PR #385, opened 2016-06-13
Addresses Issue #169

pschloss:
We edited the callout about the control key and the text that surrounds it to tie it in better to the nano example.

### DIFF
--- a/02-create.md
+++ b/02-create.md
@@ -91,7 +91,7 @@ $ nano draft.txt
 > another directory the first time you "Save As..."
 
 Let's type in a few lines of text.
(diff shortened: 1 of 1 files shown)


gdevenyi:
Thanks for the contribution!

Does it make sense to mention the difference between Mac and PC keyboards for this? (I don't have a Mac to check)

Ry4an:
The text (predating this change) does mention that sometimes the key is labeled "Control" (Mac) and sometimes "Ctrl" (PC).  It also mentions that on Mac ^O *really* means Control not command, so I think it's good to go.

gdevenyi:
Great

PR #388, opened 2016-06-13
Replace /home/amanda by /Users/amanda to be more consistent with the rest of the lesson

mstimberg:
A tiny modification to a challenge of `01-filedir`:
The lesson consistently uses `/Users/...` (OS-X style) instead of `/home/...` (Linux-style). The exercise proposed `cd /home/amanda` as an answer for the question of how to get to Amanda's home directory "which is `/Users/amanda`". I think this makes this more into a "trick question" than it was supposed to be. Also, I think we want to include a correct solution that uses an absolute path.

### DIFF
--- a/01-filedir.md
+++ b/01-filedir.md
@@ -667,7 +667,7 @@ and we will see it in many other tools as we go on.
 > 
 >1.  `cd .`
 >2.  `cd /`
->3.  `cd /home/amanda`
+>3.  `cd /Users/amanda`
 >4.  `cd ../..`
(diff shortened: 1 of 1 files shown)


gdevenyi:
I think we're going to leave the filesystem changes/harmonization to @shwina who has claimed it for the bug BBQ.

There are a bunch of places (and figures) that need consistent updating, so this is a one-person job and must be done all at once for harmony.

mstimberg:
Ok, sure.

PR #391, opened 2016-06-13
correct text of intro

btel:
I tried to fix several issues concerning the text of the introduction.

### DIFF
--- a/00-intro.md
+++ b/00-intro.md
@@ -11,15 +11,15 @@ minutes: 5
 
 At a high level, computers do four things:
 
--   run programs
--   store data
--   communicate with each other
--   interact with us
+-   run programs,
+-   store data,
+-   communicate with each other,
+-   and interact with us.
 
 They can do the last of these in many different ways,
-including direct brain-computer links and speech interfaces.
-Since these are still in their infancy,
-most of us use windows, icons, mice, and pointers.
+including direct brain-computer and speech interfaces.
+Since these interfaces are still in their infancy,
+most of us use computer screens, mice, touchpads and keyboards.
 These technologies didn't become widespread until the 1980s,
 but their roots go back to Doug Engelbart's work in the 1960s,
 which you can see in what has been called
@@ -64,13 +64,13 @@ Bash is the default shell on most modern implementations of Unix
 and in most packages that provide Unix-like tools for Windows.
 
 Using Bash or any other shell
-sometimes feels more like programming than like using a mouse.
+is more like programming than like using a mouse.
 Commands are terse (often only a couple of characters long),
 their names are frequently cryptic,
 and their output is lines of text rather than something visual like a graph.
 On the other hand,
 the shell allows us to combine existing tools in powerful ways with only a few keystrokes
-and to set up pipelines to handle large volumes of data automatically thus improving productivity and reproducibility.
+and to set up workflows to handle large volumes of data automatically. This automation not only improves  productivity and also reproducibility as exactly the same workflow can be run multiple times.
 In addition, the command line is often the easiest way to interact with remote machines and supercomputers.
 Familiarity with the shell is near essential to run a variety of specialized tools and resources including high-performance computing systems. As clusters and cloud computing systems become more popular for scientific data crunching,
 being able to interact with them is becoming a necessary skill. We can build on the command-line skills covered here to tackle a wide range of scientific questions and computational challenges.


00-intro.md
```diff
@@ -11,15 +11,15 @@ minutes: 5
...
 At a high level, computers do four things:
 
--   run programs
--   store data
--   communicate with each other
--   interact with us
+-   run programs,
+-   store data,
+-   communicate with each other,
+-   and interact with us.
```
gdevenyi: Why the addition of punctuation to a bulleted list?

00-intro.md
```diff
@@ -64,13 +64,13 @@ Bash is the default shell on most modern implementations of Unix
 and in most packages that provide Unix-like tools for Windows.
 
 Using Bash or any other shell
-sometimes feels more like programming than like using a mouse.
```
gdevenyi: This is "sometimes feels" to make a distinction between the shell and programming proper, I believe.

00-intro.md
```diff
@@ -64,13 +64,13 @@ Bash is the default shell on most modern implementations of Unix
...
+is more like programming than like using a mouse.
 Commands are terse (often only a couple of characters long),
 their names are frequently cryptic,
 and their output is lines of text rather than something visual like a graph.
 On the other hand,
 the shell allows us to combine existing tools in powerful ways with only a few keystrokes
-and to set up pipelines to handle large volumes of data automatically thus improving productivity and reproducibility.
```
gdevenyi: pipelines have a specific meaning in the shell, refering to the use of the "|" aka the pipe, to tie programs together.

00-intro.md
btel: Many style guides (like this one https://proswrite.com/2013/06/20/confused-about-punctuation-in-bullet-lists/) suggest punctuating the list unless they consist of single words:

> If the stem sentence is an introductory phrase or clause, followed by a list, we start each list item with a lower case letter, and leave end punctuation off each item but the last. We give the last one a full stop. In essence, we see the list as a run-on sentence, with white space taking the place of punctuation between each item in the list.

00-intro.md
btel: but my impression is that this sentence refers to writing scripts which handle large data. Using pipeline is covered in the first part of this sentence:

> shell allows us to **combine** existing tools in powerful ways

00-intro.md
btel: Right. But isn't saying "shell is more **like** programming" enough to make this distinction? I am not super strong about it, but I think it would be an improvement. If you disagree I can remove this change.

00-intro.md
gdevenyi: Okay, great, but the goal of style guide is consistency, can you check the other SWC bulleted lists to see how they do things, we're going to stick with that way, whatever it is.

btel: It depends. In exercises usually no punctuation is used (but it makes sense in MCQs with lists of possible commands). On the other hand, learning objectives use punctuation. There is no more bullet list of the style above in the whole lesson. 

I tend to treat the bullets as part of the sentence, so the sentence should be grammatically correct if we remove the bullets and put everything on a single line. I think that is the rationale behind the cited style rule.

If this is the only block to merge this PR, I am happy to remove it (and maybe do another PR with the change above). Let me know what you recommend. Thanks!

00-intro.md
jevbelikov: for me the key part is the motivation: improving productivity and reproducibility, so wouldn't drop it, perhaps even move to a more prominent position or make bold instead

btel: Thanks. It appears in the line below :-) I also try to explain why it improves productivity and reproducibility.

00-intro.md
```diff
@@ -64,13 +64,13 @@ Bash is the default shell on most modern implementations of Unix
...
+and to set up workflows to handle large volumes of data automatically. This automation not only improves  productivity and also reproducibility as exactly the same workflow can be run multiple times.
```
shwina: and also reproducibility -> but also reproducibility.

00-intro.md
shwina: I also interpret that "pipeline" here refers to the use of `|`, rather than the use of shell scripts. In the pipes and filters lesson, we show how small tools like `wc` and `head` can be "combined" together to perform a complex task and potentially filter a large amount of data. I'm -1 for renaming to "workflow".

00-intro.md
```diff
@@ -11,15 +11,15 @@ minutes: 5
...
+-   communicate with each other,
+-   and interact with us.
 
 They can do the last of these in many different ways,
-including direct brain-computer links and speech interfaces.
-Since these are still in their infancy,
-most of us use windows, icons, mice, and pointers.
+including direct brain-computer and speech interfaces.
+Since these interfaces are still in their infancy,
+most of us use computer screens, mice, touchpads and keyboards.
```
shwina: Why get rid of the references to windows and icons?

00-intro.md
btel: I don't know the intention of the author, but in the original sentence he says 

> shell allows us [...] to set up pipelines to handle large volumes of data automatically. 

How would pipes (`|`) help to handle large data automatically? As you said they are good for combining existing tools into pipelines. Should we change the sentence then into:

> [...] with only a few keystrokes shell allows us to combine existing tools into powerful pipelines and handle large volumes of data automatically [...]

00-intro.md
btel: windows and links are not interfaces in the same sense as mice and keyboards. I replaced "windows and icons" with "computer screens".

00-intro.md
shwina: Yes - I like that!

gdevenyi:
Thanks for your contribution, but I think these changes move the intro away from it's original intent, as per the large amount of discussion.

btel:
@gdevenyi Could you specify which changes blocked this PR? I think we were reaching consensus on some of them. If you wish, I am happy to rebase this PR and keep only uncontroversial changes.

PR #395, opened 2016-06-13
Minor improvements for `05-script.md`

mstimberg:
A few minor improvements (I hope!) for the "shell script" episode:
* The introduction of command line arguments duplicated an almost identical phrase:

>  Instead, let’s edit middle.sh and replace octane.pdb with a special variable called $1:
> 
> $ nano middle.sh
> 
> Now, within “nano”, replace the text octane.pdb with the special variable called $1:

* The `sorted.sh` script was introduced with `cat sorted.sh`, which is a somewhat advanced convention of introducing a new script -- a learner typing this in will get an error because `sorted.sh` does not exist. Using `nano sorted.sh` seems to be more consistent with the rest of the episode.
* The "Usage" comment for `middle.sh` read: "Usage: middle.sh filename end_line num_lines" but we always call `middle.sh` with `bash` -- I changed it to "Usage: bash middle.sh ..."

### DIFF
--- a/05-script.md
+++ b/05-script.md
@@ -174,7 +174,7 @@ $ nano middle.sh
 ~~~
 ~~~ {.output}
 # Select lines from the middle of a file.
-# Usage: middle.sh filename end_line num_lines
+# Usage: bash middle.sh filename end_line num_lines
 head -n "$2" "$1" | tail -n "$3"
 ~~~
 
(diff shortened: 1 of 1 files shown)


05-script.md
```diff
@@ -208,9 +207,11 @@ to handle the case of parameters containing spaces
 Here's an example:
 
 ~~~ {.bash}
-$ cat sorted.sh
+$ nano sorted.sh
```
gdevenyi: The idea with "cat" is for it to correspond to the {.output} below. we're not editing the file, just viewing it.

mstimberg: There is no file `sorted.sh` yet, typing in `cat sorted.sh` does lead to an error for the learner. The style of showing a script's content with `cat` is often used in tutorials etc. online, but it is based on an understanding that the file does not exist on the reader's system, which is not clear for novice learners, I think.
The `nano ...` + `{.output}` (for file content) convention has been used earlier in the lesson.

gdevenyi: Ah, I see

PR #401, opened 2016-06-14
Changed Typing to Pressing

vfaghirh:
Changed Typing to Pressing wherever it refers to pressing the Ctrl key.


shwina:
+1 from me

btel:
just a minor issue - in the file discussion.md we still refer to "typing" Ctrl key:

> After typing our magnum opus directly into the shell prompt, we can then type Ctrl-D to tell `wc` that we're done and we'd like to see the results of the word count.

btel:
This change is duplicated in #400. Should we close this PR?

Issue #402, opened 2016-06-19
Add this image (if we can find the original)


gvwilson:
https://twitter.com/49SD/status/744189404030316544

ppgardne:
I've knocked the attached together (fixed some issues with case, some spelling, descriptions, etc.). 
[standard-filesystem-hierarchy.pdf](https://github.com/swcarpentry/shell-novice/files/322515/standard-filesystem-hierarchy.pdf)
Github wont let me upload the SVG (:-/) -- but the pdf should give you an idea. Where would you like to include it (if at all)? Comments/edits welcome.

ppgardne:
Fixed some of the issues I introduced:
[standard-filesystem-hierarchy.pdf](https://github.com/swcarpentry/shell-novice/files/322516/standard-filesystem-hierarchy.pdf)

PR #407, opened 2016-06-20
added a svg image 

ppgardne:
standard-filesystem-hierarchy.svg -- this gives an overview for a standard *nix filesystem and is in response to this issue:
https://github.com/swcarpentry/shell-novice/issues/402


btel:
Where should this file be linked? 

A few comments:

* There are lots of folder users do not usually care for (like `\boot` or `\var\lock`). Should we just keep only the top folder in the list without going into its hierarchy (like you do for `/dev` or `/lib`)? In contrast people may care for the content of their home dir, so perhaps it would be good to populate it somewhat. Obviously it depends very much on their system.
* The file hierarchy depends a lot on the platform. Should we include similar tree for Windows, MacOS users?
* `/proc` is missing
* `/home` appears twice in the list.

ppgardne:
Thanks for the comments Bartosz. 

- I've removed the duplicate /home & added /proc:
https://github.com/ppgardne/shell-novice/blob/gh-pages/fig/standard-filesystem-hierarchy.svg

- I could move /proc to the top and list some of it's contents and /boot to the right side. I agree that the contents of /proc may be more useful than /boot. 

- I think /home contents would be too system dependent to add much more to.

btel:
Thanks Paul. I think it would be useful to keep the hierarchy of folders that a beginner may encounter. Both `/proc` and `/boot` (but also `/var` and `/etc`) are too specialised. They are rather useful for sys admins.

I still miss  the context. At which place of the lesson should this figure be linked? How about MacOS and Windows users?

shwina:
@btel - very good points. Although I like this addition a lot, I think it only works for Linux users (and with some work, maybe for OS X users).

gvwilson:
I agree we should cut /proc and /boot, but novices are likely to see /var and /etc fly by.  And if we have something for Linux, hopefully it'll prompt people to do something for Mac and Windows, so I'm :+1: to merge after a bit of pruning.

ppgardne:
OK. I've removed /boot which I've never had to use. I have used /proc/*info files a bit. Usually when jumping around different machines. Am happy to remove if really needed -- of course anyone else can edit it too. ;-)

I'll add it to wikicommons too (and maybe to https://en.wikipedia.org/wiki/Unix_filesystem and/or https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard -- which probably should be merged).

shwina:
I echo @btel's sentiment. This is nice to have, but where in the lesson should we have it?

ppgardne:
It's a fair question. It may fit under "Navigating Files and Directories", or, I was wondering if a "resources" section would be a useful addition to the lessons. 

This could contain some simple "cheat sheets" (e.g. https://files.fosswire.com/2007/08/fwunixrefshot.png), directory structures and links to other useful online resources to help cement some of the material that was covered in the class.

shwina:
Perhaps `reference.md`?

shwina:
@ppgardne. What do you think?

ppgardne:
Hi @shwina, 
reference.md is a fine place for the figure, for now. 
P.

shwina:
Can you change the PR to reflect this? Otherwise, I'm happy to merge and then move the file.

ppgardne:
I wont get to it for a while, sorry. If you have time, that'd be swell!

shwina:
Merged and added a note to `reference.md`.

PR #408, opened 2016-06-20
Flow chart is not entirely representing for loop

lexnederbragt:
I find the flow chart part of the [shell_script_for_loop_flow_chart](https://github.com/swcarpentry/shell-novice/blob/gh-pages/fig/shell_script_for_loop_flow_chart.svg) in 04-loop not really representing a for-loop in that it says `$filename found --> No` --> `End`? If `$filename` is not found, the shell will throw an error.

This PR changes the text to `Next $filename?` in the hope that would work better.

(I don't know how to recreate the `vsd`file)

### DIFF
--- a/fig/shell_script_for_loop_flow_chart.svg
+++ b/fig/shell_script_for_loop_flow_chart.svg
@@ -557,7 +557,7 @@
 			<text x="27.61" y="360.56" class="st11" v:langID="3081"><v:paragraph v:horizAlign="1"/><v:tabList/>Start</text>		</g>
 		<g id="shape19-77" v:mID="19" v:groupContext="shape" v:layerMember="0" transform="translate(50.3263,-144.84)">
 			<title>Decision</title>
-			<desc>$filename found?</desc>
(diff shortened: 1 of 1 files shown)


shwina:
1. As @lexnederbragt mentions, `$filename found?` isn't the right question to ask. Maybe the better question is "Are we done with all files?". What say?

2. I'm not entirely convinced about what the `stdin`. Shouldn't `stdin` be `input-1.dat original-input-1.dat`, etc.?

3. Finally, the chart is created using Microsoft Visio. At the moment, we've got a copy of both the `.vsd`, and the corresponding `.svg`. This is a PR that updates only the `.svg`, which will make the `.svg` fall out of sync.

@fgacenga - copying you as the original contributor. Any suggestions?

gdevenyi:
I think I need to check back with the lesson-template on how we're supposed to handle updating SVG/PNG pairs as I'm not entirely sure right now what' the right process is.

gcapes:
I think this diagram is fixed now - please re-open if I've misunderstood.

PR #416, opened 2016-06-23
an update to loop part of the lesson (04-loop.md)

elizabethpermina:
I've added a suggestion for the problem with white spaces in file names (which is disscussed in the lesson as a side note). A lot of us deal with files created by somebody else and I though it might be useful to have a note in the lesson on how to get rid of all white spaces in the directory - but if it makes lesson too crowded it is easy to skip.

### DIFF
--- a/04-loop.md
+++ b/04-loop.md
@@ -70,6 +70,7 @@ In this case, the list is the two filenames.
 Each time through the loop,
 the name of the thing currently being operated on is assigned to
 the **variable** called `filename`.
+
 Inside the loop,
 we get the variable's value by putting `$` in front of it:
 `$filename` is `basilisk.dat` the first time through the loop,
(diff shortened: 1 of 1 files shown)


04-loop.md
```diff
@@ -230,6 +231,16 @@ the `head` and `tail` combination selects lines 81-100 from whatever file is bei
...
 > but it's simpler just to avoid using spaces (or other special characters) in filenames.
+> 
+> If you have a series of files with spaces, you can run a loop to rename files:
+~~~ {.bash}
+for filename in *
+do 
+	mv "$filename" "${filename/ /_}"
+done
+~~~
+> this will scan all the files in the directory for white spaces and replace them with underscore.
```
shwina: Without having real files to try this on, this command wouldn't work. Also it's quite cryptic (no explanation about the `/ /_`).

gvwilson:
This is a neat trick, but I think it's beyond our audience.

PR #417, opened 2016-06-25
clarify difference between shell and GUIs in introduction

btel:
This is a mutation of closed PR #391, which keeps only uncontroversial changes. It attempts to keep the description more internally consistent and be more specific about advantages of the shell. I happy to discuss these changes, so if you agree or disagree please leave your comment.

### DIFF
--- a/00-intro.md
+++ b/00-intro.md
@@ -14,16 +14,17 @@ At a high level, computers do four things:
 -   run programs
 -   store data
 -   communicate with each other
--   interact with us
+-   and interact with us
 
 They can do the last of these in many different ways,
-including direct brain-computer links and speech interfaces.
-Since these are still in their infancy,
-most of us use windows, icons, mice, and pointers.
-These technologies didn't become widespread until the 1980s,
-but their roots go back to Doug Engelbart's work in the 1960s,
-which you can see in what has been called
-"[The Mother of All Demos](http://www.youtube.com/watch?v=a11JDLBXtPQ)".
+including direct brain-computer and speech interfaces.
+Since these hardware interfaces are still in their infancy,
+we still have to rely on screens, mice, touchpads and keyboards.
+Although most modern desktop operating systems communicate with their human users by
+means of windows, icons and pointers, these software technologies didn't become
+widespread until 1980s. The roots of such *graphical user interfaces*  go back
+to Doug Engelbart's work in the 1960s, which you can see in what has been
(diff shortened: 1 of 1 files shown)


00-intro.md
```diff
@@ -14,12 +14,12 @@ At a high level, computers do four things:
...
--   interact with us
+-   and interact with us
 
 They can do the last of these in many different ways,
-including direct brain-computer links and speech interfaces.
-Since these are still in their infancy,
-most of us use windows, icons, mice, and pointers.
+including direct brain-computer and speech interfaces.
+Since these interfaces are still in their infancy,
+most of us use computer screens, mice, touchpads and keyboards.
```
shwina: By getting rid of windows and pointers, we lose the reference to the "software" part of how computers interact with us (and vice-versa). In fact, maybe we can rephrase to explicitly differentiate between software and hardware interfaces?

00-intro.md
```diff
@@ -69,8 +69,9 @@ Commands are terse (often only a couple of characters long),
 their names are frequently cryptic,
 and their output is lines of text rather than something visual like a graph.
 On the other hand,
-the shell allows us to combine existing tools in powerful ways with only a few keystrokes
-and to set up pipelines to handle large volumes of data automatically thus improving productivity and reproducibility.
+with only a few keystrokes shell allows us to combine existing tools into
```
shwina: "with only a few keystrokes**, the** shell allows us to combine existing tools into..."

00-intro.md
```diff
@@ -69,8 +69,9 @@ Commands are terse (often only a couple of characters long),
 their names are frequently cryptic,
 and their output is lines of text rather than something visual like a graph.
 On the other hand,
-the shell allows us to combine existing tools in powerful ways with only a few keystrokes
-and to set up pipelines to handle large volumes of data automatically thus improving productivity and reproducibility.
+with only a few keystrokes shell allows us to combine existing tools into 
+powerful pipelines and handle large volumes of data automatically. This automation not only improves
```
shwina: I would suggest something a little more explicit about what is productive and reproducible:

"not only makes us more productive, but also improves the *reproducibility* of our workflows".

"Exactly the same workflow can be run multiple times" even by clicking exactly the same buttons every time. The stress should be on how easily this can be done on the shell, for instance:

... but also improves the *reproducibility* of our workflows" by allowing us to repeat them with just a few commands.

btel:
Thanks @shwina for your feedback. I modified the commit along with your suggestions.

shwina:
Thanks, @btel

PR #419, opened 2016-06-27
Update 02-create.md

r-bowman:
Added a challenge concerning hiding a file for Software Carpentry Certification


gvwilson:
Not sure we should recommend hiding files as `.filename` as a general practice.

PR #420, opened 2016-07-01
Add shebang to script

TomKellyGenetics:
As suggested by @nickb- during Otago (New Zealand) workshop.

I agree with Nick that this is good practice to encourage users to do. It may reduce errors in the workshop as `middle.sh` should run as an executable without specifying `bash middle.sh`. I note that the `\bin\bash` directory may differ for particular machines, particularly the GitBash Windows tool and this may be addressed before merging.

### DIFF
--- a/05-script.md
+++ b/05-script.md
@@ -22,14 +22,17 @@ these are actually small programs.
 Let's start by going back to `molecules/` and putting the following line into a new file, `middle.sh`:
 
 ~~~ {.bash}
-$ cd molecules
-$ nano middle.sh
+cd molecules
+nano middle.sh
 ~~~
-The command <code>nano middle.sh</code> opens the file `middle.sh` within the text editor "nano" (which runs within the shell). If the file does not exist, it will be created. We can use the text editor to directly edit the file. We'll simply insert the following line:
+The command <code>nano middle.sh</code> opens the file `middle.sh` within the text editor "nano" (which runs within the shell). If the file does not exist, it will be created. We can use the text editor to directly edit the file. We'll simply insert the following lines:
 
 ~~~
+#! \bin\bash
+
 head -n 15 octane.pdb | tail -n 5
 ~~~
+
 This is a variation on the pipe we constructed earlier:
 it selects lines 11-15 of the file `octane.pdb`.
 Remember, we are *not* running it as a command just yet:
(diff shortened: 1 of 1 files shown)


TomKellyGenetics:
That makes sense, we likely should avoid permissions in the interest if time.

However, a script starting with `#! \bin\bash` would still run with `bash script.sh` without changing the rest of the lessons. Participants could be aware of best practices and recognise other scripts provided to them without needing to understand permissions. Any file generated by them (or given by someone more experienced) would be able to run in this way.

I'm not suggesting we change the rest of the lesson materials or go into any depth. Simply mention briefly "this is what a script looks like". It's ok if this is already been decided upon by the wider community, I wasn't aware this discussion had already been had when submitting Pull Request.

shwina:
Sorry, I don't think it is a good idea to put in the shebang, and especially not if we're not going to discuss it.

TomKellyGenetics:
What gave you the impression this wasn't a discussion about it? Thank you for making your opinion on the matter clear but I was hoping for a wider range of perspectives.

However, as I mentioned, I was not aware before submitting this that it had been discussed before as you have linked above and here: https://github.com/swcarpentry/shell-novice/pull/301 and https://github.com/swcarpentry/instructor-training/issues/119

I think it is a separate issue to `chmod` but it may still be better suited to [Shell Extras](https://github.com/swcarpentry/shell-extras) as @gvwilson has suggested in a related Pull request.

shwina:
@TomKellyGenetics - sorry for not being clear. A useful addition of the shebang into the lesson will involve:

1. A description of what the shebang does (provides a path to the executable that will run the script)
2. A discussion about permissions and `chmod`, so that the script can be made executable.

I don't see how the two are separate issues. Step 1 by itself achieves nothing. For all practical purposes, we have simply introduced a (very cryptic) comment line. Because we will not be covering permissions, and hence not making our scripts executable, I don't think there is a reason to bring up `#!`. Doing so will perhaps tempt some instructors to take a long aside explaining permissions an executable scripts, or otherwise leave an empty space in the concept maps that learners build for themselves through the lesson.

Sorry for not mentioning shell-extras. Perhaps a good place for this would be [`permissions.md`](https://github.com/swcarpentry/shell-extras/blob/gh-pages/04-permissions.md)?

iglpdc:
I agree with @shwina, but, in any case, I think that it'd be better to use `#!/usr/bin/env bash` to make it more portable.

TomKellyGenetics:
Thanks for clarifying, I've moved suggestion of this to shell-extras.

PR #421, opened 2016-07-01
changed keys for Ctrl commands to \<kbd\>

TomKellyGenetics:
Render keys as: <kbd>Ctrl</kbd>+<kdb>X</kbd> rather than `Ctrl+X`

`Ctrl` notation kept in discussion of alternate notation for keys

### DIFF
--- a/02-create.md
+++ b/02-create.md
@@ -136,12 +136,14 @@ return to the shell.
 >
 > * `Control-X`
 > * `Control+X`
+> * `Ctrl+X`
+> * `Ctrl+X`
 > * <kbd>Ctrl</kbd>+<kdb>X</kbd>
 > * <kbd>Ctrl</kbd>+<kdb>X</kbd>
 > * `^X`
 >
 > In nano, along the bottom of the screen you'll see `^G Get Help ^O WriteOut`.
-> This means that you can use `Control-G` to get help and `Control-O` to save your
+> This means that you can use <kbd>Ctrl</kbd>+<kdb>G</kbd> to get help and <kbd>Ctrl</kbd>+<kdb>O</kbd> to save your
 > file. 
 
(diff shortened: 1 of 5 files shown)


shwina:
I'm neither +1 nor -1 on this. @gdevenyi can you weigh in please?

gvwilson:
I'm -1 on anything that requires HTML in Markdown - I realize `kbd` is 
the right markup to use, but until Markdown/Kramdown/Jekyll give us a 
notation for it, I think we should stick to plain old back quotes.

TomKellyGenetics:
Fair enough, I spotted this notation on StackOverflow and found it easier to read in the text without further explanation. I've intentionally left mention of alternative notations in `reference` and `discussion`. 

In case new users are the reason for sticking with markdown: let it be noted that I am new to HTML and it was not difficult to figure this out emulating the use of it by others (just as I learned markdown).

TomKellyGenetics:
Please note that I have also used this \<kbd\> notation here and will change accordingly depending on the outcome of this PR.

https://github.com/swcarpentry/git-novice/pull/301

gvwilson:
I'm -1 on using HTML tags in Markdown, especially when there's no 
visible difference between the `kbd` tag and plain old backquotes.

TomKellyGenetics:
It renders differently on the web (as shown above). I've noted concerns here and elsewhere: `kbd` has been removed from my other pull request (on git) so the notes will remain consistent if we stay with backticks.

gdevenyi:
We're definitely moving to pure markdown (thanks to @gvwilson's great work), so as pretty as this is, I'm -1 on embedding HTML.

gdevenyi:
Thanks for your help, can you update the other PR to incorporate this change? Thanks!

TomKellyGenetics:
Switched to backticks on the other one already :)

PR #427, opened 2016-07-08
A small change to clarify part of "Using wildcards" challenge.

asherkhb:
Previous directions appeared to suggest that only a single answer could be true, but both 1 & 3 are true.
Wording was changed to reflect that multiple answers might be correct.
NOTE: This is being submitted as an instructor training checkout task for Asher Baltzell (ahaug@email.arizona.edu)


shwina:
Thanks, @asherkhb

Issue #425, opened 2016-07-08
Add table w/ commands for working with files and folders


lexnederbragt:
The last time I taught the shell, I wrote the table below on the blackboard. When helping learners that got stuck, when I said 'do you remember the command for ...', I could then point to the table, which helped them a lot. If instructors find it useful, it could be added to the lesson or the instructor notes.

|Action      |Files|Folders     |
|------------|-----|------------|
|Inspect     |ls   |ls          |
|View content|cat  |ls          |
|Navigate to |cd   |cd          |
|Move        |mv   |mv          |
|Copy        |cp   |cp -r       |
|Create      |nano |mkdir       |
|Delete      |rm   |rmdir, rm- r|

shwina:
+1. I think `reference.md` would be the best place for this.

PR #428, opened 2016-07-11
Add summary table of navigation commands

lexnederbragt:
Summarises commands for working with files and folders.

Addresses #425


shwina:
May I suggest avoiding a heading for the table? Maybe add a new bullet point like "Here is a table summarizing some of the commands used in this and the previous section:"

lexnederbragt:
Sure!

shwina:
Thanks!

PR #430, opened 2016-07-14
Describe the shell being a different interface to files.

ddevcc:
Mention that files & directories created in the shell are no different than files & directories created "normally" using a GUI program.

This can sometimes be a stumbling block for people are new to using a shell.  Their mental model can be that things done inside the shell only exist inside the shell and anything done inside is "trapped" there.

Having someone see that their newly created files exist the same as any other file can be a good way of changing that mental model.

### DIFF
--- a/02-create.md
+++ b/02-create.md
@@ -54,6 +54,12 @@ Desktop/    pizza.cfg
 molecules/  solar.pdf
 ~~~
 
+Using the shell to create a directory is no different than using a file explorer.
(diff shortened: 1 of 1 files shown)


shwina:
Make this a callout please? Perhaps with a heading like "two ways of doing the same thing"?

gvwilson:
I'll merge and then turn into a callout.

shwina:
Thanks @ddevcc

PR #434, opened 2016-07-23
Adds missing link to lesson template.

crocha700:
This tiny commit adds  a missing link to Lesson template –– the text was correct, but the there was no  link for the ID [lesson-example].


gvwilson:
Thanks for catching that - we really need to add a link checker to our linting tool...

PR #437, opened 2016-08-03
Added challenge - Stefan Wyder

swyder:
add exercise to sort numbers in scientific notation using sort -g
(sort -n doesnt sort correctly)

instructor training


gcapes:
Thanks for the contribution @swyder 

I think this exercise is too similar to the `sort -n` exercise to justify inclusion in an episode which already contains lots of exercises.

PR #442, opened 2016-08-25
Removed redundant content on script comments

DamienIrving:
In the shell scripting lesson there are two different scripts that are created. The importance of commenting your scripts is explained while creating the first one (`middle.sh`) and then re-explained like it had never been explained previously with the second one (`sorted.sh`). What's more, the`sorted.sh` example starts by putting in comments right away and then seems to forget that it did so later on. In this pull request I've tried to address these issues.


DamienIrving:
I've also just updated the introduction so it says that Nelle Nemo has 1520 samples. That's consistent with what it says in lessons 02 and 04 later one (I think the confusion arose because there are 300 proteins).

shwina:
Thanks! Our actual data has only 17 files; should we mention somewhere that this is a representative subset of Nelle's data?

DamienIrving:
@shwina Thanks for merging. Sorry I didn't get around to making a change to indicate that the 17 files is just a representative sample. I think that's a good idea.

PR #443, opened 2016-09-01
02:Clarification of single quotes on PS1='$ '

mjones01:
Current description doesn't specify why the quotes are needed (and the aren't if you don't also want the space after $.  
I asked Greg about this during my instructor demo today and he suggested that it could be better clarified.

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -46,7 +46,8 @@ let's open a shell window:
 > If you type the command:
 > `PS1='$ '`
 > into your shell, followed by pressing the 'enter' key,
-> your window should look like our example in this lesson.  
+> your window should look like our example in this lesson. The single
+> quotes around the $ allows you to specify the space after the $ prompt.   
 > This isn't necessary to follow along (in fact, your prompt may have 
 > other helpful information you want to know about).  This is up to you!  
 {: .callout}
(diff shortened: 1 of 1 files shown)


gdevenyi:
Hi, were there questions in during a lesson that prompted this addition? It feels a bit like excess information.

mjones01:
No, I haven't taught this lesson yet.  I did this for my instructor checkout lesson with Greg Wilson and asked him why the quotes were needed as I was just doing PS1=$ and it seemed to work -- I hadn't realized that there was a space after the $.  He suggested that I suggest of change to clarify.  Feel free to ignore if you don't think it is necessary.

rgaiacs:
@mjones01 Thanks for the lesson. The change looks good to me since we didn't introduce yet the concept that strings with space need quotes. I would only ask that you use `$` instead of $.

mjones01:
@rgaiacs Done!  Thanks.

shwina:
I agree with @gdevenyi that this is excess information, and doesn't really explain itself too well. I think it's more important that we say: "Don't forget the empty (space) character after the `$` sign".

shwina:
@mjones01 - thoughts?

mjones01:
As it seems like others find little need for this change, I'm happy to close the PR. Or I can revise to @shwina 's suggestion.  
In alignment with the Discuss listserv from today/yesterday, this change is probably only relevant to independent learners who come across the lessons, not those in the workshop as the instructor will likely say what what they are doing when they do this.

gdevenyi:
I think we can leave this for now.

PR #444, opened 2016-09-01
Add solutions to renaming, and moving and copying

ssmiller:
Added solutions for two of the challenges, renaming files and moving and copying.


iglpdc:
Thanks @ssmiller! Looks good to me.

PR #458, opened 2016-09-29
Clean fig/shell_script_for_loop_flow_chart.svg

rgaiacs:
Old file when export to PNG by Inkscape (0.91 r13725 (Jul 31 2016)):

old

New file when export to PNG by Inkscape (0.91 r13725 (Jul 31 2016)):

new


shwina:
Thanks so much for this, @rgaiacs

PR #460, opened 2016-09-29
Remove mention of Python

rgaiacs:
Because self-learners can be pushed to Python instead of R
or any other programming language because we mention it
on our lesson. Every programming language have their good side.

### DIFF
--- a/_episodes/07-find.md
+++ b/_episodes/07-find.md
@@ -463,9 +463,9 @@ $ grep "FE" $(find .. -name '*.pdb')
 > formulas?
 >
 > The third choice is to recognize that the shell and text processing have
-> their limits, and to use a programming language such as Python instead.
+> their limits, and to use another programming language.
 > When the time comes to do this, don't be too hard on the shell: many
-> modern programming languages, Python included, have borrowed a lot of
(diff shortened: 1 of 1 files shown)


_episodes/07-find.md
```diff
@@ -463,9 +463,9 @@ $ grep "FE" $(find .. -name '*.pdb')
 > formulas?
 >
 > The third choice is to recognize that the shell and text processing have
-> their limits, and to use a programming language such as Python instead.
+> their limits, and to use another programming language.
 > When the time comes to do this, don't be too hard on the shell: many
-> modern programming languages, Python included, have borrowed a lot of
+> modern programming languages have borrowed a lot of
```
shwina: Isn't Python more suited to performing shell-type and text-processing tasks? Not to start a language war at all, but R promotes itself as "a language for statistical computing and graphics":

https://www.r-project.org/about.html

I see no harm as mentioning Python as an example.

gdevenyi: I like this change, this is more commentary on the limits of shell than the benefits of other languages, lets leave the learners up to decide what's better for them.

shwina: OK with me then

PR #475, opened 2016-10-30
shell lesson: fix typos, make content a bit clearer

daattali:


shwina:
Thanks, looks great to me.

PR #479, opened 2016-11-13
Iteration figure now changes file names

HoWol76:
I noticed that the figure for the loop always has the same filename, 'input-1.dat', when it should clearly change in every iteration. This very tiny pull request addresses this one issue, and changes the filenames to 'input-2.dat' and 'input-3.dat' on successive iterations.


gdevenyi:
Awesome thanks!

PR #481, opened 2016-11-18
Clarified 'for loop' description

BrianBill:
Clarified the explanation of how a for loop works in general and related it more clearly to the example.


gdevenyi:
Looks great. Thanks for the clarification!

PR #482, opened 2016-11-22
Color version of filesystem figures were introduced

arminms:
The BW version of the following files were replaced by more appealing and consistent color versions:

- filesystem.svg
- find-file-tree.svg
- home-directories.svg


shwina:
Excellent. Any chance we can get a monospace font for the directory names?

arminms:
I've used OCR-A font which is monospaced (https://en.wikipedia.org/wiki/OCR-A). But on systems which OCR-A is not available, it may be replaced by another font which is not monospaced.  If this is the case, I can change it to Courier New or any other font that you like.

shwina:
I'm OK to merge!

Issue #464, opened 2016-10-10
Flags in ls command


crocha700:
Currently we teach the use of two or more flags of  `ls` as separate inputs, e.g., 

``` bash
ls -F -l -t
```

But equally valid –– and convenient –– is to patch the flags

```
ls -Fht
```

Is the first form preferred on pedagogical grounds? Would it be useful to teach the second form? Perhaps using a "pin box"?

rgaiacs:
> Is the first form preferred on pedagogical grounds?

Yes. For a novice it makes things easy to read.

> Would it be useful to teach the second form? Perhaps using a "pin box"?

I'm -1 to add it to the novice lesson since beginners are already presented with a lot of new information.

shwina:
Agreed. Something needs to go out if we're going to add in new information.

iglpdc:
This (patching the flags) used to be in the lesson. I think it is useful to have: it's standard practice, saves a lot of time when typing, and instructors may accidentally use it during the lesson or solo-learners will find it in the web.

When I teach the lesson I just say that, say, `ls -F -l -t` is equivalent to `ls -Flt` and enter the second comment after the first, so people can actually see it's really the same.

I would add a really short sentence to the lesson just saying this.

gdevenyi:
I'm okay only with a small callout if anything for improving this. It's advanced usage, and makes the lesson longer.

shwina:
PRs welcome on this

jarekbryk:
I think this can be closed given that my PR #496 was merged a while ago.

PR #483, opened 2016-11-22
Add sentence equating ls -F -a to ls -Fa

btmoyers:
This is a suggested change to address issue #464, and is part of my SWC instructor training checkout. I've added a single sentence to explain that shell flags can be concatenated if the learner wishes to type fewer characters. I appreciate any critical feedback!


gdevenyi:
Thanks for the addition.

Can you expand this to explain this is a general principle for most command line tools? Thanks.

gdevenyi:
Fixed by #496

Issue #487, opened 2016-12-03
Desktop directory in data-shell


ateucher:
I'm just getting ready to teach next week, and I see there is a `Desktop` directory in the `data-shell` directory that learners download. Is there a reason for that? It seems like it could be confusing

shwina:
I don't know if any instructors have used this to their advantage (i.e., demonstrating that the two `Destop` directories have different absolute paths, and are distinct directories, etc.,)

I'm all in favour of getting rid of it otherwise.

rgaiacs:
> > I see there is a Desktop directory in the data-shell directory that learners download. Is there a reason for that? It seems like it could be confusing

`data-shell` try to replicate the file system that learners will find in any Unix-like machine with `/`, `/bin`, `/home`, ... 

> I'm all in favour of getting rid of it otherwise.

I would like to keep it.

shwina:
@rgaiacs, I am a little confused: `data-shell` has none of this other than `Desktop`. 

└── data-shell
    ├── Desktop
    ├── creatures
    ├── data
    ├── molecules
    ├── north-pacific-gyre
    ├── notes.txt
    ├── pizza.cfg
    ├── solar.pdf
    └── writing

ateucher:
I think it's the filesystem directory that mimics the contents of /.  In that directory there is Users/nelle and Users/Larry both of which contain Desktop folders

shwina:
Ah, I understand now. In any case, I think that `data-shell/Desktop` can be removed. `filesystem/Users/Larry/Desktop` and `/filesystem/Users/nelle/Desktop` may be retained, depending on the discussion in #490.

rgaiacs:
> I think it's the filesystem directory that mimics the contents of /. In that directory there is Users/nelle and Users/Larry both of which contain Desktop folders

@ateucher You are right.

@shwina Sorry to not be clear. `data-shell` was created as a copy of `filesystem/Users/nelle` as you can check at [data-shell](https://github.com/swcarpentry/shell-novice/tree/gh-pages/data-shell) and [filesystem/Users/nelle](https://github.com/swcarpentry/shell-novice/tree/gh-pages/filesystem/Users/nelle).

> I'm just getting ready to teach next week, and I see there is a Desktop directory in the data-shell directory that learners download. It seems like it could be confusing

[...]

PR #489, opened 2016-12-06
Remove Desktop dir from data-shell

ateucher:
This PR removes the Desktop directory from `data-shell`, which learners download to their desktop. I understand this may be there for a teaching moment about relative and absolute paths etc, but I think it could be confusing. In addition it is not mentioned in the lesson that I can find.

Addresses #487


ateucher:
Closing until #371 is sorted.

Issue #177, opened 2015-05-28
Add example of "grep -r"


rgaiacs:
In `06-find.md` we don't mention recursive grep. I think this is very useful and we **must** mention it.

ChristinaLK:
+1

gvwilson:
@rgaiacs please PR and we'll get someone to review.

PR #491, opened 2016-12-06
Episode 7: Add grep -r exercise

petersm3:
matthew@cgrb.oregonstate.edu
Instructor Training: Checkout Procedure
Part 1: Submit a Small Contribution to One of Our Lessons

Possible lesson addition for issue: Add example of "grep -r" #177

### DIFF
--- a/_episodes/07-find.md
+++ b/_episodes/07-find.md
@@ -191,6 +191,50 @@ $ grep -n -w -v "the" haiku.txt
 ~~~
 {: .output}
 
+
+`grep` also has the ability to search for a a pattern recursively through a set of 
+files in subdirectories.
+
+To demonstrate this we will `split` the haiku into one file per line (for a total 
+of 11 files) and move each file to its own subdirectory.
+
+~~~
+# Create 11 files (xaa - xak) each containing one line of the haiku
+$ split -l 1 haiku.txt
+# Create a nested directory tree
+mkdir -p d01/d02/d03/d04/d05/d06/d07/d08/d09/d10/d10
+# Move each file into a subdirectory
+mv xaa d01
+mv xab d01/d02
+mv xac d01/d02/d03
+mv xad d01/d02/d03/d04
+mv xae d01/d02/d03/d04/d05
+mv xaf d01/d02/d03/d04/d05/d06
+mv xag d01/d02/d03/d04/d05/d06/d07
+mv xah d01/d02/d03/d04/d05/d06/d07/d08
+mv xai d01/d02/d03/d04/d05/d06/d07/d08/d09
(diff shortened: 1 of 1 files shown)


petersm3:
The example may have too many files and subdirectories; it may be better to split the haiku.txt into 2 or 3 lines per file, with 6 or 4 subdirectories respectively (as opposed to 11).

petersm3:
The setup step could be eliminated, i.e., the `mkdir`/`split`/`mv`, and the files pre-staged in [shell-novice-data.zip](http://swcarpentry.github.io/shell-novice/setup/)

shwina:
@petersm3 

Yes, I think it would be better to pre-stage the required files/directories in the `.zip` file. It's not necessary that this example on `grep -r` use `haiku.txt` as input.

petersm3:
Should I update this commit or create a new commit which adds the assets created by the `split` of `haiku.txt` in the `data-shell/writing` directory along with the updating of the `07-find.md` documentation, i.e., `grep -r` example, to reflect using the pre-staged files?

"It's not necessary that this example on grep -r use haiku.txt as input."

Suggestions as to how a `grep -r` example could be illustrated in the lesson? (other than using the `haiku.txt` example)

gdevenyi:
I think us (the maintainers) need to do an audit on the content of the assets to make sure they currently match the lesson, please hold off on adding the assets until we sort that. Thanks!

gcapes:
I'm always keen to avoid adding to the lesson unless required. Could we not have a `grep -r` exercise using the existing files in `data-shell/writing`?

gcapes:
Something like this would show recursive grep, using the existing files in `data-shell/writing`:

```
$ grep -r Yesterday
data/LittleWomen.txt:"Yesterday, when Aunt was asleep and I was trying to be as still as a
data/LittleWomen.txt:Yesterday at dinner, when an Austrian officer stared at us and then
data/LittleWomen.txt:Yesterday was a quiet day spent in teaching, sewing, and writing in my
haiku.txt:Yesterday it worked
```

drj11:
I would avoiding teaching `grep -r` as it is not portable. In particular, the `-r` is not required by POSIX: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/grep.html

gcapes:
Nonetheless I just successfully tested `grep -r` on Linux, macOS and Git Bash, so it should work for all learners.

mattipt:
A concrete example where grep -r may not work is Solaris.

gcapes:
While that may well be true, I'm not sure this is likely to pose a problem in workshops, due to the lack of Solaris users.

gdevenyi:
So at this point I think we'd like this rewritten to take advantage of ``data-shell/writing`` since it has the appropriate material.

PR #492, opened 2016-12-18
Info about http://tldr-pages.github.io/ manual tool

klemensnoga:
Info about http://tldr-pages.github.io/ manual tool added to Instructor Notes. I think that tool could be useful for instructors, especially due to lack of `man` command in Git Bash used on Windows.


gdevenyi:
Thanks!

PR #495, opened 2017-01-05
Correction of two typos in 04- (line 306) and 05- (line 237)

jarekbryk:
Part of the checkout procedure for SC Instructor Training.


shwina:
Thanks!

PR #496, opened 2017-01-06
Added sentence equating ls -F -a to ls -Fa #483

jarekbryk:
I noticed the previous change has not yet been merged after a few months, maybe you'll like this solution better?
This is me again contributing to the checkout procedure for instructor training.


gdevenyi:
Thanks!

PR #514, opened 2017-02-14
Adding solutions to exercises in 03-create

edenisen:
Adding solutions to exercises in 03-create 'Working With Files and Directories': 'Organizing Directories and Files' and 'Copy with Multiple Filenames' as a part of the Checkout Procedure for the SWC Instructor Training.

PR #515, opened 2017-02-15
Add solutions to challenges 03create

kemya:
Adding solutions To the challenges in lesson 03 
as part of instructor training checkout


shwina:
@kemya - thank you for the contribution.

It looks like an older PR #514 already adds a solution to the "Organizing Directories and files" exercise. 

* Perhaps you can wait for that PR to be merged, and then rebase?
* Or (I think this is the right way): you can use `git revert` (https://git-scm.com/docs/git-revert) to undo the changes made by commit `dc37cca`

Let me know if I can help!

PR #517, opened 2017-02-16
Fix #516 - Reword script creation for clarity

### LINKED ISSUE #516: 06-script Confusing wording creating first script file
Currently we have :

> Let’s start by going back to molecules/ and putting the following line into a new file, middle.sh:
> 
> ```
> $ cd molecules
> $ nano middle.sh
> ```
> The command nano middle.sh opens the file middle.sh within the text editor “nano” (which runs within the shell). If the file does not exist, it will be created. We can use the text editor to directly edit the file—we’ll simply insert the following line:
> 
> ```
> head -n 15 octane.pdb | tail -n 5
> ```
So "the following" mentioned in the first line, isn't encountered until a good while later. I'll create a PR to reword this more clearly.

gcapes:
Fixes confusing wording described in #516

### DIFF
--- a/_episodes/06-script.md
+++ b/_episodes/06-script.md
@@ -26,7 +26,8 @@ a bunch of commands saved in a file is usually called a **shell script**,
 but make no mistake:
 these are actually small programs.
 
-Let's start by going back to `molecules/` and putting the following line into a new file, `middle.sh`:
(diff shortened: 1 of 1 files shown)


gdevenyi:
Thanks!

gcapes:
Thanks for the quick turnaround!

PR #519, opened 2017-02-20
Fix #508 : explain `Ctrl-R` in the text

gcapes:

PR #520, opened 2017-02-20
Fix #518 for episodes 03 and 04

### LINKED ISSUE #518: Missing solutions to, context for, and files required by exercises

> 
> None of the above.

gcapes:
More to follow - is it easier if I create a new PR for each episode? Or would you rather I keep all these changes together?
Cheers
G

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -631,7 +631,7 @@ but it does find the copy in `thesis` that we didn't delete.
 >     When you look at your home directory using the GUI file explorer,
 >     does the file show up?
 >
-> 2.  Use `ls -l` to inspect the file's.  How large is `my_file.txt`?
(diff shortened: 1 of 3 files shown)


shwina:
Smaller PRs would be much better going forward - thank you!

shwina:
Also, I see that the commit from #519 is here as well. Do you want to close that PR?

gcapes:
My mistake - I branched at the wrong point. I can close PR #519, or fix it and remove the commit from this PR. If you need me to split this PR, please let me know, and I'll give that a go. Otherwise, I'll just stop adding to it, and instead reference the issue number I'm fixing when I submit new, separate PRs for  episodes 5/6/7.

gcapes:
I'll close this, and open 3(+) separate pull requests, grouping the changes to each episode.

shwina:
Thanks!

PR #521, opened 2017-02-21
Fix #518 for 03-create

### LINKED ISSUE #518: Missing solutions to, context for, and files required by exercises

gcapes:
Add solutions, provide working directory context, and where necessary modify the exercises to work with files in data-shell (so that learners can test commands rather than having to work everything out in their heads).

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -632,7 +632,7 @@ but it does find the copy in `thesis` that we didn't delete.
 >     When you look at your home directory using the GUI file explorer,
 >     does the file show up?
 >
-> 2.  Use `ls -l` to inspect the file's.  How large is `my_file.txt`?
+> 2.  Use `ls -l` to inspect the files.  How large is `my_file.txt`?
 >
 > 3.  When might you want to create a file this way?
 {: .challenge}
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -556,12 +556,13 @@ but it does find the copy in `thesis` that we didn't delete.
 > {: .output}
 >
 > > ## Solution
-> > Jamie needs to move her files `fructose.dat` and `sucrose.dat` to the `analyzed` directory:
-> >
-> > ~~~
-> > $ mv fructose.dat sucrose.dat analyzed
-> > ~~~
+> > ```
+> > mv *.dat analyzed
+> > ```
 > > {: .bash}
+> > Jamie needs to move her files `fructose.dat` and `sucrose.dat` to the `analyzed` directory.
+> > The shell will expand *.dat to match all .dat files in the current directory.
+> > The `mv` command then moves the list of .dat files to the "analyzed" directory.
 > {: .solution}
 {: .challenge}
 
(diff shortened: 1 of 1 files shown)


_episodes/03-create.md
```diff
@@ -556,43 +556,54 @@ but it does find the copy in `thesis` that we didn't delete.
 > {: .output}
 >
 > > ## Solution
```
shwina: Why get rid of the text explanation?

_episodes/03-create.md
```diff
@@ -556,43 +556,54 @@ but it does find the copy in `thesis` that we didn't delete.
...
 > > {: .bash}
+> > The shell will expand *.dat to match all .dat files in the current directory.
+> > The `mv` command then moves the list of .dat files to the "analyzed" directory.
 > {: .solution}
 {: .challenge}
 
 > ## Copy with Multiple Filenames
 >
-> What does `cp` do when given several filenames and a directory name, as in:
+> For this exercise, you can test the commands in the data-shell/data directory.
```
shwina: Use backticks: `data-shell/data`

_episodes/03-create.md
```diff
@@ -556,43 +556,54 @@ but it does find the copy in `thesis` that we didn't delete.
...
+>
+> In the example below, what does `cp` do when given several filenames and a directory name?
 >
 > ~~~
+> $ cd ~/data-shell/data
```
shwina: I would leave out the `cd` here. Many participants will have put the `data-shell` directory in some other location (not `~`).

_episodes/03-create.md
```diff
@@ -556,43 +556,54 @@ but it does find the copy in `thesis` that we didn't delete.
...
 > ~~~
-> $ cp intro.txt methods.txt survey.txt
+> $ cp amino-acids.txt animals.txt morse.txt 
 > ~~~
 > {: .bash}
 >
 > > ## Solution
-> > When given several filenames and a directory name, `cp` copies the files to the destination directory (the last argument).
-> > When given three or more filenames, `cp` expects the last argument to be the destination directory, therefore it returns an error: `target survey.txt is not a directory`.
+> > If given more than one file name followed by a directory name, `cp` copies the files
```
shwina: Let's make it explicit that the directory name must be the last argument

_episodes/03-create.md
gcapes: I think it should stay after all. Perhaps I was going to add it to the exercise text instead.

gcapes:
Thanks for the review. Do I just rebase onto the same branch, after fixing the points in the comments?

gcapes:
Hi @shwina, 

I hope this is what you intended? I've rebased and fixed the issues addressed in your comments. I had to force push the rebased branch, which means the original commits have disappeared/now show as out of date. 

I also dropped the commit about `Ctrl-R` because that has already been merged and shouldn't have been on this branch. 

Could you confirm this is how I should do the rebasing?
Thanks

shwina:
Yup - it looks good to me. If you are done adding commits, I will merge.

gcapes:
Excellent! All done. I'll let you get back to reviewing the other PRs then ;)

PR #522, opened 2017-02-21
Fix #518 for 04-pipefilter

gcapes:
Add solutions, provide working directory context, and where necessary modify the exercises to work with files in data-shell (so that learners can test commands rather than having to work everything out in their heads).

PR #523, opened 2017-02-23
_episodes/04-pipefilter.md: edit excercises with a file animals.txt

tinavisnovska:
1) Change content of `animals.txt` in exercises Which Pipe? and Appending Data to the content of file data-shell/data/animals.txt

The exercises work with the smaller input just fine and there is file data-shell/data/animals.txt with exactly the same content. I think it is better to refer directly to that file in the exercises.

2) Add location of a file `animals.txt` (data-shell/data/) to exercises which use the file

Exercises Pipe Reading Comprehension, Pipe Construction, Which
Pipe?, and Appending Data use file `animals.txt` but the location
of the file is not mentioned and so it is not entirely easy for
participants to run the commands themselves. When the location is
mentioned, it is easier to go and try to run the commands.

3) Rephrase text of exercise Appending Data

I find introduction of the exercise a bit confusing and suggest a few changes.

### DIFF
--- a/_episodes/04-pipefilter.md
+++ b/_episodes/04-pipefilter.md
@@ -639,7 +639,9 @@ so this matches all the valid data files she has.
 
 > ## Pipe Reading Comprehension
 >
-> A file called `animals.txt` contains the following data:
+> Explore a file called `animals.txt` located in `data-shell/data`
+> directory of the data downloaded for the shell lesson. The file
+> contains the following data:
 >
 > ~~~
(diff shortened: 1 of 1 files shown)


gcapes:
Great minds think alike! 
See 9e3f660e4068a41178d21f8ca91ea1ca7896be89 on #522

gdevenyi:
Thanks for your contribution, this bug has been fixed on #522

PR #524, opened 2017-02-23
Added a solution to 'Creating Files in a Different Way' in ep.03

mcadamme:


gcapes:
Thanks, this adds to #521. Could you correct the typo in solution 3?
> ... with it's output
 
should be 'its'.

gdevenyi:
Hi,

Can you add the solution labelling as per https://swcarpentry.github.io/lesson-example/04-formatting/

Thanks!

gdevenyi:
Thanks!

PR #525, opened 2017-02-27
fix typo (heirarchy) and minor tweaks

bbolker:


shwina:
Looks great, thanks @bbolker!

PR #526, opened 2017-02-27
Fix #518 for 05 loop

### LINKED ISSUE #518: Missing solutions to, context for, and files required by exercises

gcapes:

### DIFF
--- a/_episodes/05-loop.md
+++ b/_episodes/05-loop.md
@@ -597,13 +597,13 @@ so she decides to get some coffee and catch up on her reading.
 >
 > A loop is a way to do many things at once --- or to make many mistakes at
 > once if it does the wrong thing. One way to check what a loop *would* do
-> is to echo the commands it would run instead of actually running them.
+> is to `echo` the commands it would run instead of actually running them.
 > 
 > Suppose we want to preview the commands the following loop will execute
 > without actually running those commands:
 >
 > ~~~
-> for file in *.dat
+> for file in *.pdb
 > do
 >   analyze $file > analyzed-$file
 > done
@@ -615,7 +615,7 @@ so she decides to get some coffee and catch up on her reading.
 >
 > ~~~
 > # Version 1
-> for file in *.dat
+> for file in *.pdb
 > do
 >   echo analyze $file > analyzed-$file
 > done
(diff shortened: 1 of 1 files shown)


_episodes/05-loop.md
```diff
@@ -428,98 +428,142 @@ so she decides to get some coffee and catch up on her reading.
 
 > ## Variables in Loops
 >
-> Suppose that `ls` initially displays:
+> This exercise refers to the data-shell/molecules directory.
```
shwina: Backticks

_episodes/05-loop.md
```diff
@@ -428,98 +428,142 @@ so she decides to get some coffee and catch up on her reading.
...
 > done
 > ~~~
 > {: .bash}
 >
 > Why do these two loops give different outputs?
+>
+> > ## Solution
+> > The first code block gives the same output on each iteration through
+> > the loop.
+> > Bash expands the wildcard `*.pdb` within the loop body (as well as
```
shwina: How about explicitly writing out the expanded loop?

Bash expands the wildcards before the loop starts, so the loop is equivalent to:

~~~
for datafile in cubane.pdb  ethane.pdb  methane.pdb  octane.pdb  pentane.pdb  propane.pdb
do
    ls cubane.pdb  ethane.pdb  methane.pdb  octane.pdb  pentane.pdb  propane.pdb
done
~~~

_episodes/05-loop.md
```diff
@@ -428,98 +428,142 @@ so she decides to get some coffee and catch up on her reading.
...
 > ## Saving to a File in a Loop - Part Two
 >
-> In another directory, where `ls` returns:
->
-> ~~~
-> fructose.dat    glucose.dat   sucrose.dat   maltose.txt
-> ~~~
-> {: .output}
+> In the same directory, ensure you first delete `alkanes.pdb` if you created it by testing the code
+> from the previous challenge. Use the command `rm alkanes.pdb`.
```
shwina: Instead of this, can we use another filename than `alkanes.pdb` in the previous challenge? Such as `alkanes.txt`? This would keep the exercises independent of each other.

_episodes/05-loop.md
```diff
@@ -428,98 +428,142 @@ so she decides to get some coffee and catch up on her reading.
...
+> > 3 is the correct answer. `>>` appends to a file, rather than overwriting it with the redirected
+> > output from a command.
+> > Given the output from the `cat` command has been redirected, nothing is printed to the screen.
+> {: .solution}
 {: .challenge}
 
 > ## Limiting Sets of Files
 >
-> In the same directory, where `ls` returns (without the `sugar.dat` file):
+> In the same directory, delete the `alkanes.pdb` file if you completed the last exericse.
```
shwina: Same as above - change `alkanes.pdb` to `alkanes.txt`?

_episodes/05-loop.md
gcapes: Yes, I can change the file name, but nearly all these exercises start by confirming that the learner should keep in the same directory as the first exercise. As such they're not independent from each other. Should I change the beginning of each exercise to state which directory each exercise refers to?

gcapes:
Hi @shwina, 

I think I've addressed all the comments in my rebase and forced push.

Thanks

PR #527, opened 2017-02-27
Fix #518  for 06-script

### LINKED ISSUE #518: Missing solutions to, context for, and files required by exercises

gcapes:

### DIFF
--- a/_episodes/06-script.md
+++ b/_episodes/06-script.md
@@ -457,10 +457,26 @@ Of course, this introduces another tradeoff between flexibility and complexity.
 > ~~~
 > {: .source}
 >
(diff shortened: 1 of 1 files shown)


gcapes:
@shwina 
Just a bump, because there have been no comments on this yet.

shwina:
I'm sorry, @gcapes - I'll get to this tonight.

gcapes:
Oops! Yep, that's a mistake.
I'll rebase and fix it given it hasn't get been merged.
Cheers

gcapes:
Ok, done!

shwina:
Great! Thank you!

PR #528, opened 2017-02-27
Fix #518 for 07-find

### LINKED ISSUE #518: Missing solutions to, context for, and files required by exercises

gcapes:
Also add files and directories where using existing files wasn't possible.


gcapes:
@shwina 
Just a bump, because there have been no comments on this yet.

_episodes/07-find.md
```diff
@@ -514,20 +519,38 @@ about them."
 > wc -l $(find . -name '*.dat') | sort -n
 > ~~~
 > {: .bash}
+>
+> > ## Solution
+> > Find all files with a `.dat` extension in the current directory.
+> > Count the number of lines each of these files contains, then sort the output numerically.
+> {: .solution}
```
shwina: How about we list (enumerate) them:

1. Find all files with a `.dat` extension in the current directory
2. Count the number of lines in each of the above files
3. Sort the output from step (2) numerically

_episodes/07-find.md
```diff
@@ -579,13 +618,39 @@ about them."
...
+> > ```
+> > {: .source}
+> >
+> > Alternative, slightly inferior solution:
+> > ```
+> > for sis in Jo Meg Beth Amy
+> > do
+> > 	echo $sis:
+> >	grep -ocw $sis LittleWomen.txt
+> > done
```
shwina: I'd say let's leave out the inferior solution altogether. Any advantage to keeping it?

gcapes: Only that the question suggests two different ways to achieve the outcome. I don't mind removing the alternative solution, so long as we also remove the hint which suggests it.

Or I could add a short note explaining that there is often more than one way to complete a programming task.

shwina: Thanks, @gcapes - I'd suggest keeping the hint and adding the note.

gcapes: Done!

_episodes/07-find.md
gcapes: Done!

shwina:
Thanks!

PR #534, opened 2017-03-17
Created new instructions to setup identical environments for ALL less…

marskar:
…ons across Windows, MacOS and Linux using Anaconda or Miniconda; adapted UNIX lesson setup to use Jupyter QtConsole and/or Notebook

### DIFF
--- a/AnacondaSetup.md
+++ b/AnacondaSetup.md
@@ -1,0 +1,51 @@
+---
+layout: page
+title: Anaconda Setup
+permalink: /setup/
+---
+
+# Setup
+
+## Step 1: Install Anaconda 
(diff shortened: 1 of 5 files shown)


shwina:
@marskar Thanks for this PR! However, I think that this PR belongs in https://github.com/swcarpentry/workshop-template/. @gdevenyi?

gdevenyi:
Agree, this is for the style template

marskar:
Good morning,
Sorry for the mistake.
OK, I will move this to workshop-template.
Cheers,
Martin

shwina:
Thank you!

PR #536, opened 2017-03-23
Consider adding "tree command"

Wolflab:
I decided to add the “tree” command to the Shell novice lesson. I am not 100% certain that this would be worth incorporating (maybe it will not work on all shell versions?). But I think it is a useful command for beginners who are trying to get a rough concept of directory structure. It is a nice stepping stone from the GUI folders to file paths. But mostly wanted show I can use github effectively to qualify as a software carpentry instructor.

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -22,6 +22,7 @@ keypoints:
 - "`pwd` prints the user's current working directory."
 - "`whoami` shows the user's current identity."
 - "`/` on its own is the root directory of the whole file system."
+- "`tree` Displays the entire hierarchy of directories and files."
 - "A relative path specifies a location starting from the current location."
 - "An absolute path specifies a location from the root of the file system."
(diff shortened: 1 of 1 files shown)


shwina:
I may be wrong, but `tree` requires special installation of the `tree` package. Also, is it available on Windows?

Wolflab:
As far as I can tell, most newer shells are coming with the tree package, but it looks like several would need an install, making it less than ideal for software carpentry. That's a pity because I think the command is useful for beginners as well a s a good teaching tool. Oh well.

shwina:
Couldn't agree more! :(

gdevenyi:
Sadly tree isn't included in most Linux distros by-default, we'd like to stick to standard gnu/bsd tools included on the shell. ``find`` provides a (much uglier) alternative that can achieve similar goals and is included.

Thanks for your submission, maybe shell-extras might like it, but for now we'll have to decline for shell-novice.

PR #544, opened 2017-04-03
Solution-using-rm-safely

Marisa89:
Added solution to Using rm Safely (03-create.md).
This is also part of the completion of Software Carpentry instructor training.


gdevenyi:
Thanks!

Issue #545, opened 2017-04-04
Enhancement: show error for unknown commands


mondkaefer:
Section: http://swcarpentry.github.io/shell-novice/02-filedir/

After showing the first command (whoami) we could show what happens if you type non-existing commands. I remember when I started with the Shell many years back I was confused initially because I didn't fully understand what can be typed in the terminal and what not, and learned quickly that it had to be existing commands that i also had to type correctly.
Adding a little paragraph of the type like "Preparation Magic" to show what error message the shell prints if you type a non-existing command like "hello" or the like would perhaps be useful.

I'm happy to contribute if people think that's a good idea.

shwina:
Yes please. There is a distinction to make:

1. Using an existent command incorrectly
2. Using a non-existent command

If possible, could we show examples of both?

mondkaefer:
Yes, sounds good to me.
Question is then where to place these examples in the current tutorial. 
How about putting 2. immediately after the first command that is being used (whoami), and 1. (using an invalid option) right before "ls --help"?

mondkaefer:
Contribution has been merged.

PR #548, opened 2017-04-13
#545: Explain unknown commands or command-line parameters

mondkaefer:
#545: Explain unknown commands or command-line parameters


gdevenyi:
Thanks for the contribution, this looks good, but I think perhaps this should go in a callout box, can you add those changes?

mondkaefer:
Happy to do so.

gdevenyi:
Thanks for the fixup!

Issue #547, opened 2017-04-12
Found a v minor typo


cjieming:
Hi I am new here. 

Found a typo while trying to complete my instructor training checkout: http://swcarpentry.github.io/shell-novice/ > Under 'Schedule' > The line containing 'Setup' > Typo is "Download"; it is currently "Dowload".

shwina:
Thank you @cjieming - nice catch! That particular part of the lesson actually comes from another repository "styles":

https://github.com/swcarpentry/styles

and I notice that this issue was recently fixed there:

https://github.com/swcarpentry/styles/commit/12506a4b66016088a5c8582ac6e34ccb2cdb13f3

Periodically, we update this repository from the changes to the "styles" repository, so the next time that happens, this will be fixed.

cjieming:
Ok, I will close this issue then.

PR #551, opened 2017-04-19
Fix typo: Dowload --> Download

petebachant:


shwina:
Thank you @petebachant - good catch. But please see https://github.com/swcarpentry/shell-novice/issues/547

petebachant:
Ah, no worries. I had a hunch there was a template repo or something.

PR #553, opened 2017-04-19
Add template

ErinBecker:
Add template language about lesson contribution and maintenance process


ErinBecker:
I realized after submitting this PR yesterday that I didn't do a good job of documenting the rationale behind this or what it's trying to accomplish. My apologies.

I've been getting a significant number of questions from new instructors and instructor trainees about the lesson contribution process. Specifically, many trainees are worried that if their contributions aren't accepted they won't be able to qualify as instructors. Some are also worried or frustrated about not hearing back about their contributions for a while. I wanted to help communicate to contributors that the maintainers are volunteers who have many other time commitments and sometimes delay in reviewing PRs is inevitable, also that even if a specific contribution isn't incorporated, their input is still appreciated and welcome.

I'm hopeful that this language will help clarify the process and assuage some of these concerns, especially for new members of our community.

shwina:
Rather than opening an identical PR in each of the lesson repositories, I think it would be appropriate to have a single PR in https://github.com/swcarpentry/styles.

Otherwise, it might be a good idea to bring this up for discussion in the maintainers mailing list.

gdevenyi:
Agree with @shwina  this is material for the styles/lesson template.

ErinBecker:
Thanks @gdevenyi and @shwina. I wasn't sure how the template connected back up to the live lessons or if/when changes to the styles repo would propagate to the individual lessons. 

I've put the PR here: https://github.com/swcarpentry/styles/pull/143

PR #557, opened 2017-04-21
Fix #556 - illustrate effect of quoting variables

### LINKED ISSUE #556: Better example needed for "spaces in names" box
```
for filename in "red dragon.dat" "purple unicorn.dat"
do
    head -n 100 "$filename" | tail -n 20
done
```
These files don't exist, so this isn't a good demonstration. I used `ls` instead, with $filename and "$filename" to illustrate the difference: the former passes two arguments, the latter passes one.

gcapes:


gdevenyi:
Looks great, thanks!

PR #558, opened 2017-04-24
Add subheadings to introduction

alexmorley:
Intro is quite a large singular block of text that is difficult to get into. Maybe subheadings would be useful (not necessarily the ones I chose)?


gdevenyi:
An excellent idea, walls of text are bad!

PR #559, opened 2017-04-25
Delete `mvdir`

alexmorley:
No need to mention lack of `mvdir` command since we didn't mention `rmdir`. Alternatively we could mention both.


shwina:
Thanks!

PR #560, opened 2017-05-02
Add VIM, Emacs and bash notation for control characters.

Ianvdl:
The current list of control key notations is missing the one commonly used by VIM, Emacs and some man pages (such as the bash man page). This commit adds that notation.

This pull request serves to fulfill my instructor training checkout requirement.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -172,6 +172,7 @@ return to the shell.
 > * `Ctrl-X`
 > * `Ctrl+X`
 > * `^X`
+> * `C-x`
 >
 > In nano, along the bottom of the screen you'll see `^G Get Help ^O WriteOut`.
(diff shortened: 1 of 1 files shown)


Ianvdl:
An example from the bash man page:
                                                     
```
Readline Notation 

In this section, the Emacs-style notation is used to denote keystrokes.
Control keys are denoted by C-key, e.g., C-n means Control-N.  Similarly, meta
keys are denoted by M-key, so M-x means Meta-X.  (On keyboards without a meta
key, M-x means ESC x, i.e., press the Escape key then the x key.  This makes
ESC  the  meta  prefix.   The combination  M-C-x  means ESC-Control-x, or press
the Escape key then hold the Control key while pressing the x key.) 
```

PR #561, opened 2017-05-02
changed "enter" to be consistent across two sections

brattex:
The word 'enter' was used in one section and "Enter" (without the ""'s) the following paragraph.

I don't mind which is used but just make it consistent at least.


gdevenyi:
Thanks for the correction.

Can you please check the rest of the lessons for consistency with this?

In addition, perhaps quote 'return' as well?

gdevenyi:
Hi, have you had a chance to audit all the pages?

gcapes:
Fixed in #884

PR #563, opened 2017-05-04
Updated 04-pipefilter - learning objective for tee

kelleyo:
Added on to the first learning objective "and also show it on the screen simultaneously." (line 8),  added "tee" (line 22), and updated the lesson (lines 152-153). I don't believe this would add more than about 30 seconds to the current exercise. It just allows you to view and write at the same time, so one less command to type if you want to do both.

### DIFF
--- a/_episodes/04-pipefilter.md
+++ b/_episodes/04-pipefilter.md
@@ -5,7 +5,7 @@ exercises: 0
 questions:
 - "How can I combine existing commands to do new things?"
 objectives:
-- "Redirect a command's output to a file."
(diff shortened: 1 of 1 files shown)


gdevenyi:
Thanks for your contribution.

Instead of rewriting the first bullet point, can you add an additional one, as we also cover redirecting to file exclusively.

gdevenyi:
Thank you for your contribution.

Closing as stale because original submitter never responded with changes and lesson has drifted now.

Please feel free to submit a new PR or reopen if you would like to update this PR.

PR #564, opened 2017-05-04
Clarify how Ctrl-R searches in reverse chronological order

willfurnass:
It doesn't 'cycle' through matches in a truly cyclic fashion.


gdevenyi:
Wonderful, thanks!

PR #566, opened 2017-05-10
Replaced line: "It's pretty simple...  "

KirstyPringle:
Replaced Line:

It's pretty simple:
every alphanumeric character matches against itself.
After the pattern comes the name or names of the files we're searching in.

With:

The grep command searches through the file, looking for matches to the pattern specified.

To use it type grep, then the pattern we're searching for and finally the name of the file (or files) we're searching in.

(As I think it reads more clearly).

### DIFF
--- a/_episodes/07-find.md
+++ b/_episodes/07-find.md
@@ -74,11 +74,7 @@ Today it is not working
 ~~~
 {: .output}
 
-Here, `not` is the pattern we're searching for.
-
-The grep command searches through the file, looking for matches to the pattern specified.
-
-To use it type grep, then the pattern we're searching for and finally the name of the file (or files) we're searching in.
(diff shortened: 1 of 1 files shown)


_episodes/07-find.md
```diff
@@ -75,9 +75,11 @@ Today it is not working
 {: .output}
 
 Here, `not` is the pattern we're searching for.
-It's pretty simple:
-every alphanumeric character matches against itself.
-After the pattern comes the name or names of the files we're searching in.
+
```
rgaiacs: Remove this line.

_episodes/07-find.md
```diff
@@ -75,9 +75,11 @@ Today it is not working
...
+The grep command searches through the file, looking for matches to the pattern specified.
+
```
rgaiacs: Remove this line.

KirstyPringle:
Hi,  that's fine.  I have implemented the changes you suggest.

gdevenyi:
Thanks!

PR #570, opened 2017-05-18
remove judgmental language

ErinBecker:
Pointed out by a trainee at an instructor training workshop - some of the language in this episode is quite judgemental about using nano. This PR removes the judgmental language.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -135,7 +135,7 @@ $ nano draft.txt
 > not be powerful enough or flexible enough for the work you need to do
 > after this workshop. On Unix systems (such as Linux and Mac OS X),
 > many programmers use [Emacs](http://www.gnu.org/software/emacs/) or
-> [Vim](http://www.vim.org/) (both of which are unintuitive), 
+> [Vim](http://www.vim.org/) (both of which require more time to learn), 
 > or a graphical editor such as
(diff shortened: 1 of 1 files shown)


jttkim:
two very minor suggestions: (1) how about "simplest" rather than "least complex" and (2) would it make sense to replace "are unintuitive" with "require more time to learn"?

ErinBecker:
Great suggestions. I'm trying to avoid calling things "simple" as it
implies "easy" rather than "has fewer concepts/steps", which is what I'm
going for with "least complex".

I really like the change to "require more time to learn.

AFK now but will implement this when I get back to the ​computer. Thanks!

On May 18, 2017 5:05 PM, "Jan T. Kim" <notifications@github.com> wrote:

two very minor suggestions: (1) how about "simplest" rather than "least
complex" and (2) would it make sense to replace "are unintuitive" with
"require more time to learn"?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/swcarpentry/shell-novice/pull/570#issuecomment-302573789>,
or mute the thread

ChristinaLK:
I like this change!

gdevenyi:
Thanks!

ErinBecker:
Thanks!

PR #571, opened 2017-05-19
Fixed an exercise in the episode on loops

alex-konovalov:
Do not call variable `species` in a loop over alkanes.

This was spotted and fixed during the Software Carpentry workshop at St Andrews and was used as a demonstration how to submit a pull request.


gdevenyi:
Thanks!

alex-konovalov:
@gdevenyi thanks! We thought briefly whether this may be intentionally done in this exercise in a reference to a discussion on naming variables, but it seemed like that's not the main point of this exercise, and may be distracting the learner from the key point that the output file will be overwritten.

PR #574, opened 2017-05-19
Fixed bug in an episode 3 challenge question solution.

clarka34:
Solutions included a directory that was not specified in the original question, so I removed that directory from the solutions.

Original question: ## Copy a folder structure sans files


gdevenyi:
Good catch, thanks!

PR #575, opened 2017-05-19
Added solution to an episode 3 challenge question.

clarka34:
Added a solution to a challenge question in episode 3 that did not have a solution. Also, this question was used in episode 2 as well (not sure if that matters).

Original question: ## Listing Recursively and By Time


gcapes:
Hi @clarka34, and thanks for the PR!

Sorry no-one responded to this. This exercise now appears only [here in episode 2](http://swcarpentry.github.io/shell-novice/02-filedir/index.html#listing-recursively-and-by-time). It has a solution, but yours is more complete. Would you be interested in submitting a new PR to add your solution?

I've created issue #892 to summarise this.

Thanks again!

PR #578, opened 2017-05-19
Fixed solution in episode 7 challenge question.

clarka34:
Fixed solution for question ## Matching and Subtracting. The solutoins had a leading slash that caused them not to work correctly.


gdevenyi:
Thanks!

PR #579, opened 2017-05-24
Update index.md

dlebauer:
Modified to be more accurate and welcoming

---

---

### DIFF
--- a/index.md
+++ b/index.md
@@ -21,7 +21,8 @@ These lessons will start you on a path towards using these resources effectively
 > the word “file” and either “directory” or “folder” (two common words
 > for the same thing), you're ready for this lesson.
 >
-> If you're already comfortable manipulating files and directories,
+> Even if you're already comfortable manipulating files and directories,
 > searching for files with `grep` and `find`, and writing simple loops
-> and scripts, you probably won't learn much from this lesson.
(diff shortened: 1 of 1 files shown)


shwina:
We may be sacrificing correctness here. It *is* true that attendees already familiar with these topics may not get so much out of the workshop. Such attendees have sometimes expressed disappointment with the level of content, and we don't want them to have the wrong idea at the beginning of the lesson that we will be covering topics more advanced than these.

Thoughts? I'd be interested to hear from others as well.

dlebauer:
Good points but I don't think that the use of find and grep covers the scope of the material. Shell scripting and control flow are more advanced than these. 

Hopefully students can assess if the material is useful before they arrive.

From my perspective as an instructor and frequent user of the shell I usually learn / remember something when I read these lessons (of course students don't get all of the content). But it doesn't hurt to be reminded and see how other people use these tools.  Perhaps suggest that students who get bored look at man pages and learn something new.

gdevenyi:
I think this preface is appropriate and up-front about the target audience. If leaners already know those topics, shell-novice isn't for them.

gdevenyi:
Thanks for your discussion and contribution!

PR #582, opened 2017-05-31
episode 7 find output doesn't match current data-shell

pevans-gfz:
I've adjusted the sample `find` output in the text. Also noted `find` doesn't order its output, and fixed some trivial typos.


shwina:
Thanks!

PR #584, opened 2017-06-14
Add Solution to Listing Recursively Challenge

alexkassil:
The Listing Recursively and By Time challenge was the only one without a solution in episode 3, so I added a short one.


gdevenyi:
Thanks for the contribution!

alexkassil:
Glad to help such a useful resource!

PR #587, opened 2017-06-24
Made suggested changes to the Parameters vs. Arguemnts section of 02-…

carol-rowe666:
…filedir.md

The Unix Shell - Navigating Files and Directories

Parameters vs. Arguments

Previously in this lesson, you referred to the -F of the ls command as a flag. Then, in this section you call it either a parameter or an argument. As this lesson continues, you go back to referring to -F as a flag. I think this would be very confusing to beginners.

As I understand it, a flag or option is preceded by a single "-" or double "--" and can be left out. A command or parameter comes after a command and provides information to either the command or one of its options.

For example:

ls -F Desktop

ls is the command
-F is the option or flag
Desktop is the argument or parameter

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -420,15 +420,19 @@ without it,
 the shell thinks we're trying to run a command called `ls-F`,
 which doesn't exist.
 
-> ## Parameters vs. Arguments
+> ## Flag or Option vs. Parameters or Arguments
 >
-> According to [Wikipedia](https://en.wikipedia.org/wiki/Parameter_(computer_programming)#Parameters_and_arguments),
-> the terms **argument** and **parameter**
-> mean slightly different things.
-> In practice,
-> however,
-> most people use them interchangeably or inconsistently,
-> so we will too.
+> Although there may be technical differences between the terms, Flag and Option,
+> or between Parameter and Argument, most people use them interchangeably and 
+> sometimes inconsistently.
+> We will too.
+> In general, a flag or option is preceded by a single "-" or double "--" and can be 
+> left out. Alternatively, a command or parameter comes after a command and provides
+> information to either the command or one of its options.
+> As you will see in the next example: ls -F Desktop
+> ls is the command
+> -F is the option or flag
+> Desktop is the argument or parameter
 {: .callout}
 
 We can also use `ls` to see the contents of a different directory.  Let's take a


_episodes/02-filedir.md
```diff
@@ -420,15 +420,19 @@ without it,
 the shell thinks we're trying to run a command called `ls-F`,
 which doesn't exist.
 
-> ## Parameters vs. Arguments
+> ## Flag or Option vs. Parameters or Arguments
```
shwina: Perhaps let's just re-name this to "Flags, Options, Parameters or Arguments"

_episodes/02-filedir.md
```diff
@@ -420,15 +420,19 @@ without it,
...
 >
-> According to [Wikipedia](https://en.wikipedia.org/wiki/Parameter_(computer_programming)#Parameters_and_arguments),
-> the terms **argument** and **parameter**
-> mean slightly different things.
-> In practice,
-> however,
-> most people use them interchangeably or inconsistently,
-> so we will too.
+> Although there may be technical differences between the terms, Flag and Option,
```
shwina: How about rephrasing as:

Although there may be technical differences between the terms "flags" (or "options") and "parameters" (or "arguments"), most people use them interchangeably and sometimes inconsistently.

_episodes/02-filedir.md
```diff
@@ -420,15 +420,19 @@ without it,
...
+> or between Parameter and Argument, most people use them interchangeably and 
+> sometimes inconsistently.
+> We will too.
+> In general, a flag or option is preceded by a single "-" or double "--" and can be
```
shwina: Use backticks: `-` instead of "-".

_episodes/02-filedir.md
```diff
@@ -420,15 +420,19 @@ without it,
...
-> most people use them interchangeably or inconsistently,
-> so we will too.
+> Although there may be technical differences between the terms, Flag and Option,
+> or between Parameter and Argument, most people use them interchangeably and 
+> sometimes inconsistently.
+> We will too.
+> In general, a flag or option is preceded by a single "-" or double "--" and can be 
+> left out. Alternatively, a command or parameter comes after a command and provides
+> information to either the command or one of its options.
+> As you will see in the next example: ls -F Desktop
```
shwina: Use backticks with inline code or commands: `ls -F Desktop`, not just ls -F Desktop.

_episodes/02-filedir.md
```diff
@@ -420,15 +420,19 @@ without it,
...
+> ls is the command
```
shwina: `ls`

_episodes/02-filedir.md
```diff
@@ -420,15 +420,19 @@ without it,
...
+> -F is the option or flag
```
shwina: `-F`

_episodes/02-filedir.md
```diff
@@ -420,15 +420,19 @@ without it,
...
+> Desktop is the argument or parameter
```
shwina: `Desktop`

gdevenyi:
@carol-rowe666 have you had a chance to make the changes requested by @shwina

gcapes:
I think this is a good suggestion, and something I was thinking about too. I would leave the `argument vs parameter` discussion as it was because it is informative, but be more consistent when referring to the flags.

Correct me if I've got it wrong, but isn't the flag just a modifier for the command? So the command could either refer to `ls` or to `ls -F`, and the arguments are then what follow this (e.g. a filename or directory).

_episodes/02-filedir.md
gcapes: Sorry, @shwina  but I think this muddies the waters rather than clarifying. Most people don't use arguments and flags interchangeably -- it's parameter and arguments that get confused.

gcapes:
I should add, that I'm happy to make the changes and submit a PR if this one has gone cold?

gdevenyi:
Closing as stale. @gcapes has addressed changes requested in #618

PR #603, opened 2017-08-15
fixed spelling of 'comand' to 'command' in section 7, topic is Unsupp…

reshama:
…orted comand-line options

@naupaka (part of instructor training checkout)
cc:  @ErinBecker

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -370,7 +370,7 @@ Many bash commands, and programs that people have written that can be
 run from within bash, support a `--help` flag to display more
 information on how to use the commands or programs.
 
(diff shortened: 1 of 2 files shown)


reshama:
@naupaka (part of instructor training checkout)
cc: @ErinBecker

fixed this one:
>http://swcarpentry.github.io/shell-novice/07-find/
Here, not is the pattern we’re searching for. The grep command searches through the file, looking for matches to the pattern specified. To use it type grep, then the pattern we’re searching for and finally the name of the file (or files) we’re searching in.
--> type "grep" ; put grep in quotes

gdevenyi:
Wonderful thanks!

PR #604, opened 2017-08-16
Update 01-intro.md

jestill:
Changed goostat to goostats since that is the name used in other Episodes and goostats is the name of the program that is downloaded in shell-novice-data.zip


shwina:
Thanks!

PR #605, opened 2017-08-16
Solution to "Wildcard expressions": fix formatting, provide better answer

arlin:
Currently the solution is somewhat confusing for readers, because the numbered list is rendered as 1 1 2 instead of 1 2 3.  This is because a chunk of code appears between the first two answers and this disrupts the Markdown list syntax.  

I worked around this by putting the code chunk first.  

Also, I edited the first answer to point out that, although the solution seems to require 2 expressions, they can be executed in a single command: `ls *A.txt *B.txt` instead of `ls *A.txt` then `ls *B.txt`.  This will not generate an error in the case where one of the filename patterns has an empty expansion.

### DIFF
--- a/_episodes/04-pipefilter.md
+++ b/_episodes/04-pipefilter.md
@@ -816,13 +816,12 @@ so this matches all the valid data files she has.
 >     where the original one would not?
 >
 > > ## Solution
-> > 1. 
-> >
 > > ```
 > > $ ls *A.txt
 > > $ ls *B.txt
 > > ```
 > > {: .bash}
+> > 1. Two expressions can be run as separate commands (or as one: `ls *A.txt *B.txt`). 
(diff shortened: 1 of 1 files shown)


shwina:
I think this can be fixed using 4 spaces at the beginning of each line that includes the code block. The following should render correctly (with proper numbering):

~~~
1. Test text
    ~~~~
    Test code
    ~~~~
  
1. Test text 2
~~~

arlin:
I can't fix this because I do not know how the rendering of these accordions is done (this is not just MarkDown).  I will open up two separate issues for this, one about the botched formatting, and the other about the improved answer.

arlin:
I turned this into 2 separate issues, #614 (list numbering) and #615 (better solution).  Closing this pull request.

Issue #585, opened 2017-06-20
Using nano to create empty file 


brooksph:
Under the section "Let’s re-create that file and then move up one directory to /Users/nelle/Desktop/data-shell" you're instructed to use nano to create a file named draft.txt. The next step uses ```ls``` to list the files in the directory. No instructions are provided for exiting nano and because the no text is added to file it is not saved.

marwahaha:
Hmm. So in this section:
![image](https://user-images.githubusercontent.com/2541209/27620348-057857c2-5b97-11e7-9a90-6e13897c5ce0.png)

It wouldn't hurt to add a graphic showing the `nano` screen, and adding some text, and then the commands to quit. Or, we could add a comment after `nano draft.txt` saying (make sure the learners enter text in the file, and close the file with `Ctrl-X`, then `y`, then `Enter`.

gdevenyi:
Hi, contributions of this type are welcome to enhance the lesson.

reshama:
@marwahaha @gdevenyi 
Would it be easier to introduce `touch` command and create a blank file?

aliceminotto:
I tend to agree with @reshamas, I've always used `touch` a lot to create placeholder files I was going to write in later. I think it would also make way easier to show examples or giving some small exercises during the lecture (e.g. moving, renaming and copying files, or even show some wildcard behaviour later on). There's no need to explain about the access and modification time neither in my opinion.

scttfrdmn:
While the suggestion to introduce `touch` appears to have been addressed through #606 the original issue remains unresolved. I just stumbled through this during instructor training the other day. This issue appears twice in the episode. Since the purpose is to regenerate the file again maybe just a quick notation is needed and I propose the following.

In the referenced "Deleting Is Forever" section:

>    ```bash
>    $nano draft.txt
>    ```
>
>Let’s again type in a few lines of text. Once we’re happy with our text, we can press ```Ctrl-O``` to write our data to disk. Once our file is saved, we can use ```Ctrl-X``` to quit the editor and return to the shell.
>
>    ```bash
>    $ls
>    ```

and then later in the "With Great Power Comes Great Responsibility" section:

>    ```bash
>    $mkdir thesis
>    $nano thesis/draft.txt
>    ```
>
>Let’s again type in a few lines of text. Once we’re happy with our text, we can press ```Ctrl-O```  to write our data to disk. Once our file is saved, we can use ```Ctrl-X``` to quit the editor and return to the shell.
>
[...]

PR #606, opened 2017-08-17
#Instructor check out#episodes of 02,03,04,07 were modified and some suggestions provided based on teaching

Cacagator:
#       modified:   02-filedir.md
Line 194: add "Assume" before "underneath `/Users`," to avoid learners looking for training data folder "data-shell" 
Line 243-250: add "man ls" as an option to Mac, because "ls --help" working on Linux, but invalid on Mac 
#       modified:   03-create.md
Line 126-137: add "touch draft.txt", considering empty file can't be saved by nano Line 151: introduced Windows version nano 
Line 286-293: add "rmdir" command
#       modified:   04-pipefilter.md
 Line 834: replace "The file `data-shell/data/animals.txt` contains 586 lines" to "The file `data-shell/data/animals.txt` contains 8 lines" 
#       modified:   07-find.md
change of Line 834 of "04-pipefilter.md" should also refected on Line 605 of "07-find.md"


_episodes/03-create.md
```diff
@@ -273,6 +283,14 @@ $ rm -r thesis
...
 {: .bash}
 
+However, if the `thesis` directory is empty, we can use `rmdir thesis` to remove the directory
+
+~~~
+$ rm thesis/draft.txt
+$ rmdir thesis
+~~~
+{: .bash}
+
```
shwina: -1. Let's not introduce a new command `rmdir` without an explanation for why it's preferable to `rm -r`

_episodes/03-create.md
```diff
@@ -138,9 +148,9 @@ $ nano draft.txt
 > [Vim](http://www.vim.org/) (both of which require more time to learn), 
 > or a graphical editor such as
 > [Gedit](http://projects.gnome.org/gedit/). On Windows, you may wish to
-> use [Notepad++](http://notepad-plus-plus.org/).  Windows also has a built-in
-> editor called `notepad` that can be run from the command line in the same
-> way as `nano` for the purposes of this lesson.  
+> use [Notepad++](http://notepad-plus-plus.org/) or [nano](https://github.com/lhmouse/nano-win).
```
shwina: We are trying to give learners examples of text editors other than `nano` here.

shwina:
Thank you!

_episodes/03-create.md
Cacagator: I'm not trying to argue here. Please note that not every leaner can make "nano" work as they or this tutorial wish (considering OS differences, etc) if you are really doing the teach. And every one wants to experience what we are doing and what's happening.  If the example not working, what is the point of example :)

shwina: I agree, and sorry - I should have been more clear: I only mean that the purpose of this callout is expressly to give learners examples of **other** (more powerfuly/friendly) editors than `nano`:

> However, because of this trait, it may not be powerful enough or flexible enough for the work you need to do after this workshop

So mentioning `nano` as an example for Windows is unnecessary.

_episodes/03-create.md
```diff
@@ -123,6 +123,16 @@ then run a text editor called Nano to create a file called `draft.txt`:
...
+$ ls -F
+~~~
+{: .bash}
+
+Apparently, if we create empty file using nano, the file won't be saved.
+But we can use command `touch` to create a file called `draft.txt`:
+
+~~~
+$ touch draft.txt
+$ ls -F
```
aliceminotto: I'd be happy to see `touch` in the lecture, as per issue #585 , I believe it needs some more words though.
I'm not sure I understand what you mean by "apparently if we create empty file using nano the file won't be saved"?
If you go through the usual write out>file name>exit the empty file is indeed created, definitely more time consuming and more complicated than needed IMO, but the file is saved.

Cacagator: Yes, you are right! I didn't explore it enough, the file can't be save by using "cntrl + x", but do with "cntrl +o" as you pointed out.

I still think nano is kinda of challenge (most people don't how to get out for their fist time), if you this material is focus on learners with zero background CML interface. 

Thanks!

gdevenyi:
Hi,

Thanks for your contribution.

Can you please split these changes into separate PRs, so they can be evaluated and merged separately. Thanks!

PR #609, opened 2017-08-29
Update 01-intro.md

grhea:
Tweaked wording in the intro since direct brain interface has only been accomplished in the lab and is not common.

Instructor checkout


gdevenyi:
@grhea Can you update with the changes requested by @shwina thanks!

shwina:
Thank you!

Issue #538, opened 2017-03-27
goodiff and goostats do not give usage instructions when used without arguments


mbeyeler:
The two bash scripts `goodiff` and `goostats` in [data-shell/north-pacific-gyre/2012-07-03](https://github.com/swcarpentry/shell-novice/tree/gh-pages/data-shell/north-pacific-gyre/2012-07-03) tend to confuse workshop participants. It would help if the scripts' usage would be explained [on the website](http://swcarpentry.github.io/shell-novice/06-script/).

In addition, it would be great if the scripts printed usage information to stdout whenever people attempt to use them without input arguments (as is common with bash scripts).

shwina:
+100.

gcapes:
Fixed in #616

PR #616, opened 2017-09-18
Issue538

AnjaLeBlanc:
#538 

Modified the two shell scripts 'goostats' and 'goodiff' by checking input arguments and parameters. When using them with the wrong number of arguments the scripts produce some help text.
Added a couple of lines to the story about the scripts. (As the don't do anything useful without too much detail!) 

(Pull request as part of Instructor training checkout process)


_episodes/06-script.md
```diff
@@ -320,6 +320,9 @@ The file `redo-figure-3.sh` now contains:
 ~~~
 {: .source}
 
+
+(goostats is a shell script which calculates some complicated statistics from a datafile -- the first parameter -- and writes them to a file -- the second parameter; the script goodiff compares two datafiles provided as parameters. Nele's superviser provided them without too many explainations.)
```
gcapes: Please change ~parameter~ argument. We've just tidied up the usage of argument vs parameter in the shell lesson #624.
Could you also use backticks for `goostats` and `goodiff`.
And fix ~Nele's~ Nelle's

shwina:
Perhaps this is a good opportunity to get rid entirely of the optional arguments in `goostats`. We don't use them in the lesson as far as I'm aware. @AnjaLeBlanc, @gcapes what do you think?

AnjaLeBlanc:
@shwina -- the optional args don't do anything but they are mentioned in the story line (Chapter 6 -- Nelle’s Pipeline: Creating a Script). But I agree they are not essential and could shorten the lesson by a few seconds.

shwina:
Thanks @AnjaLeBlanc! Removing them also has the added benefit of making the `goostats` script actually readable. Often after the exercise I will open the script up to explain what it is doing (i.e., a few operations like `uniq` and `cut` that we have seen before and then sleeping for 2 seconds).

I'd be in favour of removing them entirely.

gcapes:
As Anja mentioned, the extra flags are the justification given in the story as to why Nelle needs to re-run the script. 

If we remove this part of the story, we're re-running the script for no reason. I think we need something in the story to justify re-running the script - those extra flags were ideal. So my vote is to keep the flags, or remove them and add a different credible justification for re-running the script.

shwina:
@gcapes Thanks! In practice I do not use these switches. Instead I will ask learners to imagine they wille obtain a few more files the next day and will have to re-do the analysis: having a script can avoid re-typing the same commands.

@gcapes, if you think that's a good enough motivation, @AnjaLeBlanc, could I please request you to add it as part of the lesson narrative?

@AnjaLeBlanc thanks for all your contribution so far!

gcapes:
@shwina The problem with that reasoning is that you would only want to process the new files rather than reprocess the old files too (so you're implicitly suggesting bad practice to the learners). It doesn't seem a good enough justification - sorry! I think the original story should stay.

data-shell/north-pacific-gyre/2012-07-03/goostats
```diff
@@ -1,24 +1,8 @@
...
-done
-
-shift $((OPTIND-1))
-
 # check for the right number of input arguments
 if [ $# -ne 2 ]
   then
-    echo "goodstat [-J number] [-r] file1 file2"
-    echo "call goostat with two arguments it accepts the optional options -J with a number and -r"
+    echo "goodstat file1 file2"
```
shwina: `goodstat` -> `goostats`

_episodes/06-script.md
```diff
@@ -312,14 +312,17 @@ $ history | tail -n 5 > redo-figure-3.sh
...
+297 bash goostats NENE01729B.txt stats-NENE01729B.txt
 298 bash goodiff stats-NENE01729B.txt /data/validated/01729.txt > 01729-differences.txt
 299 cut -d ',' -f 2-3 01729-differences.txt > 01729-time-series.txt
 300 ygraph --format scatter --color bw --borders none 01729-time-series.txt figure-3.png
 301 history | tail -n 5 > redo-figure-3.sh
 ~~~
 {: .source}
 
+
+(`goostats` is a shell script which calculates some complicated statistics from a datafile -- the first argument -- and writes them to a file -- the second argument; the script `goodiff` compares two datafiles provided as arguments. Nelle's superviser provided them without too many explainations.)
```
shwina: 1. I suggest renaming datafile (not a word) to either "input file", "file" or "protein sample file".

2. I also suggest breaking up the first sentence into two, as `goostats` and `goodiff` are sufficiently different scripts. `goodiff` is not supposedly comparing files, but rather "Compar[ing] the statistics for each protein with corresponding statistics for each other protein ". I would re-word as "`goodiff` is a script for comparing statistics between two files, provided as arguments".

3. A few small typos: superviser->supervisor, explainations->explanations

data-shell/north-pacific-gyre/2012-07-03/goostats
```diff
@@ -1,11 +1,10 @@
...
-    case "$option" in
-    J) Jarg="$OPTARG"
-       shift $((OPTIND));;
-    r) shift $((OPTIND-1));;
-    esac
-done
+# check for the right number of input arguments
+if [ $# -ne 2 ]
+  then
+    echo "goodstat file1 file2"
```
shwina: Small typo: `goodstat` -> `goostats`

AnjaLeBlanc: @gcapes 
How about this justification of the script: 
Nelle's supervisor insisted that all her analytics must be reproducible. The easiest way to capture all the steps is in a script.

AnjaLeBlanc:
@gcapes
How about this justification of the script:
Nelle's supervisor insisted that all her analytics must be reproducible. The easiest way to capture all the steps is in a script.

gcapes:
Yep, that sounds perfect!

shwina:
Looks good to me! @gcapes any suggestions before merge?

gcapes:
Merge, merge, merge! :)

shwina:
Thanks @AnjaLeBlanc for all your work on this!

gdevenyi:
Wonderful work, I appreciate the careful coordination of edits to both the code and the narrative.

PR #618, opened 2017-09-18
Use `flag` consistently in episode 2

gcapes:
I didn't think so, but it looks like flags technically count as
arguments to a command.
However, introducing them as flags then referring to them as arguments
is confusing. So I have reworded to avoid calling them arguments.
This also sidesteps the possibly opinion-based question of
"Does a flag count as an argument to a command?".

Fix #587

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -225,7 +225,8 @@ system and how you have customized your filesystem.)
 `ls` prints the names of the files and directories in the current directory in
 alphabetical order,
 arranged neatly into columns.
-We can make its output more comprehensible by using the **flag** `-F`,
+We can make its output more comprehensible by using the **flag** `-F`
+(also known as a **switch** or an **option**) ,
 which tells `ls` to add a trailing `/` to the names of directories:
 
 ~~~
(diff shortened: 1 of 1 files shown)


gdevenyi:
Thanks, that looks good, but can you do a little updaing to https://github.com/swcarpentry/shell-novice/blob/gh-pages/_episodes/02-filedir.md#parameters-vs-arguments as well?

gcapes:
Ok great, glad you think so too!

I'm sorry, I don't follow what you're asking. I'm happy to make them, but what changes are you looking for here?

I was thinking of making a further PR which changes the usage of parameters and arguments to be correct, rather than following the current practice of "everyone else is sloppy with their terminology so we won't bother to use it correctly either". Standby for that ;)

gdevenyi:
Right now, I'd be happy just to include "flags" in the descriptions of that section.

Lets do another PR to discuss unformizing the usage in general.

gcapes:
Sorry if I'm being slow, but I still don't understand what you're asking for. Are you suggesting a note on `Flags vs Arguments vs Parameters`?

gcapes:
I've spotted a few more edits I should have made, so I'll amend this commit. I'll raise a separate PR to address the usage of arguments and parameters.

gcapes:
Thanks - all my other PRs have been merged. Could you tell me what edits you'd like me to make in order to get this one merged?

gcapes:
@gdevenyi

gcapes:
@shwina Fancy taking a look at this and making it 3/3 ;) ?

gdevenyi:
Wonderful thanks!

gcapes:
Thanks!

PR #627, opened 2017-09-22
Remove output files before rerunning goostats loop

### LINKED ISSUE #625: Nelle cancels the run, but now has extra `stats-` files which are matched
I'll submit a PR as this has caught me twice now. We just need to delete the `stats-*` files before rerunning the loop.

gcapes:
Fix #625.


gcapes:
Thanks!

PR #630, opened 2017-10-06
Correction to exercise solution

csqrs:
The exercise "Listing Recursively and By Time" has an incorrect solution. Directories are not listed alphabetically by 'ls -R -t'.


gdevenyi:
Good catch. Thanks.

PR #632, opened 2017-10-10
Update 05-loop.md: lower case move commands

csqrs:
'Emacs' style move commands should be lower case, for these two, 'C-a' and 'C-e'.

....sorry about forgetting to delete the boiler plate message in my last submit. ;)


gdevenyi:
Oh good catch, in some tools this matters and people might be too literal.

PR #633, opened 2017-10-11
Update 05-loop.md: fix "Doing a Dry Run" exercise

csqrs:
Fix names of generated files in "Doing a Dry Run" exercise. In "version 1", the
names of the generated files are incorrect in the text.


shwina:
Thank you!

PR #634, opened 2017-10-11
Update 07-find.md to fix capitalization

kellyrowland:
In the solution to the first question about "Little Women", the file is not properly capitalized (and so the command would not work). This change corrects the capitalization of the file name.


shwina:
Thank you!

PR #636, opened 2017-10-13
Update 06-script.md: corrected file location

csqrs:
Correction to location of sample file 'animals.txt' that contains counts. The file 'data-shell/data/animals.txt' does not contain counts.


gdevenyi:
Good catch, thanks!

PR #640, opened 2017-10-17
Update 07-find.md: correction to description of contents of 'writing' directory

csqrs:
Correction to description of contents of the 'writing' directory. There is also the directory 'writing/old',
which contains '.gitkeep'.

Also, the image of the directory tree is incorrect in the same way. Looks like a previous version of this
episode did not have this directory and file.


gdevenyi:
Hi @csqrs the ``.gitkeep`` files are needed for us to have "empty" directories within this repository. They're not intended to be included in the lesson.

Can you perhaps instead make a small note that effect?

shwina:
LGTM

gdevenyi:
Great, thanks!

Issue #615, opened 2017-09-13
Better solution to "Wildcard expressions"


arlin:
The exercise is written in a way that assumes that `ls *A.txt` followed by `ls *B.txt` is the best answer to searching for the desired files without using the `[AB]` syntax.   The questions are framed based on this solution, e.g., the user is given a hint to consider two expressions.  

However, the best solution to forming a query without using  is simply `ls *A.txt *B.txt`, and this does not give different results than `ls *[AB].txt`, which renders parts 2 and 3 moot.  

Rewrite both the prompt and the answers to take into account the best answer.   The new hint might be something like "Hint: you may need more than one expression, but you don't need more than one command".

paulamool:
This exercise shows that  `ls *[AB].txt` does not throw an error on missing file expressions.
Could add `ls *A.txt *B.txt` to solution point 1, and add extra information to point 3 to clarify this.

1. The following example expressions match the same set of files
```
  $ ls *A.txt
  $ ls *B.txt
```
or
  ```
$ ls *A.txt *B.txt
 ```
2. The output from the two `ls` commands is separated because there are two commands.
3. You would get an error message for the commands in point 1 when there are no files ending in `A.txt`, or there are no files ending in  `B.txt`. While `ls *[AB].txt` would not produce an error message, unless all possible expressions are missing.

PR #647, opened 2017-10-25
04-pipefilter Wildcard Expressions-solution added content to point 1 …

paulamool:
Issue #615
This contribution is for instructor training
Edited lesson 04-pipefilter, Wildcard Expressions - Solution, added content to point 1 and 3

### DIFF
--- a/_episodes/04-pipefilter.md
+++ b/_episodes/04-pipefilter.md
@@ -802,10 +802,10 @@ so this matches all the valid data files she has.
 > this.
 >
 > 1.  Can you match the same set of files with basic wildcard expressions
->     that do not use the `[]` syntax? *Hint*: You may need more than one
+>     that do not use the `[]` syntax? *Hint*: You can use more than one
 >     expression.
 >
-> 2.  The expression that you found and the expression from the lesson match the
+> 2.  The expressions that you found and the expression from the lesson match the
 >     same set of files in this example. What is the small difference between the
 >     outputs?
 >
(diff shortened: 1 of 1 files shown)


_episodes/04-pipefilter.md
```diff
@@ -802,27 +802,35 @@ so this matches all the valid data files she has.
...
 >
 > 1.  Can you match the same set of files with basic wildcard expressions
->     that do not use the `[]` syntax? *Hint*: You may need more than one
+>     that do not use the `[]` syntax? *Hint*: You can use more than one
 >     expression.
 >
-> 2.  The expression that you found and the expression from the lesson match the
+> 2.  The expressions that you found and the expression from the lesson match the
 >     same set of files in this example. What is the small difference between the
 >     outputs?
```
colinmorris: As you mentioned in #615, this question is clearly written with the "two `ls` commands" solution in mind, but for someone who comes up with the single `ls` solution, this question has no answer. I'd suggest relaxing the wording ("is there any difference" rather than "what is the difference"), or just removing this question, and mentioning the difference in the solution section.

_episodes/04-pipefilter.md
```diff
@@ -802,27 +802,35 @@ so this matches all the valid data files she has.
...
 > > 	```
 > >	{: .bash}
-> > 2. The output from the new commands is separated because there are two commands.
-> > 3. When there are no files ending in `A.txt`, or there are no files ending in
-> > `B.txt`.
+> > 
+> > 	or
+> > 
+> > 	```
+> > 	$ ls *A.txt *B.txt
```
colinmorris: I'd actually suggest putting this solution first, since it's "nicer".

_episodes/04-pipefilter.md
```diff
@@ -802,27 +802,35 @@ so this matches all the valid data files she has.
...
+> > 	```
+> > 	{: .bash}
+> > 2. The output from the two `ls` commands is separated because there are two commands.
+> > 3. You would get an error message for the commands in point 1 when there are no files
+> >    ending in `A.txt`, or there are no files ending in  `B.txt`. While `ls *[AB].txt would
+> >    not produced an error message, unless all possible expressions are missing.
```
colinmorris: > unless all possible expressions are missing

This wording feels a little awkward. Maybe "unless no files match" instead? "unless no files end in either A or B"?

gdevenyi:
Thank you for your contribution.

Closing as stale because original submitter never responded with changes and lesson has drifted now.

Please feel free to submit a new PR or reopen if you would like to update this PR.

PR #654, opened 2017-10-31
Move explanation of goostats to loops episode

### LINKED ISSUE #653: Description of goostats script is out of place
In episode 6 we have this, which comes in the middle of an explanation of making a shell script using the history command:
> (goostats is a shell script which calculates some complicated statistics from a protein sample file – the first argument – and writes them to a file – the second argument. goodiffis a shell script for comparing statistics between two files, provides as arguments. Nelle’s supervisor provided them without too many explanations.)

This would be better somewhere in episode 5.

gcapes:
This is where the shell script is first reintroduced so it makes
sense to put the explanation here.

Fix #653.

### DIFF
--- a/_episodes/05-loop.md
+++ b/_episodes/05-loop.md
@@ -292,10 +292,15 @@ judicious use of `echo` is a good debugging technique.
 
 ## Nelle's Pipeline: Processing Files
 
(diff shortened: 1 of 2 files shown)


shwina:
Thanks!

PR #656, opened 2017-11-02
Instructor note cards: Instructor Checkout

ztrautt:
For my instructor checkout procedure, I have decided to attempt to address an issue that I have observed when I was both a student taking Software Carpentry and as an instructor teaching it. When I was a student and my organization hosted an internal Software Carpentry, the instructor used two laptops, one for presentation and one with the lesson they were following (e.g., http://swcarpentry.github.io/shell-novice/05-loop/). With this method, I noticed that the instructors spent a lot of time scrolling up and down a large webpage trying to find their place. The next time my organization hosted an internal Software Carpentry, I taught shell under the supervision of a certified instructor. During this session, the instructors experimented with printing the lessons via a markdown-to-PDF approach. This produced more problems than using two laptops as above.

I have attempted to address the above issue by making an initial attempt at instructor notecards formatted for better presentation when an instructor hosting a lesson. These note cards are not intended to be seen by the students. That is, the instructor would have these on their secondary laptop or mobile device. The format is based on https://remarkjs.com/ I used my own note cards when my organization hosted an internal Software Carpentry. When using the cards, I found it much easier to keep a good cadence and return to where I left the lesson after answering questions about material not addressed in the lesson.

I can imagine this idea might be controversial. However, if this idea is supported, it is probably better to modify the source content in such a way that instructor note cards can be generated via a script. It also might be a good idea to consider student note cards, where the helpful images and quizzes are formatted properly for presentation screens.

Thank you for your consideration.

### DIFF
--- a/instructor-note-cards/01-intro.html
+++ b/instructor-note-cards/01-intro.html
@@ -1,0 +1,180 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>1. Introducing the Shell</title>
+    <meta charset="utf-8">
+    <style>
+      @import url(https://fonts.googleapis.com/css?family=Yanone+Kaffeesatz);
+      @import url(https://fonts.googleapis.com/css?family=Droid+Serif:400,700,400italic);
+      @import url(https://fonts.googleapis.com/css?family=Ubuntu+Mono:400,700,400italic);
+
+      body { font-family: 'Droid Serif'; }
+      h1, h2, h3 {
+        font-family: 'Yanone Kaffeesatz';
+        font-weight: normal;
+      }
+      .remark-code, .remark-inline-code { font-family: 'Ubuntu Mono'; }
+    </style>
+  </head>
+  <body>
+    <textarea id="source">
+
+class: center, middle
+
+# 1. Introducing the Shell
+
+---
+
+# Background
+At a high level, computers do four things:
+
+-   run programs
+-   store data
+-   communicate with each other, and
+-   interact with us
+
+They can do the last of these in many different ways,
+including direct brain-computer interfaces and speech recognition, using systems such as Alexa or Google Home.
+While such hardware interfaces are becoming more commonplace, most interaction is still
+done using screens, mice, touchpads and keyboards.
+Although most modern desktop operating systems communicate with their human users by
+means of windows, icons and pointers, these software technologies didn't become
+widespread until 1980s. The roots of such *graphical user interfaces*  go back
+to Doug Engelbart's work in the 1960s, which you can see in what has been
+called "[The Mother of All Demos](http://www.youtube.com/watch?v=a11JDLBXtPQ)".
+
+---
+
+# The Command-Line Interface
+
+- Early days: interact with computers by rewiring them
+- Between 1950s - 1980s: input/output via characters on a standard keyboard
+    - This kind of interface is called a **command-line interface**, or CLI
+    	- Compare to **graphical user interface**, or GUI
+    - The heart of a CLI is a **read-evaluate-print loop**, or REPL
+
+Going back even further,
+the only way to interact with early computers was to rewire them.
+But in between,
+from the 1950s to the 1980s,
+most people used line printers.
+These devices only allowed input and output of the letters, numbers, and punctuation found on a standard keyboard,
+so programming languages and software interfaces had to be designed around that constraint.
+
+This kind of interface is called a
+**command-line interface**, or CLI,
+to distinguish it from a
+**graphical user interface**, or GUI,
+which most people now use.
+The heart of a CLI is a **read-evaluate-print loop**, or REPL:
+when the user types a command and then presses the Enter (or Return) key,
+the computer reads it,
+executes it,
+and prints its output.
+The user then types another command,
+and so on until the user logs off.
+
+---
+
+# The Shell
+
+- The shell is a program like any other
+- However the shell is special because its job is to run other programs
+- The most popular shell is bash as it is default on most unix/linux packages
+
+This description makes it sound as though the user sends commands directly to the computer,
+and the computer sends output directly to the user.
+In fact,
+there is usually a program in between called a
+**command shell**.
+What the user types goes into the shell,
+which then figures out what commands to run and orders the computer to execute them.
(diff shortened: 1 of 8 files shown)


gcapes:
I also find it difficult to navigate the text of a given webpage in order to teach the lesson.

In one of my own lessons (a modified and extended Software Carpentry lesson), I make sure that everything that is needed to teach the lesson is in a code block, with comments if appropriate. That way I don't have to read the text of the lesson --- I can just scroll to the next code block. I would be in favour of making this modification to the official software carpentry lessons.

I like your idea of a neater set of reference notes, but from a quick look it seems that you have just broken up the lesson text into smaller chunks, which I wouldn't find particularly helpful.

iamc:
I do paper print and spiral bind the lessons. I go through them for reviewing and write/mark quite a bit over them (see below). Things like:

- mark the code to type to ease finding my place on the lesson (same spirit to what @gcapes commented)
- add notes/sketches about my blackboard drawings
- note things I want to emphasize
- mark which examples to do eg. depending on the timing ("do if T<10:45h")
- timings at the end of every section
- sticky notes barriers
 
And so on... It has worked very good for me so far for my last ~10 lessons, since I came up with the "concept".

If something like this could be done in the computer and bring it in a pdf, pad or similar it'd great, but by the moment I'll come back to this old fashioned way :-)

bash_notes-small

shwina:
All, I believe this is an issue to be discussed at the level of general lesson layout. It would serve little purpose if instructor notecards are available only for the Shell part of a Software Carpentry workshop. Before review of the implementation, I think there should be a discussion about if/how/where these should be part of the lesson.

@gdevenyi Thoughts?

gdevenyi:
I agree with @shwina, this is an excellent idea and we appreciate your work, but it looks like something that could be perhaps automated in the overall SWC styles repo.

Pulling in @rgaiacs

ztrautt:
I appreciate the discussion and agree with what has been said. I would love to see SWC scrap what I have done and move forward from scratch with the concept across the lesson portfolio. I agree with @gcapes that the note cards would need more in-depth work than the simple chopping I performed.

rgaiacs:
@ztrautt Thanks for the issue. I looked the issue on the day that you logged it but I was giving time for others to comment before I could register my thoughts.

> I noticed that the instructors spent a lot of time scrolling up and down a large webpage trying to find their place. The next time my organization hosted an internal Software Carpentry, I taught shell under the supervision of a certified instructor. During this session, the instructors experimented with printing the lessons via a markdown-to-PDF approach. This produced more problems than using two laptops as above.

I don't want to discredit you or the instructors that you mentioned! I know that this is a issue but it is also related with instructors' personal experience and practices. Some users are more comfortable using two monitors or a paper copy of the lesson than others.

> I have attempted to address the above issue by making an initial attempt at instructor notecards formatted for better presentation when an instructor hosting a lesson. These note cards are not intended to be seen by the students. That is, the instructor would have these on their secondary laptop or mobile device. The format is based on https://remarkjs.com/ I used my own note cards when my organization hosted an internal Software Carpentry. When using the cards, I found it much easier to keep a good cadence and return to where I left the lesson after answering questions about material not addressed in the lesson.

This is in the lines of the slides that we had a while back. The problem with the slides is that they quickly got out of date in relationship with the "default" lesson. At some point we decided to drop the slides because they was being more harmful than helpful for us.

> it is probably better to modify the source content in such a way that instructor note cards can be generated via a script.

This would be the ideal solution but it would goes against our principle of making the lesson easy to edit. For example, in https://raw.githubusercontent.com/swcarpentry/shell-novice/gh-pages/_episodes/01-intro.md we could have

~~~
### Background
<div class="card">
At a high level, computers do four things:

-   run programs
-   store data
-   communicate with each other, and
-   interact with us
</div>

~~~

instead of

~~~
### Background
At a high level, computers do four things:

-   run programs
-   store data
-   communicate with each other, and
-   interact with us

~~~

so that our script (preferable a Javacript client-side) could create the card fetching the `<div class="card">` elements which would not change the visual of the default lesson. **But we are trying to avoid using HTML elements in the lesson!*** We make only one exception, see https://github.com/swcarpentry/styles/pull/165, and I would prefer to not add another one because after we have two I will ask for the third one that is replace

~~~
> ## Preparation Magic
>
> If you type the command:
> `PS1='$ '`
> into your shell, followed by pressing the 'enter' key,
> your window should look like our example in this lesson.  
> This isn't necessary to follow along (in fact, your prompt may have
> other helpful information you want to know about).  This is up to you!  
{: .callout}
~~~

with

~~~
<div class="callout">
## Preparation Magic

If you type the command:
`PS1='$ '`
into your shell, followed by pressing the 'enter' key,
your window should look like our example in this lesson.  
This isn't necessary to follow along (in fact, your prompt may have
other helpful information you want to know about).  This is up to you!  
</div>
~~~

since the lack of semantic bugs me a lot.

Suggestions of how we can generate the cards are very welcomed.

gcapes:
Just use commands with comments in code blocks, and have a script extract just these code blocks from the lesson. This would require some editing of the existing material.

rgaiacs:
@gcapes could you provide a small example of what you mean by

> Just use commands with comments in code blocks, and have a script extract just these code blocks from the lesson.

shwina:
@gcapes @rgaiacs and @ztrautt 

Could this discussion be moved to a new issue in the styles repo? I think it is moving more towards a discussion on markdown rather than the shell lesson.

ztrautt:
@shwina I don't mind

gcapes:
@shwina Quite happy for someone to move this discussion off the shell lesson repo.

@rgaiacs I don't have a script in mind to do the extraction, but the idea is just what I said before:

> In one of my own lessons, I make sure that everything that is needed to teach the live coding is in a code block, with comments if appropriate. That way I don't have to read the text of the lesson --- I can just scroll to the next code block. 

See e.g. http://gcapes.github.io/git-course/02-local/

This is adequate for my needs without needing to extract the code blocks from the lesson. I was just suggesting that my lesson would lend itself to extracting the code blocks due to my comments, whereas the Software Carpentry lessons generally wouldn't, because they don't contain comments.

shwina:
Opened an issue https://github.com/swcarpentry/styles/issues/190 for discussion about this. Closing for now.

PR #657, opened 2017-11-14
be more explicit about the "cd" step

stragu:
I think that third step is not clear enough in what you have to do, and why you have to do it. Spelling it out will make it clearer to people who are new to the shell.


gdevenyi:
Very nice.

PR #658, opened 2017-11-14
inaccuracy: argument -> flag

stragu:
According to the material, this should read "flag", not "argument".


gdevenyi:
Good catch, thanks!

PR #659, opened 2017-11-14
remove material already covered in previous lesson

stragu:
An objective and an exercise associated to getting acquainted with the `ls` command are unnecessary repeated in this lesson. Those are already covered in the previous section "Navigating Files and Directories".


gdevenyi:
Good eye on that, that's definitely a duplicate.

gdevenyi:
I *love* patches that make the lessons shorter rather than longer :+1:

PR #660, opened 2017-11-14
add keypoint about nano vs touch

stragu:
`nano` and `touch` are the only two new command introduced in this section that don't have a description in the keypoints.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -15,6 +15,7 @@ keypoints:
 - "`mkdir path` creates a new directory."
 - "`mv old new` moves (renames) a file or directory."
 - "`rm path` removes (deletes) a file."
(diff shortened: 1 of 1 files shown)


shwina:
I am not sure whether it is a good idea to draw any connections between `nano` (a text editor), and `touch` (a command ordinarily used to update timestamps on *any* kind of file - not only text files).

gdevenyi:
I have to agree with @shwina that this draws an unnecessary connection between `nano` and `touch` which isn't intended in the lesson.

Thanks for your contribution!

stragu:
Fair point, I did not know that `touch` was mainly about updating timestamps.
Should we still add separate keypoints for those two commands though? I assume keypoints should _at least_ include the new commands introduced in the lesson.

PR #662, opened 2017-11-27
move exercise - moving to the current folder

ErinBecker:
Moves "moving to the current folder" to the episode text above the discussion of `cp` and immediately after discussion of `mv` and `.`.


shwina:
Thanks, @ErinBecker!

PR #673, opened 2017-12-05
Added a brief explanation of "Unix Directory Structure" which helps in navigating files and directories in a Unix environment.

yosephabate:


shwina:
Thanks! I do think this is an important topic, but:

1. This addition to the lesson raises more questions than answers, e.g., "What is the /usr or /srv folder? Can I put my files there?"

2. (especially if we choose to answer questions like the above) What should be removed from the lesson in order to make space for this new topic?

@gdevenyi ?

gdevenyi:
I agree with @shwina that this opens too many questions for learners. Furthermore, this directory structure is different depending on the underlying shell system, as a user could have linux/unix/bsd/osx/windows.

gdevenyi:
Thanks for the submission, but it is outside the scope of these lessions!

jttkim:
I have sympathy both for the objective of not overloading the lesson and that of giving learners a glimpse / lead into how empowering understanding and exploring the file hierarchy really is. So I think that while the level of detail is too high (especially in `directorystructure.png`), the idea of giving some lead into the existence and explorability of the directory hierarchy above the login directory makes a lot of sense to me.

Would it make sense to simplify the diagram to `/etc`, `bin`, `/usr`, `/usr/bin` and perhaps `/tmp`, and render the discussion in a breakout box? Or perhaps refashion it into an optional exercise to explore the tree above the login directory, perhaps with an element of getting learners to discuss with their neighbours what they find?

gdevenyi:
Perhaps shell-extras may want this. https://github.com/swcarpentry/shell-extras

yosephabate:
I would like to thank everyone for their comments. I learned how to keep the context of the course content within a boundary relevant to the prospective audience.

PR #674, opened 2017-12-06
Update loop lession so that the first example follows lesson

krisp:
During my instructor demo of this lesson, we noticed that the material at the beginning of the lesson does not correspond to the early examples. I've added an example that goes with the top of the lesson.

### DIFF
--- a/_episodes/05-loop.md
+++ b/_episodes/05-loop.md
@@ -59,7 +59,40 @@ error.
 
 Instead, we can use a **loop**
 to do some operation once for each thing in a list.
-Here's a simple example that displays the first three lines of each file in turn:
+
+Using our original input, let's craft a **`for` loop** to acomplish the task of backing up
+our files.
+
+~~~
+$ for filename in *.dat
+> do
+>     cp $filename original-$filename
+> done
+~~~
+{: .bash}
+
+When the shell sees the keyword `for`,
+it knows to repeat a command (or group of commands) once for each item in a list.
+Each time the loop runs (called an iteration), an item in the list is assigned in sequence to
+the **variable**, and the commands inside the loop are executed, before moving on to 
+the next item in the list.
+Inside the loop,
+we call for the variable's value by putting `$` in front of it.
+The `$` tells the shell interpreter to treat
+the **variable** as a variable name and substitute its value in its place,
+rather than treat it as text or an external command. 
+
+If the previous example executes correctly, no output will be displayed, because the `cp` command 
+that this loop runs does not produce any output. To see the results, we use the `ls` command to check 
+if the files were backed up as we expected.
+
+~~~
+$ ls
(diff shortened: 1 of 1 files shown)


cdarabos:
This is a great improvement on the existing example. I'd add a definition of the **variable**. The word variable appears first at line 77, and I suspect people might not immediately see that we refer to "filename".
Maybe something along these lines could replace lines 76-78:
===
Each time the loop runs is called an iteration. At the first iteration, the first item in the list is assigned  **variable** (called "filename" in the example), and the commands inside the loop are executed. An iteration is over once the last command in the loop is executed. Before the next iteration starts, the next item in the list, if any, is assigned to the **variable**. If we have executed the last command in the last iteration (i.e. the variable contained the last value in the list), the loop is exited and the code below the loop is executed.

colinmorris:
Sorry that this PR seems to have fallen through the cracks for a while. (Thanks for bumping, @cdarabos )

A surface problem with this change is that it renders a later part of the episode ("Going back to our original file copying problem, we can solve it using this loop:...") redundant. 

But more fundamentally, I don't think I like throwing the solution to the backup problem at them right away. I think there are some execution problems with this part of the episode as it currently exists, but I like the overall approach it takes of starting with a relatively simple for loop, and building up to the final goal (adding pieces as we go about concatenating with variables and using wildcards in the loop).

sthiede:
Hi! I also noticed that the example did not follow the problem that was set up during my instructor demo lesson as well - it was awkward to teach. In teaching for loops in R to a class recently, I did find it helpful to show a task that would take a long time without loops and then immediately show how to do it in a loop - to give the learners a  motivation as to why you would want to use a loop. 

Here's my suggestion: 

1)  Show a task that would be tedious to do without a for loop
2)  Show a simple for loop anatomy in an image or text (like this):
for name in list_of_names;
do an action;
done
3) Use for loop to solve simple tedious task (as in 1)
Note: The current simple for loop example could cause some confusion because you could also use wildcards to do the same thing - decreasing the motivation to use a for loop: head -n 3 *.dat

4) As before --  Set up the complex example - we want to copy many files in a directory and save them with a new name original-name.dat. 
5) As before -- show the error that you get if you try to use wildcards (Side note, I get a different error than the one in the current lesson) 

6) Clarify in words that "During this lesson, we will combine  learn  how to solve this with a for loop"

Thoughts?

gcapes:
Thanks for the PR @krisp and the summary @sthiede.

I like the summary suggestion, and particularly note point 3 - it should be well motivated.  The current example `head -n 3 *.dat` gives very nearly the same as the loop.

Perhaps a better loop to use as the example would be to output the classification for all files:
```
$ for filename in basilisk.dat unicorn.dat
> do
>    head -n 2 $filename | tail -1	# Indentation within the loop aids legibility
> done
CLASSIFICATION: basiliscus vulgaris
CLASSIFICATION: equus monoceros
```

Issue #501, opened 2017-01-25
Time allocated for exercises is 0


jarrah42:
In all the episodes in this lesson, the time allocated for exercises is 0, even though some of the episodes have challenges. Is this by design, or for some other reason?

marwahaha:
Yea, I'm in agreement here. I feel like we should bump out the time of this lesson to around 2.5 or maybe 3 hours, to account for the challenges.

gdevenyi:
I think we will need some feedback from people who have taught the lesson recently to update the timings. Thanks!

marwahaha:
I've taught this lesson 4-5 times this calendar year, and each time, it takes nearly 3 hours.

gdevenyi:
Thanks @marwahaha that's helpful information, but we need to specify exercise time use on each episode and the final time is calculated from those.

Do you have any feedback on the exercise time use for each episode?

ChristinaLK:
I'd like to bump this again -- the timing for this lesson is still off (as far as I know) -- all the lesson pieces should take 3+ hours.

ErinBecker:
@gdevenyi - I've also taught this lesson this year and the timings are significant underestimates, as no time for exercises are included. In instructor training, we recommend half a day for this lesson and half a day for git, meaning each is about 3-3.5 hours (after allowing for coffee breaks and lunch). I recommend that we add time estimates to each episode that split this time addition equally among them (except for the first episode which has no exercises). Even though this will be inaccurate, it will be a vast improvement to the current time estimates. I will put in a PR.

PR #677, opened 2017-12-20
add time estimates

ErinBecker:
Re: https://github.com/swcarpentry/shell-novice/issues/501

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -1,7 +1,7 @@
 ---
 title: "Navigating Files and Directories"
 teaching: 15
-exercises: 0
+exercises: 20
 questions:
 - "How can I move around on my computer?"
 - "How can I see what files and directories I have?"
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -1,7 +1,7 @@
 ---
 title: "Working With Files and Directories"
 teaching: 15
-exercises: 0
+exercises: 20
 questions:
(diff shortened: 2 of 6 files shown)


gcapes:
I noted how long it took me to teach this lesson the last couple of times I taught it. Approx timings below:

1. 10
2. 30 + 10
3. 25 + 10
4. 30 + 20
5. 40 + 10
6. 30 + 15
7. 25 + 20

Which is 4:15 total excluding breaks. So my view of the current timings is that most of the episodes take longer than allotted, but some of the exercises take less time.

gdevenyi:
Thanks @gcapes do you suggest any particular changes to this PR?

gcapes:
@gdevenyi 
I would suggest changing the teaching + exercises timings to those above (unless someone strongly disagrees?). I appreciate that different groups/instructors will go at different paces, but my timings are based on a few deliveries of the lesson, and from experience the current 15 mins given for teaching each episode is way off.

ChristinaLK:
I would be okay with using @gcapes times.  In that case, @ErinBecker could either update this PR or @gcapes could submit his own. @gcapes, if you're willing, go ahead.

ErinBecker:
@gcapes thanks for taking these timings. I'd be happy to have my suggested timings updated to these. Can you put in a new PR and we'll close this one?

gcapes:
Yep, no problem. Standby...!

gdevenyi:
Closed in favour of #680

PR #681, opened 2018-01-08
Remove references to goodiff

### LINKED ISSUE #655: goodiff is never used
Two scripts are introduced in episode 1 of the lesson, but only one is ever used.

Perhaps `goodiff` should be dropped from the story if it isn't used (my preference because the lesson is long enough as it is), or included in the processing script in ep 6. It's not quite clear if `goodiff` needs to be run for all combinations of processed samples, but whatever the intention, it would add extra complication.

Thoughts?

gcapes:
Reworded initial scenario accordingly.

Fix #655

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -113,10 +113,7 @@ She has 1520 samples in all and now needs to:
     a file with one line for each protein.
 2.  Calculate statistics for each of the proteins separately
     using a program her supervisor wrote called `goostats`.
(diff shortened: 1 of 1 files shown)


gcapes:
I've left one reference to `goodiff` in episode 6 (redo-figure-3.sh) because this history snipped also contains `ygraph` as an unexplained command and the details of this history snippet aren't important for the learning objective.

gdevenyi:
Looks good thanks!

PR #685, opened 2018-01-22
Clarification of flags on first use of cut command

tmellan:
cut -d , -f 2 could perhaps be a little scary for someone who has never used the command before. Therefore I added a note on the meaning of "d" and "f" on the first occurance in the lesson to make the flags a little less cryptic to novices. Cheers!


gdevenyi:
Thanks for the great addition!

PR #688, opened 2018-01-23
enable syntax highlighting for code snippets

### LINKED ISSUE #687: No syntax highlighting
Currently all code snippets seem to be annotated with `{: .bash}`.

But according to [this](http://swcarpentry.github.io/lesson-example/04-formatting/index.html) the class name to use to get syntax highlighting is actually `.language-bash`.

colinmorris:
I think the class that was being used (`.bash`) is out of date.

Fixes #687

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -54,7 +54,7 @@ let's open a shell window:
 ~~~
 $
 ~~~
-{: .bash}
+{: .language-bash}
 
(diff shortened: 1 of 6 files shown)


rgaiacs:
Looks good to me. For details see https://github.com/swcarpentry/styles/releases/tag/v9.3.0.

gdevenyi:
Good catch, didn't see that in the styles change.

PR #692, opened 2018-01-25
Update setup.md

cryptic0:
Fixed a typo.


gcapes:
This has been fixed in de0d7d1e1d76c3562b5

colinmorris:
Ah, thanks for pointing that out @gcapes. In that case, I'll close this.

Thanks anyways, @cryptic0 . We should have merged this when it came in - sorry that it fell through the cracks.

PR #693, opened 2018-01-25
Update 02-filedir.md

cryptic0:
Added wording to clarify command descriptions better.

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -193,12 +193,12 @@ because its name begins with `/`.
 
 Underneath `/Users`,
 we find one directory for each user with an account on Nelle's machine,
-her colleagues the Mummy and Wolfman.  
+her colleagues the `imhotep` and `larry`.  
 
 ![Home Directories](../fig/home-directories.svg)
 
-The Mummy's files are stored in `/Users/imhotep`,
-Wolfman's in `/Users/larry`,
+The user imhotep's files are stored in `/Users/imhotep`,
+user larry's in `/Users/larry`,
 and Nelle's in `/Users/nelle`.  Because Nelle is the user in our
 examples here, this is why we get `/Users/nelle` as our home directory.  
 Typically, when you open a new command prompt you will be in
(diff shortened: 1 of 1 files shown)


_episodes/02-filedir.md
```diff
@@ -193,12 +193,12 @@ because its name begins with `/`.
 
 Underneath `/Users`,
 we find one directory for each user with an account on Nelle's machine,
-her colleagues the Mummy and Wolfman.  
+her colleagues the `imhotep` and `larry`.
```
shwina: We don't need the *the*, anymore. So just:

> her colleagues `imhotep` and `larry`.

_episodes/02-filedir.md
```diff
@@ -193,12 +193,12 @@ because its name begins with `/`.
...
 Underneath `/Users`,
 we find one directory for each user with an account on Nelle's machine,
-her colleagues the Mummy and Wolfman.  
+her colleagues the `imhotep` and `larry`.  
 
 ![Home Directories](../fig/home-directories.svg)
 
-The Mummy's files are stored in `/Users/imhotep`,
-Wolfman's in `/Users/larry`,
+The user imhotep's files are stored in `/Users/imhotep`,
```
shwina: Previously, we used monospaced text when refering to `imhotep` and `larry`; do you think we should do that consistently throughout?

_episodes/02-filedir.md
```diff
@@ -442,7 +442,7 @@ To navigate through the `man` pages,
 you may use the up and down arrow keys to move line-by-line,
 or try the "b" and spacebar keys to skip up and down by a full page.
 To search for a character or word in the `man` pages, 
-use "/" followed by the character or word you are searching for. 
+use "/" followed by the character or word you are searching for. At times, a search will result in multiple hits.  If so, you can move between hits using `n` key (for moving forward) and `N` key (for moving backward).
```
shwina: 1. The surrounding text uses double quotes when referring to a key, e.g., "b", not `b`. What convention to use?
2. It may be confusing to refer to both an "n" key and also an "N" key (there is only one N key on the keyboard). Perhaps we should avoid use of the word "key" throughout. How does this sound?

> At times, a search will result in multiple hits.  If so, you can move between hits using `n` (for moving forward) and `N` (for moving backward

_episodes/02-filedir.md
```diff
@@ -540,8 +540,7 @@ $ cd data
 {: .bash}
 
 These commands will move us from our home directory onto our Desktop, then into
-the `data-shell` directory, then into the `data` directory.  `cd` doesn't print anything,
-but if we run `pwd` after it, we can see that we are now
+the `data-shell` directory, then into the `data` directory.  You will notice that `cd` doesn't print anything.  This is normal.  Many linux commands will not output anything to the screen when successfully executed.  But if we run `pwd` after it, we can see that we are now
```
shwina: Let's avoid "linux commands" in favour of "shell commands".

gcapes:
Fixed in 20a156abc6deeb12cc31d1fdd0924293f4a96254

PR #696, opened 2018-01-30
Shell lesson, No.7 [Last exercise] - Better to not hard-code the username

amit491:
The last exercise (shell-novice, lesson no.7): "**Finding files with different properties**"
It is asking to find files modified by user 'ahmed' in last 24 hrs.

But not necessarily the workshop attendee would have that user name in his/her PC/Mac. And also, when I downloaded and unzipped the **shell-novice-data.zip**, none of the files in the uncompressed folder are of user 'ahmed'.

Hence it would be better to catch the username on the fly using the command that has already been introduced in lesson no.2.

Hence instead of this as solution -
`find ./ -type f -mtime -1 -user ahmed`

I suggest this -
`find ./ -type f -mtime -1 -user $(whoami)`

### DIFF
--- a/_episodes/07-find.md
+++ b/_episodes/07-find.md
@@ -18,8 +18,8 @@ keypoints:
 - "`$(command)` inserts a command's output in place."
 ---
 
-In the same way that many of us now use "Google" as a 
-verb meaning "to find", Unix programmers often use the 
+In the same way that many of us now use "Google" as a
+verb meaning "to find", Unix programmers often use the
 word "grep".
 "grep" is a contraction of "global/regular expression/print",
 a common sequence of operations in early Unix text editors.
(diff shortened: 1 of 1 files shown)


gdevenyi:
This looks okay, but I'd like the command broken out so the user checks their ownership first on a separate line.

Also, there's a bunch of whitespace changes, can you clobber those and rebase?

amit491:
Dear Gabriel,

Yes, I noticed the whitespace changes but only after I had submitted the PR.
I had used Atom text editor on Mac to make changes to the .md file. Probably the editor did something to the line end characters.
Ok I will try to sort this out.

Thanks,

gcapes:
Thanks for the contribution!

I think it would be great if the exercise was runable rather than just a thought experiment. Which directory should the learner be in for this exercise?

I'd also like to add that `whoami` is not covered anywhere in this course.

deppen8:
The "Finding files with different properties" exercise is no longer in the lesson. Closing this as part of old PR cleanup.

PR #705, opened 2018-02-06
add note about merging -l -h into -lh.

### LINKED ISSUE #704: grouping arguments together in `ls -lh`
in the section "parameters vs. arguments" in http://swcarpentry.github.io/shell-novice/02-filedir/, `ls -lh` is suggested and it is noted that `-lh` are the flags.  This is the first case where the flags are grouped together with a single `-` (previously, it was suggested to do `ls -R -t`).  

It might be worth a sentence to say that it is common to group single-letter flags together like this.

zingale:
closes #704

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -469,7 +469,9 @@ which doesn't exist.
 > ```
 > {: .language-bash}
 > `ls` is the command, `-lh` are the flags (also called options),
-> and `Documents` is the argument.
+> and `Documents` is the argument.  Notice in this example, we combined two
+separate flags, `-l -h` into a single flag, `-lh`.  This grouping is common
+with single-character flags.
 {: .callout}
 
 We can also use `ls` to see the contents of a different directory.  Let's take a


_episodes/02-filedir.md
```diff
@@ -469,7 +469,9 @@ which doesn't exist.
 > ```
 > {: .language-bash}
 > `ls` is the command, `-lh` are the flags (also called options),
-> and `Documents` is the argument.
+> and `Documents` is the argument.  Notice in this example, we combined two
+separate flags, `-l -h` into a single flag, `-lh`.  This grouping is common
```
gcapes: `-lh`  isn't a single flag, it's still two flags, just grouped together.

gcapes:
It's a good idea to explain this, but I think the wording is confusing (see comment on the code).

_episodes/02-filedir.md
colinmorris: I think the meaning is clear, but you're right, the wording is imprecise. If you have a suggestion for how to reword this, a PR would be welcome.

gcapes:
Sure - will do (tomorrow).

PR #710, opened 2018-02-13
mention backup of Salon's haikus

katrinleinweber:
Hi!

The `five years` quote seems a bit pessimistic ;-) This PR suggests to link to the C2 wiki, a generally useful resource for programmers.

Cheers!


katrinleinweber:
They will [use redirects](https://en.wikipedia.org/wiki/URL_redirection) ;-)

Issue #712, opened 2018-02-19
Unix shell Loops 'Spaces in Names' section


marisalim:
In the example for loops with file names that have spaces, the result without quotes around $filename, will actually return a result for unicorn.dat since that is one of the files in the SWC creatures/ directory. The output should look like this:

for filename in "red dragon.dat" "purple unicorn.dat"; do head -n 5 $filename; done
head: cannot open ‘red’ for reading: No such file or directory
head: cannot open ‘dragon.dat’ for reading: No such file or directory
head: cannot open ‘purple’ for reading: No such file or directory
==> unicorn.dat <==
COMMON NAME: unicorn
CLASSIFICATION: equus monoceros
UPDATED: 1738-11-24
AGCCGGGTCG
CTTTACCTTA

jttkim:
It appears to me that the scenario "Suppose our data files are named: `red dragon.dat`, `purple unicorn.dat`" was written somewhat independently of the lesson, so the assumption is that the working directory contains only those two files. But if a learner actually tries the loop in the directory of the lesson, they'd see the result you describe -- so good catch.

Perhaps a solution is to rewrite this part as something like "Suppose you have a directory with the data files `amber anteater.dat` and `brown bear.dat`", i.e. (1) so that it's clear that this part does not pertain to the directory of the lesson and (2) the files are unrelated to those used in the lesson. What do people think?

marisalim:
That sounds good to me!

gcapes:
Why not just explain the actual result? That doesn't take away from the learning objective -- spaces separate arguments.

marisalim:
That would work too. In that case, the error for unicorn.dat just needs to be replaced with an actual output from head -n # for unicorn.dat.

gcapes:
Yep - exactly. Are you planning to submit a PR for this @marisalim ?

marisalim:
Yes, I will submit a PR!

colinmorris:
Fixed by #713

PR #713, opened 2018-02-21
Edit to loops episode - spaces in names

marisalim:
I made some edits to the Loops episode based on this issue: Unix shell Loops 'Spaces in Names' section #712. Assuming the learner tests the 'red dragon.dat' and 'purple unicorn.dat' loop in the creatures directory, they will get an output from unicorn.dat. I've edited the output and added a sentence to explain why there is a result.

### DIFF
--- a/_episodes/05-loop.md
+++ b/_episodes/05-loop.md
@@ -362,7 +362,7 @@ from whatever file is being processed
 > ```
 > {: .output}
 > Try removing the quotes around `$filename` in the loop above to see the effect of the quote
-> marks on whitespace:
+> marks on whitespace. Note that we get a result from the loop command for unicorn.dat if you ran this code in the creatures directory:
 > ```
 > head: cannot open ‘red’ for reading: No such file or directory
 > head: cannot open ‘dragon.dat’ for reading: No such file or directory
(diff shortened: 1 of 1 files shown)


_episodes/05-loop.md
```diff
@@ -362,12 +362,31 @@ from whatever file is being processed
 > ```
 > {: .output}
 > Try removing the quotes around `$filename` in the loop above to see the effect of the quote
-> marks on whitespace:
+> marks on whitespace. Note that we get a result from the loop command for unicorn.dat if you ran this code in the creatures directory:
```
colinmorris: Grammar nitpick: how about "when we run this code" instead of "if you ran this code" (to be consistent with the beginning of the sentence).

colinmorris: Could you put \`backticks\` around creatures?

_episodes/05-loop.md
```diff
@@ -362,12 +362,31 @@ from whatever file is being processed
...
+> CTTCACTGTA
+> GAGGTGTACA
+> AGGATCCGTT
+> GCGCGGGCGG
+> CAGTCTATGT
+> TTTTCGACAC
+> TGGACTGCTT
+> CCCTTTGAGG
+> GTGGATTTTT
+> CGTAACGGGT
```
colinmorris: Unfortunately this takes up a lot of vertical space, just due to the command used in the example (and this callout box was already pretty big to begin with).

What do you think about changing the tail command to something like `tail -n 3`? (You might also want to change the code outside the callout box to match, but I don't think it's totally necessary.)

marisalim: Thanks for the suggestions - I made all of these changes!

colinmorris:
Looks good. Thanks!

PR #714, opened 2018-02-22
clarify: more verbose, reordered exercises

stragu:
It makes sense to group the three exercises that use the `animals.txt` file.

Intro to `sort -n` exercise more verbose to make it clear what the lines are.


colinmorris:
This is great! Thank you.

PR #715, opened 2018-02-22
Consistency with lesson #2

profgiuseppe:
The  command should reflect the directory structure. In Lesson #2, the data-shell/ directory is placed inside Desktop/ 
Even if it was placed inside the home directory, at least data-shell/ should be in the path.


colinmorris:
Good catch! Thanks

PR #718, opened 2018-02-26
Changed order of analyzed and raw in the exercise

davebridges:
The output of `ls -F` should be in alphabetical order

This was pointed out by a learner during a workshop


colinmorris:
Thanks!

PR #722, opened 2018-03-03
ISSUE-600: data-shell/data/animals.txt is truncated

agrimstrup:
Added additional records to bring the animals.txt file up to 586 lines as mentioned in the text.  Added ellipses to the file listing to indicate more data exists in the file than listed.

Before the change, checked the output of the example
[arne@localhost data]$ cat animals.txt | head -n 5 | tail -n 3 | sort -r
2012-11-06,rabbit
2012-11-06,deer
2012-11-05,raccoon

Added additional records and checked the distribution of values
[arne@localhost data]$ wc -l animals.txt 
586 animals.txt

[arne@localhost data]$ cat animals.txt | sort | uniq -c
     43 2012-11-05,bear
     35 2012-11-05,deer
     44 2012-11-05,fox
     44 2012-11-05,rabbit
     45 2012-11-05,raccoon
     42 2012-11-06,bear
     37 2012-11-06,deer
     44 2012-11-06,fox
     39 2012-11-06,rabbit
     35 2012-11-06,raccoon
     36 2012-11-07,bear
     32 2012-11-07,deer
     40 2012-11-07,fox
     36 2012-11-07,rabbit
     34 2012-11-07,raccoon

Tested the example pipeline to ensure no changes resulted 
[arne@localhost data]$ cat animals.txt | head -n 5 | tail -n 3 | sort -r
2012-11-06,rabbit
2012-11-06,deer
2012-11-05,raccoon

---


gcapes:
If this file is to be updated, the zip file for learners to download also needs updating.

colinmorris:
This looks good to me, though I'd like to get a +1 from another maintainer before merging this in case there's something I'm missing. @shwina , any thoughts?

agrimstrup:
@gcapes I didn't see the zip file in the first pass.  An updated version of the zip file has been added to the PR.

gdevenyi:
I agree that there's no need for a "full" dataset. The truncation can serve fine as an example of a subset.

colinmorris:
I'm personally fine with either fix (increasing the file size to match the existing lesson text, or changing the lesson text to match the existing file size). However the former has the advantage that we have a pull request that implements it (right here), and the latter is hypothetical. 

Is there a downside to merging this? It seems like a strict improvement over the current state of the lesson.

shwina:
+1 from me

PR #690, opened 2018-01-25
Separate shell grammar from navigation in Episode 2

sleak-lbl:
(This contribution is for instructor training checkout)

Shell grammar - and the model of the CLI being like a simple language for communicating with the computer - is a distinct concept from navigation, but ad-hoc details about shell grammar are intertwined with navigation in Episode 2.

I think a clearer flow for the lesson is to introduce what a shell command looks like first, in Episode 1, and reserve episode 2 for things related to navigation. So I've pulled those parts of Episode 2 back into Episode 1, with a few wording tweaks along the way

I think this PR addresses issues #641, #590 and #543 / #447  too

PR #723, opened 2018-03-03
Updating `ls --help` in 02-filedir.md

rrlove:
Added discussion of `man ls` as alternative to `ls --help`, because neither works across all operating systems.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -20,7 +20,7 @@ kind: "lesson"
 
 # Magic to make URLs resolve both locally and on GitHub.
 # See https://help.github.com/articles/repository-metadata-on-github-pages/.
-repository: <USERNAME>/<PROJECT>
+repository: rrlove/shell-novice
 
 # Sites.
 amy_site: "https://amy.software-carpentry.org/workshops"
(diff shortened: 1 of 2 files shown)


colinmorris:
Hi @rrlove, thanks for the suggestion. 

There's already a note in this episode on `man ls`, and an infobox on `man` and Git for Windows. You may not have noticed them because they're separated from the `ls --help` example by a bunch of callout boxes.

Do you think you could try to harmonize your proposed additions with that existing content? (A good first step might be moving that existing `man` content up, so it's not separated from the `--help` part by a bunch of boxes.)

colinmorris:
Actually, I should mention that #690 also includes significant changes to the part of the lesson dealing with `--help` and `man`. Perhaps you could take a look at the relevant part of that PR and see if you think it addresses your concerns, or if you have any feedback on their approach.

rrlove:
Apologies for my delayed response; I was teaching the workshop for which the preparation prompted this PR. Yes, #690 addresses my concerns (as well as other issues I've noticed over the years). I look forward to using that version next time.

PR #725, opened 2018-03-05
Update zipfile with correct time stamps, and add makefile rule

### LINKED ISSUE #724: Data files in zip file (nearly) all have same time stamp -- exercise meaningless
Re: http://swcarpentry.github.io/shell-novice/02-filedir/

In the file learners download, most of the time stamps are the same i.e. not preserved from the data-shell directory. This means that the second exercise is fairly meaningless `Listing Recursively and By Time`.

I'll submit a PR with an updated zip file, which will fix the exercise.

I'll also add a new Makefile rule to rebuild the zip file if required. I don't think there is an ideal way to write the rule because the `data-shell` directory timestamp won't update when files within it are modified (so the rule would never run if it depended on `data-shell`, so I'll make it a `.PHONY` rule and it'll run every time.

gcapes:
Fix #724


colinmorris:
The makefile addition is awesome!

Could you explain the time stamp thing more though? If someone runs `make zip` again, will there still be modification time differences, or did you have to do some finagling on your local file system first?

gcapes:
Sure. Buckle up for a long read ;)

A makefile rule checks to see whether the dependencies (i.e. the data-shell directory) are newer than the target (i.e. the zip file). If the dependencies are newer (i.e. have been changed), the rule will update the target. Had I written the rule as a normal make rule, then the zip file would never be built, because the data-shell directory's timestamp doesn't update when the files or directories it contains are updated.

Instead I wrote the rule using a `.PHONY` target, so that it will always run. I couldn't think of a robust workaround, but I think it's good enough given how infrequently I would anticipate needing to use it.

The main reason for writing a rule is that someone looks to have used a zip utility which didn't respect the time stamps, causing most of the files to have (presumably) the zip-file-creation-time rather than file-creation time.

TL;DR
If you run `make zip` the zip file is overwritten and gets a new time stamp, but the files inside the zip have the time-stamps as they are in the `data-shell` directory. However, `make zip`will always update the zip file even if there have been no changes in the `data-shell` directory.

colinmorris:
I was wondering more about the `data-shell` timestamps themselves and on what basis they're set. I had to do some googling on this, but it seems like file modification time is a property of the local file system, and not a piece of metadata tracked by git itself. 

So if someone clones the `shell-novice` repo, all the files in the repo (including in `data-shell/`) will have the same creation/modification time (corresponding to when they did the clone). So it seems, depending on the state of their file system, someone rerunning `make zip` could undo your work, and send the `data-shell` files back to the state of all having the same mtime. 

If we really wanted to guard against this, we could add some code to the make rule to arbitrarily jitter the timestamps in the `data-shell` directory before zipping. (Apparently this can be done using [`touch -d`](https://askubuntu.com/a/62496).)

That said, I'm happy to merge this now (and maybe track the remaining work in a new issue?).

gcapes:
You're absolutely right! I hadn't really thought about that -- thanks for pointing it out.

I think some code to ensure the time stamps are different is something worth looking to  -- sorting by time isn't a very useful exercise if all the time stamps are the same. But maybe this is a bit more complicated than it needs to be.

I guess the options include:
1. Have something clever in a makefile rule to assign random time stamps to the files -- too complicated / one more thing to keep track of during PRs?
2. Change the example to sort by some other metric like file size -- still meets the learning objectives of combining flags.

I'm happy to look into whichever option we collectively agree is best.

gcapes-test:
I'm half way through teaching this right now. The listing `recursively and by time` exercise is in the user's home directory. So the contents of the zip file don't matter at this point. It depends on the user's  home directory whether the output is very meaningful. I think we should just leave it for now.

So the time-stamps issue in the zip file doesn't matter. This PR can be closed, but I'll split out the makefile commit into another PR because I think that's still useful.

PR #726, opened 2018-03-05
Relabel root directory in filesystem figs

### LINKED ISSUE #703: directory diagram in 02-filedir uses "/" and "root" confusingly in top level folder
In the directory hierarchy figure in http://swcarpentry.github.io/shell-novice/02-filedir/, the top directory is labeled "/" and underneath "root".  This is the only folder with two labels in that figure, and it can be a source of confusion.  On my linux system, 

`/`
and 
`/root`

are two different directories, one is the root (`/`) and the other is the home directory for the `root` user.  

If a student types `ls /root` as the figure might suggest, they may be trying to look into the `/root/` directory and get permission denied.

It would be good to make the figure consistent and have only a single label `/` in that figure for the root directory.

gcapes:
Fix #703.

After some searching for a good online editor, I used http://www.drawsvg.org (in case that's useful to anyone else).

### DIFF
--- a/fig/filesystem.svg
+++ b/fig/filesystem.svg
@@ -1,150 +1,40 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<svg
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
-   xmlns:cc="http://creativecommons.org/ns#"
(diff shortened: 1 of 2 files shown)


colinmorris:
Great!

Thanks for all the recent PRs. You've been killing it. :+1:

gcapes:
Thanks @colinmorris !

You've been very responsive in replying to them. Pats on the back all round :)

PR #727, opened 2018-03-05
Add callout to explain indentation within a loop

### LINKED ISSUE #586: explain indentation in for loop
`05-loop` has indentation within the for loop, but does not explain it (is not necessary, used for legibility).

gcapes:
Fix #586.


colinmorris:
Looks good - thanks!

PR #728, opened 2018-03-05
Confirm directory for exercises and examples

### LINKED ISSUE #695: Add path to working directory after the third exercise in "Loops"
Exercise two and three are based on the material of the directory data-shell/molecules. After these exercises, the lecture is continued in a different directory (data-shell/creatures). It might be helpful to add a note of the directory where the followings section of the lecture is going to be based on.

gcapes:
Fix #695.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -226,9 +226,11 @@ draft.txt
 > {: .solution}
 {: .challenge}
 
-Let's tidy up by running `rm draft.txt`:
+Returning to the `data-shell` directory,
+let's tidy up the `thesis` directory by removing the draft we created:
 
 ~~~
+$ cd thesis
 $ rm draft.txt
 ~~~
 {: .language-bash}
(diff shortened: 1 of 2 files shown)


_episodes/03-create.md
```diff
@@ -226,9 +226,10 @@ draft.txt
 > {: .solution}
 {: .challenge}
 
-Let's tidy up by running `rm draft.txt`:
+Let's tidy up the `thesis` directory by removing the draft we created:
 
 ~~~
+$ cd -		# Return to data-shell/thesis after exercise above
```
colinmorris: I'd suggest changing this to a textual note like in the other cases (e.g. "returning to the `data-shell` directory..."). 

Even with the comment (which is great), I think this could trip someone up if:

1. They didn't get to their home directory in one try while doing the exercise, or did some additional moving around before returning to the main lesson.
2. The instructor skipped the exercise (because of time pressure, or whatever other reason)

colinmorris:
Looks good. Thanks!

PR #732, opened 2018-03-08
Remove `thesis` dir from zip file

### LINKED ISSUE #730: `thesis` directory already exists in zip file
Episode 03 creates a directory `thesis`, but this already exists in the zip file.
I'll submit a PR for this.

gcapes:
Fix #730.

Sorry -- this one was my fault.
I accidentally added this empty dir in c62335c7.

I've just seen #737, so I've fixed that in this PR too.


shwina:
Thanks, @gcapes!

PR #733, opened 2018-03-08
Add Makefile rule to create zip file

gcapes:
Split out from #725.


colinmorris:
Thanks, this is great.

PR #734, opened 2018-03-09
Use terminology correctly (argument and flag vs parameters)

gcapes:
Clarify wording and fix terminology in the excellent PR #690


gdevenyi:
Looks good, thanks for the careful cleanup!

gcapes:
Great, thanks!

PR #739, opened 2018-03-10
Clarify example's intended use of mv

marwahaha:
This confuses some learners each time I teach it (including this morning), since `mv` is overloaded (moving and renaming), 

When I instruct `mv`, I recommend adding a trailing slash (as in the change) when you intend to move a file. Then, if a path is mistyped, `mv` will not try to rename a file to the incorrectly typed path. 

I think adding the slash here clarifies the intent of `mv`. 

(The trailing `/` also would show up if you were to "Tab-complete" `recombine`. )


shwina:
Thanks!

PR #746, opened 2018-03-22
Suggesting learner to shell-extras (under development)

bagustris:
Instead of saying won’t learn much if he/she is already familiar with `grep` and `find`, we should suggest him to explore or contributing to shell-extras.


gcapes:
This seems like a good idea to me!

colinmorris:
Agreed, I like this much better. Thanks!

gdevenyi:
Wonderful!

Issue #702, opened 2018-02-01
Add external references to reference file


gdevenyi:
Collolate URLs into this area.

deppen8:
Looks like #747 took care of this. Closing.

PR #747, opened 2018-03-23
External links added to reference

martharobinson:
Attempted to address issue #702 by collating all external links in the lesson and adding them to reference.md. I have grouped links roughly by theme.


gdevenyi:
Very nice! Thanks!

Issue #708, opened 2018-02-08
Audit lessons for POSIXness


gdevenyi:
OSX provides BSD-family commands, while gitbash and linux will have GNU commands. People might run into a pure "sh" once in a while.

We should limit ourselves to POSIX to prevent non-working options.

This will need some careful auditing sadly...

deppen8:
Closing this in favor of #927 , where there is a long thread of discussion re: POSIX compliance

PR #749, opened 2018-03-27
Removed 'ls is alphabetical by default'

### LINKED ISSUE #744: ls does not print names in alphabetical order on all platforms
In episode 2 (_episodes/02-filedir.m) it says "ls prints the names of the files and directories in the current directory in alphabetical order".  On Mac this is NOT the case - files and directories are in asciibetical order (A-Z before a-z).

For example:
``
$ mkdir A b C

$ touch d E f

$ ls -F
A/	C/	E	b/	d	f
``

On Linux the man pages does state `Sort entries alphabetically if none of -cftuvSUX nor --sort  is  speci‐fied.`.  So this is true but ONLY ON THAT PLATFORM.

On Mac the man pages does not state how the files will be ordered if no (sorting) arguments are given - so it cannot be presumed that asciibetical is the guaranteed order.

This is a BSD vs Linux core tools issue, I believe.

loz-hurst:
This is only the case with the GNU implementation - the BSD version of ls (which is the default on OSX, for example) does not list files alphabetically.  The POSIX standard explicitly says the output formatting for terminals and the ordering is implementation specific (non-terminal output is guaranteed to be one file per line).

This change therefore addresses both the misinformation that ls always lists alphabetically by default and improves POSIX compliance of the text (for which there is also an outstanding issue).

I have also changed the word 'terminal' to 'settings', where is says the output may or may not be coloured by default - again this it determined by more than just the terminal supporting and having colour enabled (most importantly by where 'ls' has be aliased to a default set of flags including the colour one - RedHat do this but Debian do not, for example) so to say it is just the terminal is misleading - I believe substituting this word addresses the inaccuracy without reducing understandability or increasing the complexity of the material.

Fixes #744.  References #708.

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -137,13 +137,12 @@ Desktop      Downloads    Movies       Pictures
 (Again, your results may be slightly different depending on your operating
 system and how you have customized your filesystem.)
 
-`ls` prints the names of the files and directories in the current directory in
-alphabetical order, arranged neatly into columns.
+`ls` prints the names of the files and directories in the current directory. 
 We can make its output more comprehensible by using the **flag** `-F`
 (also known as a **switch** or an **option**) ,
 which tells `ls` to add a marker to file and directory names to indicate what
 they are. A trailing `/` indicates that this is a directory. Depending on your
-terminal, it might also use colors to indicate whether each entry is a file or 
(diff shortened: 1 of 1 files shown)


colinmorris:
This is great. Clearer and a bit shorter to boot! Thanks

PR #755, opened 2018-04-04
Correct Windows path separators

### LINKED ISSUE #754: Windows path separator is '\' rather than '\\'
Ep 02 seems to have gone to town with the backslashes in the keypoints!

gcapes:
Fix #754.


shwina:
Great catch!

PR #769, opened 2018-04-12
Add solution to redirection challenge

### LINKED ISSUE #761: Improvements to >> exercise
From the notes:

> What Does >> Mean?
What is the difference between:
$ echo hello > testfile01.txt
and:
$ echo hello >> testfile02.txt
Hint: Try executing each command twice in a row and then examining the output files.

I see a few issues with this exercise:
1. The **echo** command is not defined.
2. There is not a solution

I would suggest to reconsider this exercise to something like:

What Does >> Mean?
We have seen the use of > but there is a similar symbol >> that does something different and their use might get confused. By using the echo command for printing strings, what is the difference between them? 
$ echo hello > testfile01.txt
and:
$ echo hello >> testfile02.txt
Hint: Try executing each command twice in a row and then examining the output files by opening the nano editor.
Solution: In the first option with >, the output "hello" is written to testfile01.txt but the file gets overwritten the second time we run the same command. However, the second option writes "hello" to testfile02.txt and appends the output to the same file when we run it for the second time.

gcapes:
Fix #761.

### DIFF
--- a/_episodes/04-pipefilter.md
+++ b/_episodes/04-pipefilter.md
@@ -358,8 +358,9 @@ the output of `head` must be the file with the fewest lines.
 {: .callout}
 
 > ## What Does `>>` Mean?
->
-> What is the difference between:
+> We have seen the use of `>`, but there is a similar operator `>>` which works slightly differently.
+> By using the `echo` command to print strings, test the commands below to reveal the difference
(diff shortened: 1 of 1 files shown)


ErinBecker:
LGTM!

PR #776, opened 2018-05-10
Extended the example of errors in use of ls 

richardbeare:
to illustrate the difference in form of output when the errors come from bash
(caused by a typo) compared to when the errors come from a program (caused
by incorrect arguments).

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -333,6 +333,19 @@ or available locally via: info '(coreutils) ls invocation'
 > Try 'ls --help' for more information.
 > ~~~
 > {: .error}
+> Note the differences between this error and the type we saw earlier:
+>
+> ~~~
+> $ ls-j
+> ~~~
+> {: .language-bash}
+> 
+> ~~~
+> -bash: ls-j: command not found
+> ~~~
+> {: .error}
+> The errors have two different sources - the ls program and bash,
+> indicated at the start of the error message.
 {: .callout}
 
 #### The `man` command


shwina:
Thanks for the contribution. This is certainly interesting, but could it be explained further? Why are there two sources of errors, what does it really mean if there are two sources of errors, and how can it help the user debug?

richardbeare:
I'm unsure of the process here - I can add extra explanation but was worried about adding too much. I suspected that the best way was to leave the instructor free to comment. Should I add amend the commit with some more detail?

shwina:
@richardbeare 

Thanks! Sorry for not being more clear! My above comment was just a few questions for you about this contribution. Before you add any more commits, let's just discuss what could possibly be added/changed to this contribution.

richardbeare:
My general thought was to highlight how important it can be to read error messages carefully - something that beginners often need some guidance in doing effectively. In this pair of examples the difference to highlight is the 
``` bash
-bash
```
prefix to the message, indicating that error is being issued by bash, where as
``` bash 
ls: illegal option -- 
```
Indicates that the error is being issued by the ls program. In the first case the error means that bash can't find something, and thus an incorrect name has been typed (in this case because we missed the space). In the second case bash has correctly found the program, executed it, and the program has decided that something is wrong with the input arguments. Thus we'd be looking further along the command line for the source of error.

colinmorris:
Agreed that reading error messages carefully is an important practice to drive home. I'm not sure if this change effectively accomplishes that though. (Maybe a more effective approach would be a debugging exercise, where we show the reader a few commands and their corresponding error messages and ask them to fix them?)

I'm inclined to err on the side of brevity and not add this note. Thank you for the suggestion though.

PR #778, opened 2018-05-10
Remove markup not necessary.

rgaiacs:


gcapes:
Some markup here would be useful, even if it's bold or something. I've asked in the past about the backticks markup for in-text commands, but it is also widely used in the lessons to indicate keywords and terminology.

rgaiacs:
> Some markup here would be useful, even if it's bold or something.

I'm OK with bold or italic

> I've asked in the past about the backticks markup for in-text commands, but it is also widely used in the lessons to indicate keywords and terminology.

Sorry if I missed the discussion. :-( I'm trying to document some of this practices at http://swcarpentry.github.io/lesson-example/06-style-guide/index.html. My personal rule is that if is something that the user should type (e.g. name of one command) or had type we will use backticks. Othewise we will not use backticks. For example,

> The default shell for Linux operating systems is usually Bash

doesn't have the backticks because "Bash" refers to the name of the program when

> If your machine is set up to use something other than bash, you can run it by opening a terminal and typing `bash`.

does have the backticks because "bash" refers to what the user will type.

gdevenyi:
I agree that backtics should be reserved for runnable commands. Other emphasis should be used for non-commands.

shwina:
@rgaiacs I like this rule and it would be great if we could enforce it throughout the lesson. Also, I think it's acceptable to use backticks (and only backticks) for file/directory names or paths. Would you agree?

gcapes:
That appears to be at odds with @gdevenyi  and @rgaiacs?
Will this markup policy apply to all lessons?

rgaiacs:
> I like this rule and it would be great if we could enforce it throughout the lesson.

Is a little hard to enforce but we can try. I will send a email to discuss.

PR #780, opened 2018-05-10
Clear instructions to open the terminal

rgaiacs:


gdevenyi:
Great, can you also add as a last resort "xterm"

Not everyone will end up with KDE or Gnome

rgaiacs:
> Great, can you also add as a last resort "xterm"

Done.

PR #797, opened 2018-06-12
Corrected inaccuracy about capitalisation

### LINKED ISSUE #796: Episode 1: "capitalization matters" is not true
> Also, capitalization matters: LS is different to ls.

This depends entirely on the filesystem being used, and whether that is case sensitive.  This is not true by default on Mac OS, for example (which actually is a Unix, Linux is not):

```
$ ls
data-shell
$ LS
data-shell
$ Ls
data-shell
$ lS
data-shell
$ which ls
/bin/ls
$ which LS
/bin/LS
$ ls -l /bin/ | grep -i ls
-rwxr-xr-x  1 root  wheel    38688 19 Jan 07:37 ls
```
(note: there is only 1 `ls` binary but it can be accessed via any capitalisation)

I intend to address this by adding the words "can" and "usually".

loz-hurst:
Fixes #796

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -94,7 +94,7 @@ command doesn't always require an argument or a flag.
 In the second line of the example above, our **command** is `ls`, with a **flag** `-F` and an
 **argument** `/`. Each part is separated by spaces: if you omit the space 
 between `ls` and `-F` the shell will look for a command called `ls-F`, which 
-doesn't exist. Also, capitalization matters: `LS` is different to `ls`. 
+doesn't exist. Also, capitalization can matter: `LS` is usually different to `ls`. 
 
 Next we see the output that our command produced. In this case it is a listing 
 of files and folders in a location called `/` - we'll cover what all these mean 


_episodes/01-intro.md
```diff
@@ -94,7 +94,7 @@ command doesn't always require an argument or a flag.
 In the second line of the example above, our **command** is `ls`, with a **flag** `-F` and an
 **argument** `/`. Each part is separated by spaces: if you omit the space 
 between `ls` and `-F` the shell will look for a command called `ls-F`, which 
-doesn't exist. Also, capitalization matters: `LS` is different to `ls`. 
+doesn't exist. Also, capitalization can matter: `LS` is usually different to `ls`.
```
shwina: Good point - I think we need further explanation here, as including the word *usually* may naturally make learners curious about when capitalization matters.

loz-hurst: I was trying to avoid overcomplicating the material but correcting the inaccuracy (which causes in-lesson questions like 'but LS isn't different to ls on my computer' from Mac users).

Would 'often' be better than 'usually', or does that suffer the same issue in your mind?

shwina: What do you think about something like this:

> Also, capitalization can matter: depending on your operating system or its configuration, `LS` may be different from `ls`.

loz-hurst: It's depends on the filesystem being used, not the operating system - would being a bit more general than 'operating system' be acceptable? e.g. 'Also, capitalization can matter: depending on your system and configuration, LS may be different from ls.'?

jttkim: I favour keeping the current wording, in the interest of simplicity, especially considering that this is at the earliest stages of the shell novice lesson where avoiding unnecessary complications has a very high priority. Generally, there's no need to point learners to every possible quirk / corner case / convenience. Getting them started with simplified, safe and interoperable rules and techniques is a reasonable teaching practice that's frequently applied throughout our lessons, as far as I'm aware.

I'd also argue that paying attention to capitalisation is generally a good idea, even if some shell implementations may somehow not require that. Leading learners to expect / depend on this kind of hidden auto-canonicalisation may well turn out to be a disservice to them when they move on to Python or R.

And finally, from a perhaps rather more advanced perspective I'd also argue that sticking with consistent capitalisation is a good "clean coding" habit.

gcapes: #796 is now closed.

PR #800, opened 2018-06-12
Removed wording that it's 'not difficult'

### LINKED ISSUE #799: Lesson 01 material violates the "don't tell attendees it's easy" rule imparted during instructor training
> Is it difficult?
> 
> It isn’t difficult,

How is an attendee struggling to get to grips with the material going to interpret this?  A big part of the instructor training I was given focused on not belittling attendees by telling them something is easy for this reason, and yet the material effectively says it.

I think that the first part of the paragraph should be dropped entirely, so instead of:
> It isn’t difficult, but it is a different model of interacting than a GUI, and that will take some effort - and some time - to learn.

it would read:

> It is a different model of interacting than a GUI, and that will take some effort - and some time - to learn.

(same message, without saying that it's not difficult and therefore implying you must be stupid if finding it hard)

loz-hurst:
'it's not difficult' implies you must be stupid if finding it hard.  This is demoralising and insulting to students using the material.

Fixes #799

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -150,7 +150,7 @@ and then print the prompt and wait for you to enter another command.
 
 ### Is it difficult?
 
-It isn't difficult, but it is a different model of interacting than a GUI, and that 
+It is a different model of interacting than a GUI, and that 
 will take some effort - and some time - to learn. A GUI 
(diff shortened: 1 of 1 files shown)


shwina:
+1 from me.

Issue #798, opened 2018-06-12
'Unix' misused throughout the material


loz-hurst:
In a way this is related to #779.

Bourne Again SHell (BASH) is the shell used for this course, which is entitled "The Unix Shell".  This seems fine to me as `sh`, which BASH derives from was called "Unix Shell".

However, the description of BASH as "The Unix Shell" has led to the conflation of the shell (program) and the operating system, which non-Unix OSs frequently described as Unix (esp. with regard to Linux, and there are legal reasons for not describing or implying Linux is, or derived from, a Unix) and the reference to Unix as the OS throughout the material when, most of the time, it means Unix-like.

> There are also options for identifying and downloading a Unix Shell program, a Linux/UNIX emulator, or a program to access a UNIX server.

> Directory names in a path are separated with / on Unix, but \ on Windows.

>  On Unix systems (such as Linux and Mac OS X)

There are numerous, and increasing, places scattered throughout the material where the differences between Unix (e.g. Mac OSX/BSD), Linux and Windows are noted as well as many gitlab issues related to the same.  I wonder if it would be better to simply note that there are differences (without going into too much detail) somewhere and then audit the rest of the material to remove all the "it's slightly different on platform x from what is here" notes? (Perhaps replacing examples that are very different with ones that are the same - e.g., I think the '-F' flag was, looking at the number of issues related to just that, a poor choice to introduce flags to ls in lesson 1 (see #590 #644 for example) and '-l', which is works on all 3 platforms and is not the default alias for 'ls' on any, would have been a better but 'less clever looking' introduction for the novice.

loz-hurst:
I think I'd like to start  a discussion along the lines of "should we re-think the approach to multiple platforms" and look at if we can replace existing examples etc. with ones that work on all 3 major platforms without any major changes (e.g. just using '\\' instead of '/' on Windows.

This would require a through going through of all the existing material to achieve, but should result in less GitHub issues, less confusing, simpler and easier to maintain material.

What do people think?

loz-hurst:
Can someone label this issue 'discussion' - I don't seem to have rights to do it?

gcapes:
Thanks for raising this issue.

For the purposes of a novice learning to use a tool (or indeed as an experienced user using a tool), I think any details about the differences and similarities between the terms UNIX/Unix/Linux/Unix-like are irrelevant. This isn't a history lesson after all! That's not to say that we shouldn't use the terms correctly in the material: I think we should.

[...]

PR #801, opened 2018-06-12
Clarified use of Unix/Unix Shell

### LINKED ISSUE #779: Unix vs UNIX
The setup instruction has

> There are also options for identifying and downloading a Unix Shell program, a Linux/UNIX emulator, or a program to access a UNIX server.

I think that the mix use of Unix and UNIX is confusing for novices.

[Wikipedia](https://en.wikipedia.org/wiki/Unix) uses Unix instead UNIX and says

> The present owner of the trademark UNIX is The Open Group, an industry standards consortium. Only systems fully compliant with and certified to the Single UNIX Specification qualify as "UNIX" (others are called "Unix-like").
>
> By decree of The Open Group, the term "UNIX" refers more to a class of operating systems than to a specific implementation of an operating system (...)

and

> Unix was the original formatting, but the usage of UNIX remains widespread because it was once typeset in small caps (...)

My recommendation is to only use Unix.

@swcarpentry/shell-maintainers Could you vote on this? Thanks.

loz-hurst:
Consistent use of 'Unix' rather than a mix of 'UNIX' and 'Unix'.

Replaced 'UNIX Server' with 'server with a Unix Shell', and similar reference under Windows section - server does not have to be Unix (could be Linux or even Windows with Bash installed) to be used for this lesson.

This improves accuracy of the text and, I think, is less confusing by talking about the 'Unix Shell' (the topic of the course) rather than 'UNIX' when the actual operating system is irrelevant.

References #798 and closes #779

---

Please keep in mind that lesson maintainers are volunteers and it may be some time before they can respond to your contribution. Although not all contributions can be incorporated into the lesson materials, we appreciate your time and effort to improve the curriculum.  If you have any questions about the lesson maintenance process or would like to volunteer your time as a contribution reviewer, please contact Kate Hertweck (k8hertweck@gmail.com).  

---

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -17,12 +17,12 @@ In the lesson, you will find out how to access the data in this folder.
 >
 > Some computers include a default Unix Shell program. 
 > The steps below describe some methods for identifying and opening a Unix Shell program if you already have one installed. 
-> There are also options for identifying and downloading a Unix Shell program, a Linux/UNIX emulator, or a program to access a UNIX server. 
+> There are also options for identifying and downloading a Unix Shell program, a Linux/UNIX emulator, or a program to access a Unix Shell on a server. 
 >
-> If none of the options below address your circumstances, try an online search for: UNIX shell [your computer model] [your operating system].
+> If none of the options below address your circumstances, try an online search for: Unix shell [your computer model] [your operating system].
 >
 > ### Linux
-> The default shell for Linux operating systems is usually Bash.
+> The default Unix Shell for Linux operating systems is usually Bash.
 > On most versions of Linux, it is accessible by running the [(Gnome) Terminal](https://help.gnome.org/users/gnome-terminal/stable/)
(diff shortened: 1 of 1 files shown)


setup.md
```diff
@@ -50,16 +50,16 @@ In the lesson, you will find out how to access the data in this folder.
 >
 > Other solutions are available for running Bash commands on Windows. 
 > There is now a Bash shell command-line tool available for Windows 10. 
-> Additionally, you can run Bash commands on a remote UNIX computer or server from your Windows machine. 
-> This can be done through a Secure Shell (SSH) client. 
+> Additionally, you can run Bash commands on a remote computer or server that already has a Unix Shell, from your Windows machine. 
+> This can usually be done through a Secure Shell (SSH) client.
```
shwina: Why usually? I realize that the remote computer needs to accessible (network), and running an SSH server, but do we really want to explain this?

PR #807, opened 2018-06-27
Revised functionality of 'sort -r'

brooksph:
In this case sort -r is sorting reverse chronological order, not reverse alphabetical.


shwina:
Thanks! I think the original text is strictly speaking not false, but I agree that it is confusing to call the ordering "alphabetical" for dates.

Issue #804, opened 2018-06-19
explain -i flag the first time it is introduced


beedatadriven:
The -i flag is first introduced in the solution for the "Using rm Safely" challenge activity.  I would suggest adding here that -i stands for interactive, just so it is spelled out the very first time it is introduced.

gcapes:
Fixed in #887

PR #811, opened 2018-07-03
Explain how to interact with rm -i prompts in episode 3

cvisger:
This was added to address issue 804 seen here https://github.com/swcarpentry/shell-novice/issues/804


shwina:
Maybe we should replace the original text

> The -i option will prompt before every removal.

with this new explanation?

gdevenyi:
Agree with @shwina can you also remove the initial description quoted. Thanks.

gcapes:
Fixed in #887

PR #812, opened 2018-07-05
Move introduction of wildcards from episode 4 to episode 3

### LINKED ISSUE #699: 03-Working With Files and Directories - * used in solution before introduced
The answer to the exercise in "Copy a folder structure sans files" uses a wildcard (*), but that is only introduced in the next section, 04-Pipes and Filters. I would suggest at least briefly introducing it previously, perhaps with ls. Perhaps with an example of wanting to see only the .dat files in a directory, using $ ls *.dat. This could have the added benefit of showing the importance of choosing consistent file names that make sense.

gcapes:
Fix #699.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -604,6 +604,150 @@ but it does find the copy in `thesis` that we didn't delete.
 > {: .solution}
 {: .challenge}
 
+> ## Copy with Multiple Filenames
+>
+> For this exercise, you can test the commands in the `data-shell/data` directory.
+>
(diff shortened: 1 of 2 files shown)


gdevenyi:
Hi,

Coming in very late on the discussion, I agree with the move, but this PR needs a rebase as the repo has drifted a bit, can you fix the merge conflicts? Thanks!

gcapes:
@gdevenyi 
Done!

OKaluza:
I just noticed the key points needed updating too, creating a pull request now

PR #813, opened 2018-07-11
Emphasise OS variations for finding help

### LINKED ISSUE #543: `ls --help` step should be made more compatible (`illegal option` in MacOS)
In "Navigating Files and Directories", the command `ls --help` when run on MacOS systems yields the error: `ls: illegal option --`. I realise that if we make accomodations for syntax for every single *nix variant, the lesson will quickly lose its simplicity and flow. However when I taught this lesson, around 50% of the class had Macs and were following along with the shell, so this little difference affects a large percentage of people. Another reason why I think this should be fixed is the position in the lesson. `ls --help` is the fifth command the students type in; I think commands should be completely compatible at the beginning to build confidence. If the very command that gives them help when they're stuck won't work for them, it doesn't build much confidence. 

When I taught this lesson recently, I needed to explain why there were differences between *nix commands and syntaxes; I fielded a couple questions on this, and it seemed like the wrong point in the lesson to introduce complexity.

gcapes:
Fix #543 (although most of the work was done in #690)

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -177,17 +177,6 @@ to use a command and what flags it accepts:
 (either `man` or `--help`).**
 We'll describe both ways below.
 
-Of course there is a third way: searching the internet via your web browser. 
-When using internet search, including the phrase `unix man page` in your search
-query will help to find relevant results.
-
-> ## Manual pages on the web
->
-> GNU provides links to its
-> [manuals](http://www.gnu.org/manual/manual.html) including the
-> [core GNU utilities](http://www.gnu.org/software/coreutils/manual/coreutils.html),
-> which covers many commands introduced within this lesson.
-{: .callout}
 
 #### The `--help` flag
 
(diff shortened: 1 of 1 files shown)


shwina:
Perhaps this can be more general using the following language:

> Depending on your environment (operating system or terminal application), you may find that only one of these works (either `--help` or `man`).

?

gcapes:
I like it! I've amended that commit (more or less) as suggested.

shwina:
@gcapes - doesn't look like those changes were made

shwina:
It's strictly not the terminal application (which has almost nothing to do with it) but the shell implementation. Perhaps we can leave that bit out. Sorry!

gcapes:
I think I forgot to commit or something. D'oh! Anyway, should all be good now.

shwina:
This might be a different PR, but do you think that now the placement of the following is a bit awkward:

> Of course there is a third way: searching the internet via your web browser. When using internet search, including the phrase unix man page in your search query will help to find relevant results.

Maybe it should go after the descriptions for `man` and `--help`?

gcapes:
Yes, I thought that it was a bit awkward, which is why I reworded slightly to indicate that the explanation would come below. But you're right it makes more sense to just reorder things. I'll add a commit to do that.

shwina:
This is great - thank you!

Issue #784, opened 2018-05-10
Clear Command


hadimoh80:
clear command will help students to clean their terminal screen. This aims student to see what instructor uses during the teaching. I have been asked, "how to clear up my terminal and start from top of the page". 

This will take less than minutes for discussion, and can be located at the Working With Files and Directories session. 
---

Thanks for contributing! If this contribution is for instructor training, please send an email to checkout@carpentries.org with a link to this contribution so we can record your progress. You’ve completed your contribution step for instructor checkout just by submitting this contribution.  

Please keep in mind that lesson maintainers are volunteers and it may be some time before they can respond to your contribution. Although not all contributions can be incorporated into the lesson materials, we appreciate your time and effort to improve the curriculum.  If you have any questions about the lesson maintenance process or would like to volunteer your time as a contribution reviewer, please contact Kate Hertweck (k8hertweck@gmail.com).  

---

shwina:
I do introduce `clear` anyway since I have a terrible habit of typing it in sometimes without thinking. +1 for including it in the lesson materials. Perhaps as a callout box?

I've also found that learners often worry that typing in `clear` "undoes" the work they have done so far - like MATLAB's `clear`, which deletes all variables in the workspace. Perhaps it's important to mention that `clear` really only just clears the screen.

After typing in `clear`, some learners are also disappointed that they've now lost history (the commands that they've typed in and the output from each command). I just tell them not to worry so much, as this information is all there in the lesson materials, but I can see why they might be irked. So possibly a warning about this is also needed.

gcapes:
> After typing in clear, some learners are also disappointed that they've now lost history (the commands that they've typed in and the output from each command). 

Have they really lost it? Can't you just scroll back up to see all the previous commands? It doesn't seem to delete them from history for me either.

I think the distinction from MATLAB's `clear` is probably a useful one to make.

hadimoh80:
I would say maybe we need to have "clear" and "history" command under the same title.

shwina:
@gcapes 

> Have they really lost it? Can't you just scroll back up to see all the previous commands?

I think it depends on the terminal emulator used. In `gnome-terminal`, I can indeed scroll back up to see the previous commands, but in `xterm` I cannot,.

>  It doesn't seem to delete them from history for me either.

Yes, the `history` command will still produce a history of commands typed in, but learners are sometimes interested in also the *output* of those commands.

PR #815, opened 2018-07-16
Inserting clear terminal command

AbernathyClose:
In response to Issue #784, I inserted a small section that explains how to clear your terminal using the clear command. It is located after the "Listing Recursively and By Time" section.

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -388,6 +388,13 @@ To **quit** the `man` pages, press <kbd>Q</kbd>.
 > {: .solution}
 {: .challenge}
 
+> ## Clearing your terminal
+>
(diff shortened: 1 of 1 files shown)


gdevenyi:
Thanks for the addition!

Can you also add the ``clear`` command you can type? Thanks.

gcapes:
While keyboard shortcuts are awesome, I would be wary of introducing them. They're harder to follow (can't see what the instructor is typing) and remember than a command.

But if this is going into the lesson, could someone just check the keyboard shortcuts also work on Git Bash and Mac?

gdevenyi:
Good call @gcapes, I had no idea  this shortcut existed and I'm not sure it works on non-linux terminals

gdevenyi:
No response from contributor, closing.

You're welcome to reopen if you have time to address our requests.

Thanks.

Issue #808, opened 2018-06-27
Episode 7: find-file-tree.svg is illegible


ddietz89:
The file structure graphic is illegible for the `find` diagram. I have to zoom in several times to be able to read the leafs on the tree.

lxvii:
It looks like @cekees resolved this issue (#808) by amendment #816 but it is still showing as open

gcapes:
That PR was closed without merging. I'll fix it in a new PR.

PR #816, opened 2018-07-19
Episode 7: increase size of File Tree figure

cekees:
Attempting to address #808


gdevenyi:
Hi.

We would vastly prefer to avoid adding HTML to the existing lessons. Rather than editing the code, could you please modify size of the SVG file which is in the repo. Thanks!

PR #820, opened 2018-07-25
Word count output clarified

hburns2:
The output of the word count demonstration is unclear. Doesn't specify what the columns line up to.


gcapes:
I would suggest editing the explanation rather than the output of the command which would be misleading.

hburns2:
Absolutely. I will edit that

PR #827, opened 2018-08-15
Fix typo in 01-intro.md

nbren12:
Change 'erorr' to 'error'.

I noticed this small typo and figured I may as well fix it as part of the instructor checkout procedure.


gdevenyi:
Thanks!

PR #830, opened 2018-08-16
Update 02-filedir.md

nairps:
Updated the objectives, questions and keypoints
Corrected some typos


gdevenyi:
Thanks for the suggestions, however submissions like this are very difficult to discuss and evaluate. Can you please break up these into separate PRs so they can be discussed individually?

Thanks!

Issue #794, opened 2018-06-10
03-create "Using rm Safely" removes a file we have not created


primerano:
At the point that this challenge is given the thesis directory has been removed and creating quotations.txt is a few steps away

Using rm Safely
What happens when we type rm -i thesis/quotations.txt? Why would we want this protection when using rm?

Rather than depend on the order we could change the lesson to have them create a file and try to remove it with rm -i.

Maybe something like
---------------
Using rm Safely

Let's create a file called delete_me.txt

touch delete_me.txt 

What happens when we type rm -i delete_me.txt? Why would we want this protection when using rm?

--------------

Thoughts?

gcapes:
Good idea! The exercise should certainly be do-able at the place it appears in the lesson, and if it's independent of position, then so much the better.

PR #832, opened 2018-08-17
Updated 03-create.md to create a file before demonstrating safe deletion

dpnemergut:
This PR resolves issue #794 to create a new file before using `rm -i` on it since `thesis/quotations.txt` doesn't exist at this point in the lesson.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -312,12 +312,19 @@ $ rm -r thesis
 
 > ## Using `rm` Safely
 >
-> What happens when we type `rm -i thesis/quotations.txt`?
+> Let's create a file called delete_me.txt:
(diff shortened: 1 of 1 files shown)


gcapes:
I think this needs to incorporate some of the `with great power ...`  call out box. As proposed, the exercise is still before the call out, and the code in the call out won't work. I suggest we drop the call out, but make sure we haven't lost any of the explanation - so explaining that we need to enter `y` or `n` to delete or not delete each file would need adding.

Thoughts @gdevenyi ?

colinmorris:
+1 on merging the "Using rm Safely" exercise and "With Great Power..." callout box. These pieces feel pretty duplicative.

PR #833, opened 2018-08-18
WIP: Add some instruction on what to do when nano is run.

scttfrdmn:
This PR is in reference to my comment issue #585 - it adds some language explaining what to do with nano in episode 3. The current lesson just instructs that nano should be run followed by an `ls` command. 

This PR is also in partial fulfillment of my instructor training.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -126,12 +126,12 @@ $ nano draft.txt
 >
 > When we say, "`nano` is a text editor," we really do mean "text": it can
 > only work with plain character data, not tables, images, or any other
-> human-friendly media. We use it in examples because it is one of the 
-> least complex text editors. However, because of this trait, it may 
+> human-friendly media. We use it in examples because it is one of the
+> least complex text editors. However, because of this trait, it may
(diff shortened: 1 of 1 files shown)


_episodes/03-create.md
```diff
@@ -273,6 +273,12 @@ $ pwd
 
 ~~~
 $ nano draft.txt
+~~~
+{: .language-bash}
+
+Let's again type in a few lines of text. Once we're happy with our text, we can press `Ctrl-O` to write our data to disk. Once our file is saved, we can use `Ctrl-X` to quit the editor and return to the shell.
```
gcapes: Is there a step missing here? Nano will prompt to confirm the file name when saving, which requires the user to press *enter*, which is different to the exit command.

_episodes/03-create.md
```diff
@@ -362,6 +368,12 @@ $ pwd
 ~~~
 $ mkdir thesis
 $ nano thesis/draft.txt
+~~~
```
gcapes: This is largely repetition of the above. I would think it best to avoid repeating in the lesson text even if you would repeat it verbally as an instructor?

PR #842, opened 2018-09-15
Added an additional grep exercise 

rwcode:
Added an additional grep exercise to practice flags -i and -w with haiku.txt

### DIFF
--- a/_episodes/07-find.md
+++ b/_episodes/07-find.md
@@ -1,7 +1,7 @@
 ---
 title: "Finding Things"
 teaching: 25
-exercises: 20
+exercises: 21
 questions:
 - "How can I find files?"
 - "How can I find things in files?"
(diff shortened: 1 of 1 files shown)


_episodes/07-find.md
```diff
@@ -249,6 +249,34 @@ Miscellaneous:
 > {: .solution}
 {: .challenge}
 
+> ## More `grep` practice
+>
+> What would be the correct command for the of the following text :
```
gcapes: It's not clear what the question is asking. Aside from fixing the existing wording, perhaps stating in words what pattern the learner should search for, show the desired result (as you already have done) and ask the learner to choose the right command would be a better question format?

colinmorris:
Thanks for the PR. I agree with gcapes that the intent of the question is a bit hard to parse at first. I'd suggest changing the first line to something like...

> Which `grep` command will produce the following output (i.e. all lines containing the word "is")

But that issue aside, I think this particular episode already has pretty good exercise coverage, and the proposed exercise is fairly similar to the `grep -n -w -i "the" haiku.txt` example from the main body, and to the first exercise. So while I think it's a good exercise, I don't think it adds enough to justify the additional real estate.

(Sorry to be a party pooper, but the shell lesson is a fairly mature one, so we tend to be conservative about adding new content.)

PR #843, opened 2018-09-18
Improve wording in Nelle's pipeline in 1st episode (Introducing the Shell)

nistara:


gdevenyi:
Thanks for the suggestion, however, "shell command" and "command shell" have different meanings, and in this case, "command shell" is the meaning we're going for here, meaning "a shell in which we can run typed commands".

nistara:
Thanks for the clarification!

PR #857, opened 2018-10-31
Give justice to pdfgrep

prosoitos:


gdevenyi:
Thanks for the suggestion, however we're going to stick to POSIX or POSIX+GNU extensions (still a bit of argument on that part...)

pdfgrep is way outside that and we can't rely on people having it available for lessons.

prosoitos:
Note: I edited my PR title which was poorly chosen and misleading on my purpose.

prosoitos:
My new PR could look like this.

Replace:

> We have focused exclusively on finding things in text files. What if your data is stored as images, in databases, or in some other format? One option would be to extend tools like grep to handle those formats. This hasn’t happened, and probably won’t, because there are too many formats to support.
>
> The second option is to convert the data to text, or extract the text-ish bits from the data (...)

With:

> We have focused exclusively on finding [patterns] in text files. What if your data is stored as images, in databases, or in some other format? [A handful of tools extend grep to handle a few non text formats. But an approach which works for any format is to convert the data to text, or extract the text-ish bits from the data (...)]

PR #858, opened 2018-11-01
Remove inaccurate statement

prosoitos:
This is a follow-up to PR #857, which got closed, I believe, because I was not explicit enough about what I was doing and chose a poor title.


gdevenyi:
Thanks for your contribution, this is an excellent rewording of the section to be more clear and concise.

prosoitos:
Thank you for reconsidering it once I took the time to explain better what I was doing!

PR #859, opened 2018-11-05
Fixed a tiny typo.

ekaterinailin:


gcapes:
Thanks for the contribution. 'They' refers to the two different methods, so this isn't a typo I'm afraid.

ekaterinailin:
I see, thanks! It was a readability hiccup.

gdevenyi:
Thanks!

PR #867, opened 2018-11-27
Option 4 solution should not append

peterjc:
As written it appends (``>>``) so would not necessarily give the desired output.


gdevenyi:
Good catch!

Issue #828, opened 2018-08-15
Suggested edit to a lesson


ppprudh:
I would like to suggest possible editing of the layout of the lesson, Working With Files and Directories in The Unix Shell. https://swcarpentry.github.io/shell-novice/03-create/index.html

Following the sentence, “We now know how to explore files and directories, …” 

could the following three sections be moved right after it

Which Editor?
Good names for files and directories
What’s in a name?

in order to give users a more general understanding of guidelines, recommendations, and editing options prior to getting back to the “data-shell directory in the desktop...introducing pwd, etc.?

After “What’s In A Name” or (What’s in a filename?), the lesson remains the same at the exception of another edit of the “Renaming Exercise”, possibly where the text could recall the “What’s In A Name?” section or (What’s in a filename?)
  
Thanks,

Patrice-Andre Prud’homme -- 2018-08-15

gdevenyi:
Thanks for the suggestion, we welcome changes in the form of a pull request making the changes you suggested.

PR #871, opened 2018-12-04
Issue828 rearrangements in working with files

schmeing:
Did part of what was suggested in Issue #828 "Suggested edit to a lesson"

I moved up the "Good names for files and directories" and "What’s in a name?" boxes. However, not completely up to the first sentence as I think it helps to visually see some examples and the ls -F command two lines after does exactly that.

The "Which Editor?" box I left untouched as I think it is overloading the students to put all that info before you are using it, so I left it right in front of the first use of the nano editor.

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -23,7 +23,6 @@ keypoints:
 - "An absolute path specifies a location from the root of the file system."
 - "Directory names in a path are separated with `/` on Unix, but `\\` on Windows."
 - "`..` means 'the directory above the current one'; `.` on its own means 'the current directory'."
-- "Most files' names are `something.extension`. The extension isn't required, and doesn't guarantee anything, but is normally used to indicate the type of data in the file."
 ---
 
 The part of the operating system responsible for managing files and directories 
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -75,6 +75,30 @@ creatures/  data/  molecules/  north-pacific-gyre/  notes.txt  pizza.cfg  solar.
 > or another non-alphanumeric character, you should surround the name in quotes (`""`).
 {: .callout}
 
+> ## What's In A Name?
+>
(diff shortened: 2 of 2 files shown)


colinmorris:
Thanks for the PR (and sorry for the late response). Moving that key point from lesson 2 to 3 is definitely a great correction.

I'm ambivalent on rearranging the info boxes in this way though. I tend to think of content in the 'callout' boxes as optional or supplementary, so I think it makes sense to put them after the main content to which they're relevant. e.g. to have an example of creating a file or directory, then have a callout box with tips on naming style.

I'd be curious to hear other perspectives on this though.

gcapes:
Thanks for the contribution @schmeing 
I agree with @colinmorris  that the current position of the callouts makes sense - in general the lessons  do something first, then expand on it when the learner has something to 'hang' the new information on. 

If everyone agrees, could you reset your branch so that the PR only includes commit https://github.com/swcarpentry/shell-novice/pull/871/commits/02296dfe11e28bfe7030b8cc7beacb630fe44656. Do shout if you're unsure about the git side of things.

schmeing:
Reverted the 2 commits as asked during the merge with the base
I still think "What's in A Name" and "Good names for files and directories" belong together as they are both about how to name files and directories.
The advantage of having the "What's in A Name" at the same position where "Good names for files and directories" currently is, is that there are multiple examples of file endings and not just .txt shown in the ls.
The disadvantage is that it would be a huge chunk of additional information right there.

gcapes:
Thanks @schmeing.

I see no reason not to move the `What's in a name` callout into the `Create a text file` section. I don't think it should go right after the `Good names for files and directories` callout because it only refers to files, and at this point I think we need to be careful to avoid cognitive overload.

I have a large PR (#887) waiting review, which fixes several issues - I'll tag this suggestion onto that.

gcapes:
Reset and merged in f82338e80d9d9dfee82a569032e3acfaef060c2e

PR #873, opened 2018-12-11
Move text to more logical position

gcapes:
The exercises and help section seemed to split this section before a
logical break.


colinmorris:
Whoops, great catch.

PR #878, opened 2018-12-27
How to find the "|" - Update 04-pipefilter.md

TorstenPaul:
Added some line to type the vertical bar if someone cant't find it on his keyboard. ALT+124
"To type the vertical bar `|` if one can't find it on the keyboard
an easy way is to hold 'ALT' and type 124 for the ASCII code **ALT+124**."

### DIFF
--- a/_episodes/04-pipefilter.md
+++ b/_episodes/04-pipefilter.md
@@ -331,6 +331,8 @@ The vertical bar, `|`, between the two commands is called a **pipe**.
 It tells the shell that we want to use
 the output of the command on the left
 as the input to the command on the right.
+To type the vertical bar `|` if one can't find it on the keyboard
+an easy way is to hold 'ALT' and type 124 for the ASCII code **ALT+124**.  
 The computer might create a temporary file if it needs to,
 or copy data from one program to the other in memory,
 or something else entirely;


gcapes:
This gets a '-1' from me I'm afraid, because it doesn't work on all machines. I've not had this problem when teaching, so I would think this would be useful to only a very small set of learners.

ErinBecker:
Thanks for the contribution @TorstenPaul. I have had learners have problems finding the `|` key on their keyboards - particularly learners who are using non-English language keyboards. Has anyone else had this issue in their workshops? It is a very uncommonly used key. 😄

TorstenPaul:
> 
> 
> Thanks for the contribution @TorstenPaul. I have had learners have problems finding the `|` key on their keyboards - particularly learners who are using non-English language keyboards. Has anyone else had this issue in their workshops? It is a very uncommonly used key. 😄

Hey @ErinBecker, yes during all workshops I attended as helper we had this problem. For example a Korean student with an Chinese keyboard layout...
Maybe we can make a collection with all ways to create this key :) for the instructors and helpers.

gcapes:
I guess I've not had the issue because the first couple of times I use it, I usually say 

> it's  to the left of the 'Z' key on my keyboard

But that's for a UK keyboard. I would be keen to avoid adding more exceptions to the lesson (I think the ALT code only works on Windows):  
- where is the pipe symbol on a Mac keyboard? 
- what about a US keyboard layout?
- what about other characters, like `#` which [doesn't exist on a Mac keyboard](https://machash.com/mac-hash-key/)?

I think somewhere in `extras` would be a more suitable place for this type of information?

rgaiacs:
@TorstenPaul Thanks for the pull request. I agree that `|` is hard to find the first time but your solution adds extra complexity to the lesson. I don't have experience with Korean, Chinese or Japanese keyboards so I can't say how to find `|` on those keyboards but my experience with keyboards from South America, Europe and Africa is that the `|` is going to be placed in different places depending of the country where you got your machine, the manufacturer of your machine, the size of your machine (i.e. does it have a full keyboard with numpad?) and, to make things worse, some learners use a mapping that is different from the physical keyboard.

# Layouts

"Default" keyboard in Brazil

br

"Default" keyboard in Portugal

pt

Note that even as the "same" language, the `|` is in different parts of the keyboard.

"Default" keyboard in France

azer

Yes, France doesn't use the QWERTY layout.

"Default" keyboard in Argentina

es

# Final words

> I think somewhere in extras would be a more suitable place for this type of information?

I think is is one of the cases where we should ask learners to use the red sticker if they can't find `|` in their keyboard.

ErinBecker:
Thanks @gcapes and @rgaiacs for the thoughts. Maybe we could add something to the instructor notes for this?

gcapes:
@TorstenPaul 

Thanks for the PR, but I think that the suggestion from @rgaiacs to use a red sticky note is the most appropriate response to this situation.

Issue #879, opened 2018-12-28
Lesson 03-create needs sections


DSLituiev:
The first lesson has clear markdown headers, but [lesson 3](https://swcarpentry.github.io/shell-novice/03-create/index.html) does not. I suggest one can add headers to help navigating the content, e.g.:

```
# Creating files and directories
## See where we are
## Create a directory
## Creating text files
# Removing files and directories
# Moving files and directories
# Copying files and directories
# Operations with multiple files and directories
```

colinmorris:
Fixed in #880. Thanks!

Though this made me realize that lessons 2, 4, 5, 6, and 7 also have no section headers (except one at the end for "Nelle's pipeline"). It could be worth updating those lessons in the same way.

PR #880, opened 2018-12-28
Section structure added

DSLituiev:
This pull request addresses the issue #879

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -20,9 +20,11 @@ keypoints:
 - "The shell does not have a trash bin: once something is deleted, it's really gone."
 - "Depending on the type of work you do, you may need a more powerful text editor than Nano."
 ---
-
+## Creating directories
 We now know how to explore files and directories,
 but how do we create them in the first place?
+
+### Step one: see where we are and what we already have
 Let's go back to our `data-shell` directory on the Desktop
 and use `ls -F` to see what it contains:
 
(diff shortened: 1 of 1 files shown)


colinmorris:
Looks great. Thanks for the wording tweaks as well.

_episodes/03-create.md
```diff
@@ -152,7 +157,7 @@ holding it down, press the O key) to write our data to disk
 (we'll be asked what file we want to save this to:
 press Return to accept the suggested default of `draft.txt`).
 
-![Nano in Action](../fig/nano-screenshot.png)
+<div style="width:80% margin: auto;">![Nano in Action](../fig/nano-screenshot.png)</div>
```
colinmorris: I just reverted the change on this line because I noticed the inner markdown wasn't being rendered (i.e. it was showing up as the literal text `Nano in Action` rather than as an image). 

If you still want to make this change, the simplest thing to do might be to replace the markdown with just an `` element. There may be other workarounds though ([relevant StackOverflow post](https://stackoverflow.com/questions/29368902/how-can-i-wrap-my-markdown-in-an-html-div)).

DSLituiev: OK. The reason I added that is that the image appeared oversize making it the centerpiece in a lesson where it was supposed to be just a side note. Here is a fully HTML suggestion

```suggestion
<div style="width:80% margin: auto;"><img alt="Nano in Action" src="../fig/nano-screenshot.png"></div>
```

colinmorris: [Done](http://swcarpentry.github.io/shell-novice/03-create/index.html). Thanks!

PR #881, opened 2019-01-02
Update 04-pipefilter.md

sjmillerAZ:
change language describing wc -l
remove extraneous prompt char $ in less command

---
"shortest" is confusing when referring to files. for example, a file with one long line could be "longer" (in terms of number of bytes) than a file with a few very short lines.

---


colinmorris:
Looks good. Thanks!

PR #884, opened 2019-01-07
Render keyboard keys as described by the style guide

gcapes:


gcapes:
This would supersede and hence fix #561

PR #887, opened 2019-01-11
Reorder and rewrite episode 3

### LINKED ISSUE #585: Using nano to create empty file

gcapes:
This supersedes / fixes #833 and fixes #585.

The main benefit is to avoid creating the draft /quotes file several times.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -233,110 +233,6 @@ draft.txt
 > {: .solution}
 {: .challenge}
 
-## Removing files and directories
-
-Returning to the `data-shell` directory,
-let's tidy up the `thesis` directory by removing the draft we created:
-
-~~~
-$ cd thesis
-$ rm draft.txt
-~~~
-{: .language-bash}
-
-This command removes files (`rm` is short for "remove").
-If we run `ls` again,
-its output is empty once more,
-which tells us that our file is gone:
-
-~~~
-$ ls
-~~~
-{: .language-bash}
-
-> ## Deleting Is Forever
->
-> The Unix shell doesn't have a trash bin that we can recover deleted
-> files from (though most graphical interfaces to Unix do).  Instead,
-> when we delete files, they are unhooked from the file system so that
-> their storage space on disk can be recycled. Tools for finding and
-> recovering deleted files do exist, but there's no guarantee they'll
-> work in any particular situation, since the computer may recycle the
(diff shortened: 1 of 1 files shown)


gdevenyi:
Hi,

I'm going to take a look at this, before I start, is this a "rewrite" or a copy+paste rearrange + changes?

gcapes:
It's the latter. Have a read through the commit messages before you start and that should give you a good summary.

colinmorris:
Fixes #832. (I don't think GitHub's parser picked that up from the original comment)

gcapes:
I think this also fixes #811

_episodes/03-create.md
```diff
@@ -618,6 +485,104 @@ but it does find the copy in `thesis` that we didn't delete.
...
+```
+ls: cannot access 'quotes.txt': No such file or directory
+```
+{: .output}
+
+> ## Deleting Is Forever
+>
+> The Unix shell doesn't have a trash bin that we can recover deleted
+> files from (though most graphical interfaces to Unix do).  Instead,
+> when we delete files, they are unhooked from the file system so that
```
gdevenyi: I think the term is "unlinked" rather than unhooked

_episodes/03-create.md
```diff
@@ -618,6 +485,104 @@ but it does find the copy in `thesis` that we didn't delete.
...
+If you haven't already deleted the `thesis` directory using the interactive flag,
+you can delete the directory without any confirmation prompts using the following command:
+
+~~~
+$ rm -r thesis
+~~~
+{: .language-bash}
+
+
+> ## Using `rm` Safely
```
nsoranzo: I'd move the "Using `rm` Safely" exercise right after the "Deleting Is Forever" callout since at that point we are still dealing with removing files and it would make the "With Great Power Comes Great Responsibility" easier to understand.

gcapes: Yes, good idea.

nsoranzo: Thanks!

gcapes:
Anything outstanding, or can this be merged now?

PR #888, opened 2019-01-11
Update 02-filedir.md

jamesjbrownjr:
The "Exploring more ls Flags" section does not make clear that the -h flag only works the way explained here if it is used along with -l. In order to avoid confusion, this question and solution should indicate that you can use two flags at once and then explain what happens when using -l and then -lh separately.


gdevenyi:
Thanks for the clarification!

PR #896, opened 2019-01-17
Edited The Unix Shell: Loops

AdnanQureshi2810:
The loop lesson starts with cp command error which can be covered in the navigation lesson. This lesson should start with the head -n3 command for the individual files and the use of loop command to circumvent using individual commands for each file.

### DIFF
--- a/_episodes/05-loop.md
+++ b/_episodes/05-loop.md
@@ -31,31 +31,24 @@ we'll use the `creatures` directory which only has two example files,
 but the principles can be applied to many many more files at once.
 We would like to modify these files, but also save a version of the original files, naming the copies
 `original-basilisk.dat` and `original-unicorn.dat`.
-We can't use:
 
 ~~~
-$ cp *.dat original-*.dat
+$ mv basilisk.dat original-basilisk.dat
+$ mv unicorn.dat original-unicorn.dat
+
 ~~~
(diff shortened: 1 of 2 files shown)


_episodes/05-loop.md
```diff
@@ -31,31 +31,22 @@ we'll use the `creatures` directory which only has two example files,
 but the principles can be applied to many many more files at once.
 We would like to modify these files, but also save a version of the original files, naming the copies
 `original-basilisk.dat` and `original-unicorn.dat`.
-We can't use:
 
 ~~~
-$ cp *.dat original-*.dat
+$ mv basilisk.dat original-basilisk.dat
```
gcapes: What is the purpose of renaming the files here?

gcapes:
Thanks for the contribution.

However, I think it's rather confused. We need to cover how to copy multiple files, and starting with the error example gives a good motivation for learning loops. If it were to move into the navigation episode (2) we would lose the connection and I think it would frustrate learners to be shown something they might reasonably try, but which doesn't work, only to have to wait another 2 episodes until a solution is shown.

I'm afraid the code modifications you made break things later in the lesson (I've commented on the line itself).

Perhaps you could create an issue to discuss the problem you'd looking to fix?

Thanks

PR #902, opened 2019-01-20
Fixes for "Creating files a different way"

nsoranzo:
- No need to go to the home directory
- Enclose `touch` and `my_file.txt` in tick marks
- Use "0 bytes" instead of "0kb" for the empty file size

Also add a missing period in "Moving files and directories".

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -215,7 +215,7 @@ draft.txt
 >
 > > ## Solution
 > > 1.  The `touch` command generates a new file called `my_file.txt` in
-> >     your home directory.  If you are in your home directory, you
+> >     your current directory.  You
 > >     can observe this newly generated file by typing `ls` at the 
 > >     command line prompt.  `my_file.txt` can also be viewed in your
 > >     GUI file explorer.
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -195,7 +195,7 @@ draft.txt
 ~~~
 {: .output}
 
-> ## Creating files a different way
+> ## Creating Files a Different Way
 >
 > We have seen how to create text files using the `nano` editor.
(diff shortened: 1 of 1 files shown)


gcapes:
Hi @nsoranzo thanks for the PR!

I like the changes, but we need to be consistent. If we remove the instruction to `cd` to the home directory, the solution shouldn't refer to the home directory.

I'm putting this on hold until #887 is merged, which is a significant rearrangement of episode 3.

Nonetheless, could you fix up the solution?
Thanks

_episodes/03-create.md
```diff
@@ -215,15 +214,15 @@ draft.txt
 > 3.  When might you want to create a file this way?
 >
 > > ## Solution
-> > 1.  The touch command generates a new file called 'my_file.txt' in
+> > 1.  The `touch` command generates a new file called `my_file.txt` in
```
gcapes: `touch` creates a new file in the current directory, which isn't necessarily the home directory.

_episodes/03-create.md
```diff
@@ -215,15 +214,15 @@ draft.txt
...
 > >     your home directory.  If you are in your home directory, you
```
gcapes: Again, the reference to the home directory is misleading. Whichever directory you ran the example in, the `ls` command will show the newly created file.

nsoranzo:
@gcapes Thanks for the review, I've addressed your comments.

_episodes/03-create.md
```diff
@@ -195,18 +195,17 @@ draft.txt
 ~~~
 {: .output}
 
-> ## Creating Files a Different Way
+> ## Creating files a different way
```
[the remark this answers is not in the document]
nsoranzo: Done.

PR #903, opened 2019-01-22
Fixes for "Creating files a different way" 

nsoranzo:
- No need to go to the home directory
- Enclose `touch` and `my_file.txt` in tick marks
- Use "0 bytes" instead of "0kb" for the empty file size

Supersedes https://github.com/swcarpentry/shell-novice/pull/902 , rebasing it on top of #887.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -198,15 +198,14 @@ draft.txt
 > ## Creating Files a Different Way
 >
 > We have seen how to create text files using the `nano` editor.
-> Now, try the following command in your home directory:
+> Now, try the following command:
 >
 > ~~~
-> $ cd                  # go to your home directory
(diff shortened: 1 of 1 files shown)


_episodes/03-create.md
```diff
@@ -198,15 +198,14 @@ draft.txt
...
 >
 > ~~~
-> $ cd                  # go to your home directory
 > $ touch my_file.txt
 > ~~~
 > {: .language-bash}
 >
-> 1.  What did the touch command do?
+> 1.  What did the `touch` command do?
 >     When you look at your home directory using the GUI file explorer,
```
gcapes: This should refer to the *current* directory, not the *home* directory.

nsoranzo: Thanks, how could I miss that?!

PR #260, opened 2016-01-05
Added two exercises to 06-find/

cfgauss:
This is part of the completion of Software Carpentry instructor training.

PR #904, opened 2019-01-23
Add exercise using grep in a loop

gcapes:
Modified idea from #260 by @cfgauss.

Fix #260.

### DIFF
--- a/_episodes/07-find.md
+++ b/_episodes/07-find.md
@@ -382,6 +382,65 @@ Miscellaneous:
 > {: .solution}
 {: .challenge}
 
+> ## Using grep in a loop
+> Navigate to Nelle's `data-shell/data/pdb/` directory, which contains Protein Data Bank files.
+> Write a `for` loop which uses `echo` and `grep` to print each file name followed by the number
+> of atoms in the compound.
+>
+> Here's the beginning and end of the output your loop should return:
+> ```
+> aldrin.pdb 26
+> ammonia.pdb 4
+> ascorbic-acid.pdb 20
+> ...
+> vanillin.pdb 19
+> vinyl-chloride.pdb 6
+> vitamin-a.pdb 51
+> ```
+> {: .output}
+>
+> Hint: Read `man echo` for an option to suppress newlines
+> and `man grep` for an option to produce counts only.
+>
+> > ## Solution
+> > ```
+> > for file in *.pdb
+> >	do
+> >		echo -n $file " "
+> >		grep -c ATOM $file
+> >	done
+> > ```
+> > {: .language-bash}
+> {: .solution}
+>
+> Extend the previous solution to sort the output by the number of atoms.
+> Here's the beginning and end of the output:
+>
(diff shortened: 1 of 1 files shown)


_episodes/07-find.md
```diff
@@ -382,6 +382,65 @@ Miscellaneous:
...
+>
+> Hint: Read `man sort` and look at the `-k` option to sort by a field.
+>
+> > ## Solution
+> > ```
+> > for file in *.pdb
+> >	do
+> >		echo -n $file " "
+> >		grep -c ATOM $file
+> >	done | sort -k 2 -n
```
colinmorris: I don't think we've shown piping the output of a loop anywhere previously.

_episodes/07-find.md
```diff
@@ -382,6 +382,65 @@ Miscellaneous:
...
+> Hint: Read `man echo` for an option to suppress newlines
+> and `man grep` for an option to produce counts only.
+>
+> > ## Solution
+> > ```
+> > for file in *.pdb
+> >	do
+> >		echo -n $file " "
+> >		grep -c ATOM $file
+> >	done
```
colinmorris: It's possible to get functionally the same output with just `grep -c 'ATOM' *.pdb` (the only difference is that the name and counts are separated by colons rather than spaces). That makes this example a little less compelling to me.

_episodes/07-find.md
```diff
@@ -382,6 +382,65 @@ Miscellaneous:
...
+> methane.pdb 5
+> methanol.pdb 6
+> ...
+> cholesterol.pdb 74
+> heme.pdb 75
+> lanoxin.pdb 121
+> ~~~
+> {: .output}
+>
+> Hint: Read `man sort` and look at the `-k` option to sort by a field.
```
colinmorris: I think this is going to be difficult. The `man` entry for `-k` just says...

```
 -k, --key=KEYDEF
              sort via a key; KEYDEF gives location and type
```

If they search the man page for `KEYDEF`, they'll find...

```
   KEYDEF is F[.C][OPTS][,F[.C][OPTS]] for start and stop position, where F is a field number  and  C  a  character
   position  in  the field; both are origin 1, and the stop position defaults to the line's end.  If neither -t nor
   -b is in effect, characters in a field are counted from the beginning of the preceding whitespace.  OPTS is  one
   or  more  single-letter ordering options [bdfgiMhnRrV], which override global ordering options for that key.  If
   no key is given, use the entire line as the key.  Use --debug to diagnose incorrect key usage.
```
If I didn't already know what `-k` did, I would have no idea what any of this meant.

gcapes:
This was me rewriting an old PR, but you make good points. It doesn't address any new learning goals. I'm happy to abandon the PR unless @cfgauss has anything to add?

gdevenyi:
I think we can close for now. Thanks @gcapes and @colinmorris for the discussion, good catch on the learning goals.

Thanks @cfgauss for your contribution.

Issue #905, opened 2019-01-23
Improvement for "Find Things" lesson: more information on "grep" earlier in lesson


BenPVD:
I have a suggestion regarding the **Find Things** chapter, which I think will improve the learning flow.

At the first introduction, the current script does not mention that `grep` is case-sensitive by default. During the practical exercises and the following steps in the lesson, this is not noticeable at first. Not until much later in the script, the instruction follows: `Now we want to use the option -i to make our search case-insensitive`.

I think it is better to point out the essential characteristics of `grep` at the beginning and later introduce the flags that change this behavior. On the one hand, this improves the learning/teaching flow (during teaching we can reference back since we already mentioned that grep is case-sensitive per default), on the other hand, it will act as a repetition, which also improves learning.

gcapes:
Sounds good. Would you like to submit a pull request?

BenPVD:
> Sounds good. Would you like to submit a pull request?

Sure! I will submit a pull request.

PR #908, opened 2019-01-23
Pull request #905 from BenPVD/issue905

BenPVD:
This pull request adds some information about the default behavior of grep and resolves issue #905.

It has been tried to add as little new text as possible so that the lesson (07-find.md) is not unnecessarily lengthened.

The additions are intended to improve the flow of learning and teaching.


gcapes:
Great, thanks!

PR #913, opened 2019-01-30
Ep4: Explain echo command on first use

### LINKED ISSUE #877: echo command not explained on first use in pipes and filters

---
We have seen the use of >, but there is a similar operator >> which works slightly differently. By using the echo command to print strings, test the commands below to reveal the difference between the two operators:

Instead the echo command is explained in the next loops section. This explanation can be added where the command is first used.

The first, echo, just prints its command-line arguments to standard output. For example:
$ echo hello there
prints:
hello there
---

Murray2015:
Fixes #877

### DIFF
--- a/_episodes/04-pipefilter.md
+++ b/_episodes/04-pipefilter.md
@@ -260,10 +260,10 @@ the output of `head` must be the file with the fewest lines.
 > We have seen the use of `>`, but there is a similar operator `>>` which works slightly differently.
 > First, let's try using the `echo` command to print strings, e.g.: 
 > ~~~
-> $ The echo command prints text
+> $ echo The echo command prints text
 > ~~~
 > {: .language-bash}
-> By using the `echo` command to print strings, let's now test the commands below to reveal the difference
+> The `echo` command has many uses. Let's use the `echo` command to test the commands below to reveal the difference
 > between the two operators:
 >
 > ~~~
--- a/_episodes/04-pipefilter.md
+++ b/_episodes/04-pipefilter.md
@@ -256,10 +256,14 @@ the output of `head` must be the file with the fewest lines.
 > incorrect results and/or delete
 > the contents of `lengths.txt`.
 {: .callout}
-
 > ## What Does `>>` Mean?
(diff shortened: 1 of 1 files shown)


gcapes:
Thanks for the PR!

This doesn't render correctly I'm afraid - you've forgotten to show the command which gives the output. I'll put some more comments on the diff. Could you please fix these before I merge the PR.

thanks

_episodes/04-pipefilter.md
```diff
@@ -256,10 +256,14 @@ the output of `head` must be the file with the fewest lines.
...
 > the contents of `lengths.txt`.
 {: .callout}
-
 > ## What Does `>>` Mean?
 > We have seen the use of `>`, but there is a similar operator `>>` which works slightly differently.
-> By using the `echo` command to print strings, test the commands below to reveal the difference
+> First, let's try using the `echo` command to print strings, e.g.: 
+> ~~~
+> $ The echo command prints text
+> ~~~
```
gcapes: This is a hybrid of the output and the command.
The command should be
```
$ echo The echo command prints text
```

and the corresponding output should be 
```
The echo command prints text
```

_episodes/04-pipefilter.md
```diff
@@ -256,10 +256,14 @@ the output of `head` must be the file with the fewest lines.
...
+> {: .language-bash}
+> By using the `echo` command to print strings, let's now test the commands below to reveal the difference
```
gcapes: I'm not sure how this is an improvement over the original wording in line 262?

_episodes/04-pipefilter.md
Murray2015: Yep, what a daft mistake. Well spotted.

gcapes: Great, thanks. I think it would also be good to show the expected output in a code block.

gcapes:
Thanks! 
Merged with a few edits in 98be3175b19d6b8fcac0a31f69c9193bdb4ff6c7

PR #679, opened 2017-12-26
Update 03-create.md to include whitespace definition

kagross:
Including a definition for whitespace the first time that it comes into play in the lesson could erase some confusion for students that don't know the word.

Please delete the text below before submitting your contribution. 

---

Thanks for contributing! If this contribution is for instructor training, please send an email to checkout@carpentries.org with a link to this contribution so we can record your progress. You’ve completed your contribution step for instructor checkout just by submitting this contribution.  

Please keep in mind that lesson maintainers are volunteers and it may be some time before they can respond to your contribution. Although not all contributions can be incorporated into the lesson materials, we appreciate your time and effort to improve the curriculum.  If you have any questions about the lesson maintenance process or would like to volunteer your time as a contribution reviewer, please contact Kate Hertweck (k8hertweck@gmail.com).  

---

PR #915, opened 2019-02-04
Replace mentions of "whitespace" with "spaces"

colinmorris:
As discussed in #679.

### DIFF
--- a/_episodes/05-loop.md
+++ b/_episodes/05-loop.md
@@ -406,7 +406,7 @@ from whatever file is being processed
 > ~~~
 > $ for filename in "red dragon.dat" "purple unicorn.dat"
 > > do
-> >     head -n 100 "$filename" | tail -n 3
+> >     head -n 100 "$filename" | tail -n 20
 > > done
 > ~~~
 > {: .language-bash}
--- a/_episodes/05-loop.md
+++ b/_episodes/05-loop.md
@@ -389,10 +389,10 @@ from whatever file is being processed
 
 > ## Spaces in Names
 >
-> Whitespace is used to separate the elements on the list
-> that we are going to loop over. If on the list we have elements
-> with whitespace we need to quote those elements
-> and our variable when using it.
+> Spaces are used to separate the elements of the list
+> that we are going to loop over. If one of those elements
+> contains a space character, we need to surround it with
+> quotes, and do the same thing to our loop variable.
 > Suppose our data files are named:
 >
 > ~~~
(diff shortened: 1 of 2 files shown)


rgaiacs:
Hi @colinmorris, thanks for the pull request. I'm against use "spaces" because it is incorrect. If you look at [List of Unicode Characters of Category “Space Separator”](https://www.compart.com/en/unicode/category/Zs) you will see that users have many options to type.

I understand that "whitespace" can be something new for novices but it should be a huge cognitive load for learners.

gcapes:
I don't feel too strongly whether we use the term space or whitespace, but we should link to a definition of whitespace if we decide to use that.

However, using the term whitespace brings up the questions of
1. What is whitespace
2. Why/how would I put non-space whitespace characters into a file/directory name?

I've just tried to create a file which has a tab character in the name. I can't do it using the shell or a GUI file browser. It might be possible programmatically I suppose. Anyway, the point is it isn't easy to get these other whitespace characters into filenames where they would wreak havoc with the command line.

I've also just tried to execute  a command in the shell, separating arguments with tabs. It doesn't work because the tab is used for tab completion.

My main question though is In what way is using 'spaces' incorrect? Perhaps incomplete as an explanation, but don't we sometimes sacrifice completeness for clarity?

gdevenyi:
+1 for merge. I think for the purposes of shell-novice the complexities of non-printing characters can be ignored for clarity.

_episodes/05-loop.md
```diff
@@ -401,7 +401,7 @@ from whatever file is being processed
 > ~~~
 > {: .source}
 > 
-> We need to use
+> Then we would need to write
```
gcapes: I don't really like the use of 'write'. I think 'execute' or 'use' would be better.

colinmorris: Any particular reason you prefer those words? I kind of prefer "write" here because we're talking about some surface modifications to the source code we used earlier (adding some double quotes). To my mind, we're still *using* the same tools (a for loop, the `head` command piped to `tail`).

Some other possible wordings to sidestep the issue:

> Then we would need to add double quotes like so:

or

> The code to loop over these files would look like:

gcapes: Sorry for the lack of justification :) The reason I don't like it is the implication that writing (typing) a command is the same as running it.
Your first suggestion sounds good to me. Just something which conveys the message 'we'll need to modify the code thusly'

gcapes:
Good stuff @colinmorris

PR #916, opened 2019-02-11
Remove repeated material on deleting interactively

### LINKED ISSUE #914: Ep3: Redundancy between "Using rm Safely" exercise and "With Great Power Comes Great Responsibility"
This was mentioned by gcapes in #832. 

The "With Great Power Comes Great Responsibility" callout box introduces the `rm -i` flag as if it were a new thing, even though it was already introduced in the "Using rm Safely" exercise just a little earlier. 

I think we could consolidate these into one for the sake of brevity. They both cover roughly the same ground. Failing that, we should at least update the wording in "With Great Power" to say something like "once again, we can use the `-i` flag...".

gcapes:
Fix #914.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -527,10 +527,10 @@ ls: cannot access 'quotes.txt': No such file or directory
 >
 > > ## Solution
 > > ```
-> > $ rm: remove regular file 'thesis_backup/quotations.txt'?
+> > $ rm: remove regular file 'thesis_backup/quotations.txt'? y
 > > ```
 > > {: .language-bash}
-> > The `-i` flag will prompt before every removal.
+> > The `-i` flag will prompt before (every) removal (use <kbd>Y</kbd> to confirm deletion or <kbd>N</kbd> to keep the file).
 > > The Unix shell doesn't have a trash bin, so all the files removed will disappear forever.
 > > By using the `-i` flag, we have the chance to check that we are deleting only the files that we want to remove.
 > {: .solution}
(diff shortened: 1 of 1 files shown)


_episodes/03-create.md
```diff
@@ -527,10 +527,10 @@ ls: cannot access 'quotes.txt': No such file or directory
 >
 > > ## Solution
 > > ```
-> > $ rm: remove regular file 'thesis_backup/quotations.txt'?
+> > $ rm: remove regular file 'thesis_backup/quotations.txt'? y
 > > ```
 > > {: .language-bash}
-> > The `-i` flag will prompt before every removal.
+> > The `-i` flag will prompt before (every) removal (use <kbd>Y</kbd> to confirm deletion or <kbd>N</kbd> to keep the file).
```
gdevenyi: We're a bit too explicit here, with capital Y and N, small y and n work also.

_episodes/03-create.md
```diff
@@ -553,35 +553,15 @@ rm: cannot remove `thesis': Is a directory
...
-you can delete the directory without any confirmation prompts using the following command:
+recursive flag `-r`, and it will do so *without any confirmation prompts*:
 
 ~~~
 $ rm -r thesis
 ~~~
 {: .language-bash}
 
+Given that there is no way to retrieve files deleted using the shell,
+`rm -r` *should be used with great caution* (you might consider adding the interactive flag `rm -ri`).
```
gdevenyi: Broken formatting around "caution"?

gdevenyi: Also, we've squashed the options together here, I think we should keep them separate.

_episodes/03-create.md
gcapes: I've shown the lower-case 'y' in the code block, but I think the keyboard representation is correct. Keyboards show the upper case letter, so I think this is clear as it can be that it's 'press y' rather than 'press shift + y'.

gdevenyi: Ah, I see. I have missed that.

_episodes/03-create.md
gcapes: I can remove the italic if that's what you're suggesting?

gcapes: Why should the flags be kept separate? Combining them is introduced in episode 2.

gdevenyi: Ugh, I scanned the line and missed the start of the "*". Sorry.

I'm aware we introduced combining arguments, but I think we should stick to expressing them as individual ones for the lessons for clarity (this brings up whether there's other places where we are combining them...)

gcapes:
Is this all good to merge now?

PR #920, opened 2019-02-13
Acknowledge that error message might differ from lesson

### LINKED ISSUE #919: Ep3 - Copy with Multiple Filenames Example Clarification Needed
In the "Working With Files and Directories" episode (3), the example "Copy with Multiple Filenames" needs some clarification regarding the second part in that execution of:
**$ cp amino-acids.txt animals.txt morse.txt**

Does not always result in:
**cp: target ‘morse.txt’ is not a directory**

On my system (Mac), the error is not produced and instead I receive the output (double checked my working directory too):
**usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file target_file
       cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file ... target_directory**

I am fairly inexperienced with terminal, but others completing this lesson may also have the same problem.  From what I can tell based on my experimentation of "cp", it may work differently than when then episode was created/last updated.  Either way, it was strange to me as a novice user.

Anyone else seeing this result on their system?  It might be a simple modification of stating that the command will not work as expected or produce an error.

gcapes:
If learner has a different implementation of Bash e.g. BSD,
the error message won't always be the same.

Fix #919.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -618,7 +618,7 @@ Oftentimes one needs to copy or move several files at once. This can be done by 
 > > If given more than one file name followed by a directory name (i.e. the destination directory must 
 > > be the last argument), `cp` copies the files to the named directory.
 > >
-> > If given three file names, `cp` throws an error because it is expecting a directory
+> > If given three file names, `cp` throws an error such as the one below, because it is expecting a directory
 > > name as the last argument.
 > >
 > > ```


gdevenyi:
Hrm, I think this addition is a bit too vague to address the original issue.

Maybe the addition of "(other systems may vary slightly)"?

gcapes:
I would like to avoid cluttering the lesson with such comments, see #798. Perhaps that issue can be discussed further either before this is merged, or this issue can be merged as-is and reconsidered in the light on a decision on #798?

gdevenyi:
Understood, lets move the discussion there and hold this for now.

gcapes:
Ok, any further thoughts on this? Can it be merged, or should I reword it along the lines of
> an example of the sort of error message you might get is given below

gdevenyi:
+1

Issue #898, opened 2019-01-18
Episode 7: Add reminder which directory to run commands in


hollybik:
In the "Finding things" subsection, there could be a quick reminder to `cd data-shell/writing` before running the `find` commands. After the `grep` exercises, users are likely to be in another directory or subdirectory.

(Previously mentioned up in issue #854 as part of instructor checkout - closed and reopened each point as separate issue)

gdevenyi:
Thanks @hollybik for the suggestion! Newcomers and instructor checkout folks, this would be a nice quick contribution.

PR #925, opened 2019-02-25
implemented fix for issue 898

vwkoppejan:
Implemented a fix for issue #898 by adding a note to run the `find .` command in the "data-shell/writing" folder. Also added a reminder that the `pwd` can be used to print the path to the current working dir.

### DIFF
--- a/_episodes/07-find.md
+++ b/_episodes/07-find.md
@@ -423,6 +423,8 @@ $ find .
 ~~~
 {: .output}
 
+> If you're seeing a file structure that's different than the one shown above, chances are you are not executing the find command in the appropiate folder (`data-shell/writing`). Reminder: you can use the `pwd` command to show what you're current path is. {: .output}
+
 As always,
 the `.` on its own means the current working directory,
 which is where we want our search to start.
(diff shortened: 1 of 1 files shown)


_episodes/07-find.md
```diff
@@ -423,6 +423,8 @@ $ find .
 ~~~
 {: .output}
 
+> If you're seeing a file structure that's different than the one shown above, chances are you are not executing the find command in the appropiate folder (`data-shell/writing`). Reminder: you can use the `pwd` command to show what you're current path is. {: .output}
```
gcapes: Please correct the wording below 
> ... you're seeing a file structure that's different **to**

gcapes: > Reminder: you can use the `pwd` command to show what you're current path is.
You're -> your

gcapes: > {: .output}
I can't see that this markup is needed? There's no code block it would be associated with.

gcapes:
Thanks for the updates. Having re-read #898 I think the suggestion was only for a reminder which directory you should be in before running the example, rather than how to diagnose problems after running the example. Sorry I didn't re-read the issue before commenting.

I think this addition is more the help someone might give following a red sticky-note in a workshop?

A separate PR would be welcome, which adds a reminder which directory you should be in before following the example.

Thanks again.

PR #930, opened 2019-03-08
Replaced the "taster" regex example with one that I think is a

gjanee:
Replaced the "taster" regex example with one that I think is a little more realistic.
---

### DIFF
--- a/_episodes/07-find.md
+++ b/_episodes/07-find.md
@@ -258,27 +258,30 @@ Miscellaneous:
 > these is **regular expressions**, which
 > is what the "re" in "grep" stands for.) Regular expressions are both complex
 > and powerful; if you want to do complex searches, please look at the lesson
-> on [our website](http://v4.software-carpentry.org/regexp/index.html). As a taster, we can
-> find lines that have an 'o' in the second position like this:
+> on [our website](http://v4.software-carpentry.org/regexp/index.html).
+> As a taster, suppose you wanted to find a phone number in a
+> document.  Knowing just the structure of the phone number, without
+> knowing the phone number itself, you could search for it like this:
 >
 > ~~~
-> $ grep -E '^.o' haiku.txt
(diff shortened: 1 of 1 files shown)


gcapes:
Thanks for the PR!

I would support the move to a better motivated example like this. However I have two reservations about this suggestion:
1. `document.txt` doesn't exist in `data-shell` so the example isn't runnable. We could add such a file but we're trying not avoid further bloating the lesson.
2. The number format is rather US-centric. Perhaps a URL would be a more inclusive example, but again it would require adding to the existing data set.

Thoughts from @swcarpentry/shell-maintainers?

gdevenyi:
I have the same concerns as @gcapes. I think we want runnable examples everywhere. Maybe an ISO date?

gcapes:
Yep, that would work from the data-shell directory:
```
$ grep -rE '[0-9]{4}-[0-9]{2}-[0-9]{2}' *

data/animal-counts/animals.txt:2012-11-05,deer,5
data/animal-counts/animals.txt:2012-11-05,rabbit,22
data/animal-counts/animals.txt:2012-11-05,raccoon,7
data/animal-counts/animals.txt:2012-11-06,rabbit,19
data/animal-counts/animals.txt:2012-11-06,deer,2
data/animal-counts/animals.txt:2012-11-06,fox,4
data/animal-counts/animals.txt:2012-11-07,rabbit,16
data/animal-counts/animals.txt:2012-11-07,bear,1
data/animals.txt:2012-11-05,deer
data/animals.txt:2012-11-05,rabbit
data/animals.txt:2012-11-05,raccoon
data/animals.txt:2012-11-06,rabbit
data/animals.txt:2012-11-06,deer
data/animals.txt:2012-11-06,fox
data/animals.txt:2012-11-07,rabbit
data/animals.txt:2012-11-07,bear
```

deppen8:
Closing this because this issue is being addressed in #1134

PR #935, opened 2019-03-14
Use better motivated example for introducing loops

gcapes:
- Use example which can't be easily approximated with wildcards.
- Start with a definition of what a loop is
- First example is better motivated, but still simple
- Postpone the more complicated example until later in the episode

Fix #674.

### DIFF
--- a/_episodes/05-loop.md
+++ b/_episodes/05-loop.md
@@ -39,7 +39,7 @@ We'll use a loop to solve this problem, but first let's look at the general form
 ```
 for thing in list_of_things
 do
-    operation_using $thing
+    operation_using $thing    # Indentation within the loop is not required, but aids legibility
 done
 ```
 {: .language-bash}
@@ -49,7 +49,7 @@ and we can apply this to our example like this:
 ```
 $ for filename in basilisk.dat unicorn.dat
 > do
->    head -n 2 $filename | tail -1	# Indentation within the loop is not required, but aids legibility
+>    head -n 2 $filename | tail -1
 > done
 ```
 {: .language-bash}
@@ -67,11 +67,6 @@ COMMON NAME: unicorn
 > typing in our loop. The second prompt, `>`, is different to remind
 > us that we haven't finished typing a complete command yet. A semicolon, `;`,
 > can be used to separate two commands written on a single line.
-{: .callout}
-
-> ## Indentation of code within a for loop
-> Note that it is common practice to indent the line(s) of code within a for loop.
-> The only purpose is to make the code easier to read -- it is not required for the loop to run.
 {: .callout}
 
 When the shell sees the keyword `for`,
--- a/_episodes/05-loop.md
+++ b/_episodes/05-loop.md
@@ -411,8 +411,8 @@ from whatever file is being processed
 > {: . output}
 {: .callout}
 
-We would like to modify these files, but also save a version of the original files, naming the copies
-`original-basilisk.dat` and `original-unicorn.dat`.
+We would like to modify each of the files in `data-shell/creatures`, but also save a version
+of the original files, naming the copies `original-basilisk.dat` and `original-unicorn.dat`.
 We can't use:
 
 ~~~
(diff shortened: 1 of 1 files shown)


gcapes:
bump
@swcarpentry/shell-maintainers

_episodes/05-loop.md
```diff
@@ -22,63 +22,45 @@ keypoints:
...
-Instead, we can use a **loop**
-to do some operation once for each thing in a list.
-Here's a simple example that displays the first three lines of each file in turn:
-
-~~~
+```
 $ for filename in basilisk.dat unicorn.dat
 > do
->    head -n 3 $filename	# Indentation within the loop aids legibility
+>    head -n 2 $filename | tail -1	# Indentation within the loop is not required, but aids legibility
```
colinmorris: Maybe this comment about indentation should be moved up to the new 'general form' code?

_episodes/05-loop.md
```diff
@@ -22,63 +22,45 @@ keypoints:
...
 > done
-~~~
+```
 {: .language-bash}
 
-~~~
+```
```
colinmorris: Just to be clear, maybe have some text between these blocks saying something like "When run, this produces the following output:"?

_episodes/05-loop.md
colinmorris: Also, I wonder if we really need the "Indentation of code within a for loop" callout in addition to this comment.

_episodes/05-loop.md
```diff
@@ -434,9 +416,35 @@ from whatever file is being processed
 > {: . output}
 {: .callout}
 
-Going back to our original file copying problem,
-we can solve it using this loop:
+We would like to modify these files, but also save a version of the original files, naming the copies
```
colinmorris: At this point, it's been a while since the basilisk and unicorn files have been mentioned (if you count the 'Spaces in Names' callout - which visually blends in very closely with the main text), so "these files" is a bit of a non-sequitur. Suggested alternative wording:

> Suppose that we plan to modify each of the files in the `data-shell/creatures` directory, but we would first like to save copies of the originals, using the filenames `original-basilisk.dat` and `original-unicorn.dat`.

_episodes/05-loop.md
```diff
@@ -465,7 +473,7 @@ cp unicorn.dat original-unicorn.dat
 Since the `cp` command does not normally produce any output, it's hard to check 
 that the loop is doing the correct thing. By prefixing the command with `echo`
```
colinmorris: This is not directly related to the goal of this PR, so feel free to leave it alone, but I'm not a fan of how the echo stuff is worded. It makes it seem like "prefixing a command with `echo`" is some piece of magic (rather than something that can be understood in terms of the semantics of `echo` that were taught in episode 4). I would rather word it along the lines of:

> Recall the `echo` command we learned earlier, which prints its arguments to stdout. If we prefix our command with `echo`, everything that follows (including "cp") will be treated as a string and printed to stdout. Our loop will print each command as it *would* have been executed (without actually executing it). The following diagram shows what happens when the modified loop is executed, and demonstrates how the judicious use of `echo` is a good debugging technique.

Maybe that's overly verbose though. (I also question whether the diagram helps or hinders understanding, but I'll refrain from going further into the weeds)

_episodes/05-loop.md
gcapes: Does the code block colour not already convey this?

_episodes/05-loop.md
gcapes: Well spotted!

gcapes:
Thanks @colinmorris 

I think I've addressed all the issues raised.

_episodes/05-loop.md
colinmorris: Looking back at other episodes, I'm realizing how common this pattern is (of putting output in a block immediately under the command), so yeah, I guess this shouldn't be hard to grok.

PR #936, opened 2019-03-18
Use -n flag consistently with head and tail

### LINKED ISSUE #853: Inconsistent usage of head/tail in Episode 4 exercise "Appending Data"
There was already an issue open about consistency. Thought the example works, it is important to be consistent when dispensing a lesson.

This concerns Episode 4 (Appending Data example, the solution part)
            Option 3 is correct. For option 1 to be correct we would only run
            the head command. For option 2 to be correct we would only run the
            tail command. For option 4 to be correct we would have to pipe the
            output of head into tail -2 by doing  head -3 animals.txt | tail -2>> animalsUpd.txt . 
            Though head -3 and tail -2 will return same output as head -n 2 and tail -n 2 it is 
            inconsistent with the command option argument synopsis that was predefined

gcapes:
Fix #853.

### DIFF
--- a/_episodes/04-pipefilter.md
+++ b/_episodes/04-pipefilter.md
@@ -322,7 +322,7 @@ the output of `head` must be the file with the fewest lines.
 > > Option 3 is correct. 
 > > For option 1 to be correct we would only run the `head` command.
 > > For option 2 to be correct we would only run the `tail` command.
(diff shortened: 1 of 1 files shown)


drj11:
I particularly like this because POSIX requires the `-n` option for portability. http://pubs.opengroup.org/onlinepubs/9699919799/utilities/head.html

gdevenyi:
Great!

PR #939, opened 2019-03-25
Add callout for case sensitivity.

levlafayette:
Takes less than 30 seconds to explain, very handy when encountered in other environments which are case-insensitive or case-preserving. Appropriate to include at point of directory/file creation and in discussion about good file and directory names.

---

Please keep in mind that lesson maintainers are volunteers and it may be some time before they can respond to your contribution. Although not all contributions can be incorporated into the lesson materials, we appreciate your time and effort to improve the curriculum.  If you have any questions about the lesson maintenance process or would like to volunteer your time as a contribution reviewer, please contact Kate Hertweck (k8hertweck@gmail.com).  

---

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -83,6 +83,14 @@ creatures/  data/  molecules/  north-pacific-gyre/  notes.txt  pizza.cfg  solar.
 > the files and directories themselves are the same.
 {: .callout}
 
+> ## Case Sensitive
+> Is the directory `thesis/` the same as `THESIS/` ?
+> It depends on the operating system and file system. In the Unix shell environment files
+> and directories are case-sensitive. Which means that they are different directories 
+> and `thesis/`, and `THESIS/` etc can be present the same time on the same directory level 
+> with different content.
+{: .callout}
+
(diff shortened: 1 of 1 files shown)


gcapes:
Hi @levlafayette and thanks for the PR!

Whether or not the shell is case sensitive depends on the file system, so this is going to vary between learners. Episode 1 does point out that capitalisation matters for commands, and this is true regardless of the file system.

So I'm going to close this PR, but invite the other maintainers to give their thoughts.
@swcarpentry/shell-maintainers 
I'm unsure whether possible case-sensitivity of files and directories is something that we need to cover and if so how. Those following the course on macOS or Windows wouldn't be able to test this property, so perhaps this can left as a discussion point if it comes up via a red sticky-note?

gdevenyi:
While this is incredibly important in general I think this is another case of "simplify to reduce the cognitive load of learners". We don't have control of their environments so we can't guarantee sensitivity, and explaining why this is the case is too complicated for this introduction material.

levlafayette:
OK, but be aware that people can over-write their files and directories without this knowledge. Oh quel dommage!

PR #940, opened 2019-03-26
Use descriptive file name in exercise

### LINKED ISSUE #937: Ep4 confusing file name `animalsUpd.txt`
http://swcarpentry.github.io/shell-novice/04-pipefilter/index.html#appending-data

What is the significance of this file name? It doesn't mean anything to me, and I guess might also be a source of confusion to learners.

Initial discussion on #853 suggested something to do with garbage collection. I would suggest `animals-subset.txt` or `animals-middle.txt` or something like that, so the file name better describes the content.

gcapes:
Fix #937.


gdevenyi:
Great, odd name to begin with..

PR #941, opened 2019-03-28
Add directory diagram for wildcards exercise

### LINKED ISSUE #793: Ep3: Filesystem clarification for "More on Wildcards" exercise
http://swcarpentry.github.io/shell-novice/03-create/index.html#more-on-wildcards

I think the exercise titled "More on Wildcards" could benefit from showing a filesystem diagram similar to the one found in "Relative Path Resolution" in Unix Shell: Navigating Files and Directories (Image found here:https://github.com/swcarpentry/shell-novice/blob/gh-pages/fig/filesystem-challenge.svg).

The exercise shows files within a current working directory and asks students how to use wildcards to cp files to other directories.  However, I think students might benefit from being reminded how the filesystem hierarchy looks.  This might, therefore improve their reasoning for copying files in their explicitly stated destination directories.

Another improvement could be to show the filesystem changes after the files are copied to a new directory for each scenario in the "Solutions" section.

gcapes:
Fix #793.

This does take quite a bit of vertical space but should help clarify the purpose of the exercise.

I've created the initial directories so that this remains focused on copying and wildcards, and to simplify the diagram I've created the destination directories as subdirectories.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -671,18 +671,25 @@ Oftentimes one needs to copy or move several files at once. This can be done by 
 > the datasets:
 >
 > ~~~
-> 2015-10-23-calibration.txt
-> 2015-10-23-dataset1.txt
-> 2015-10-23-dataset2.txt
-> 2015-10-23-dataset_overview.txt
-> 2015-10-26-calibration.txt
-> 2015-10-26-dataset1.txt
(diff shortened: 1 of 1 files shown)


colinmorris:
Looks good. The vertical space issue is a bit bothersome, but at least this is in a block of exercises at the end of the episode, so it doesn't break up the main text.

PR #944, opened 2019-03-29
? wildcard explained in slightly different wording

steltenpower:
Trying to make it a little less likely people misunderstand


gdevenyi:
Great change!

PR #960, opened 2019-04-16
Syntax error 04 pipefilter

rantahar:
The solution to the first part of the Wildcard Expressions exercise.
The fix creates a code block in the list, which in my mind looks alright.


gcapes:
Thanks for this. Markdown is a bit tricky here, so I've made a small change to your PR.

PR #962, opened 2019-04-18
Render letter keys as lower-case

mikerenfro:
Adjusting case of hotkeys used in man (technically less).

It could be confusing for new users to read a capital B in the text when we really mean a lowercase b, and later use the notation of Shift-B to mean an actual capital B.

There are other places this could be adjusted (q versus Q or ctrl-o versus ctrl-O, for example), but I've avoided changing anything that works, even if it's not the default that most people would use.


gcapes:
Thanks for the PR.

We use upper case letters in this scenario which match what the user will see on their keyboard. See the style guide here:
http://carpentries.github.io/lesson-example/06-style-guide/index.html#keyboard-key

PR #966, opened 2019-04-26
Replace use of 'flag' with 'option'

### LINKED ISSUE #938: Prefer `option` over `flag`
The Shell lessons (example, http://swcarpentry.github.io/shell-novice/01-intro/index.html) has the text "some flags (also called options or switches)" and then uses _flag_ throughout.

Not only is _flag_ somewhat historical, [POSIX documents them](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02) as «known as "options" (or, historically, "flags")» (and then is not completely consistent in using "options", but mostly so). [had intended to add here: , but a flag can only be present or absent, the common meaning of flag makes nonsense of "flags having arguments"]

There is a lot of varying terminology in this area (flag/option/switch/parameter/etc), but I believe we are in a position to promote terminology that is both standard and existing.

gcapes:
Fix #938.

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -9,8 +9,8 @@ objectives:
 - "Explain when and why command-line interfaces should be used instead of graphical interfaces."
 keypoints:
 - "Explain the steps in the shell's read-run-print cycle."
-- "Most commands take flags (options) which begin with a `-`."
-- "Identify the actual command, flags, and filenames in a command-line call."
+- "Most commands take options (flags) which begin with a `-`."
+- "Identify the actual command, options, and filenames in a command-line call."
 - "Explain the steps in the shell's read-run-print cycle."
 - "Demonstrate the use of tab completion and explain its advantages."
 keypoints:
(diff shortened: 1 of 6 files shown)


gdevenyi:
Extra file committed?

gcapes:
Sorry! Not sure how those were created. Fixed now.

_episodes/01-intro.md
```diff
@@ -89,14 +89,14 @@ when typing commands, either from these lessons or from other sources,
...
 `ls -F /` in the second line of the example,
 typically has the following structure: a **command**,
-some **flags** (also called **options** or **switches**) and an **argument**.
+some **options** (also called **switches** or **flags**) and an **argument**.
 Flags start with a single dash (`-`) or two dashes (`--`), and change the behaviour of a command.
 Arguments tell the command what to operate on (e.g. files and directories).
-Sometimes flags and arguments are referred to as parameters.
-A command can be called with more than one flag and more than one argument: but a
-command doesn't always require an argument or a flag.
+Sometimes options and arguments are referred to as parameters.
```
gdevenyi: Maybe:
Sometimes options and arguments are together referred to as **parameters**.

gdevenyi:
Looks good. One tiny suggestion of clarification/formatting.

gdevenyi:
Wonderful!

Issue #927, opened 2019-02-26
POSIX compliance


OH-AU:
In _extras/guide.md you suggest "Stay within POSIX-compliant commands" and helpfully give an example, however no assistance/referral is provided to check compliance. Having used various UNIX and UNIX-like operating systems, I couldn't tell you if a command I regularly use is POSIX complaint or not. The closest related issue I could find was #708 which asks for assistance in auditing. Perhaps the guide could be updated with a reference to:
1. https://pubs.opengroup.org/onlinepubs/9699919799/idx/utilities.html
Unfortunately, this doesn't cover bash itself. So also include a reference to:
2. https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_01

Perhaps something like; You can check for POSIX compliance for the shell [here](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_01) and command line utilities [here](https://pubs.opengroup.org/onlinepubs/9699919799/idx/utilities.html)
Thanks.

colinmorris:
Great point. I agree that, as it currently stands, this piece of advice is not very helpful. Even with the additional links you suggest, I think it would still be a little unclear (i.e. it's not clear the spirit that's motivating this rule, or what specific actions instructors are expected to take in response to it).

This is somewhat related to our ongoing discussion at #798.

I wonder about deleting this advice and replacing it with a section with some more informal, general advice about cross-platform issues and consistency, basically stuff like:
- Where possible, try to use commands that work (and do the same thing) on all platforms. We've aimed for this in the examples given in the lesson.
- Try to avoid teaching features that are specific to a particular operating system or shell dialect (unless *all* learners are using it?). For example, ____
- You may still encounter some scenarios where a command fails or gives different results for certain learners according to their system's configuration. It's worth explaining to learners that there are some slight differences between Bash shells on different systems, but try not to dwell on them, or get bogged into technical explanations for why these differences exist.
- [Some practical advice for powering through these hiccups during a workshop? e.g. if you have a minority of learners using system Y, you may want to seat them together so they can help each other out?]

gdevenyi:
+1 for enhancing that in the instructor's guide

We still have the issue of what to do with the lessons themselves though...

OH-AU:
[...]

PR #967, opened 2019-04-30
Mention alternate method for controlling lines of output from head

frenchwr:
A more concise method for controlling the lines of output from head/tail!


gdevenyi:
Hi, thanks for your suggested modification, however this is a non-posix option for head which we're trying to avoid suggesting, as its not guaranteed to work on all unicies, you can find more discussion at #798 and #927

PR #968, opened 2019-05-02
Correction to initial loop example

pgmccann:
Output indicated in lesson was not consistent with the previous description of the intended purpose of the loop, and the tail command had the wrong option. This corrects both.


gdevenyi:
Good catch!

Issue #836, opened 2018-08-21
Move "Why Isn't it Doing Anything?" to an earlier episode


Bioscibrarian:
The note "Why Isn't it Doing Anything?" is introduced in Episode 6.
It describes how a command will wait for input if the user forgets to add a filename for a command to take action on.

The command used for an example is part of a piped setof commands. It is **wc -l**, without a filename given, so **wc** is waiting for input.

I think that shell novices might make this mistake earlier in the class--I know I did when I first learned bash commands. 

It is possible to introduce "Why Isn't it Doing Anything?" as early as **Episode 4** (Pipes and Filters):

- The text in the lesson is:
If we run wc -l instead of just wc, the output shows only the number of lines per file:
$ wc -l *.pdb

- Could we introduce "Why Isn't it Doing Anything?" here, and describe what would happen if the user just typed "wc -l" (without *.pdb)?

- Then it might be useful to introduce Ctrl-Z or some other command-suspending intervention. I know this might be too much to introduce, but I have been a helper in sessions when at least one or two people have gotten "stuck" in a bash command that was waiting for input. Students need to know at some point how to suspend a command.

This is part of my Instructor Training Checkout, but it's also something I would like to see introduced a bit earlier in the lessons. -Thanks!
Pamela Shaw, Northwestern University

jlevell:
I totally agree I think this would be much better if introduced earlier.  It's part of what I consider basic usage of the Unix Shell.  I am not advocating that we teach every Unix shortcut but I think it would be very beneficial to touch on at least a few of them like ctrl-z.

I'm thinking it might be useful to add a list of shortcuts as supplemental material.   Consider something like this link but as a hand-out.  https://www.computerhope.com/ushort.htm.

Much like Bioscibrarian I am using this as part of my Instructor Training Checkout but I am also interested in making this better for our students.

gcapes:
A pull request addressing this issue would be welcome :)

gcapes:
Fixed by #976 and #969

PR #969, opened 2019-05-08
Added callout for non-responsive commands

cmentzel:
Responding to issue #836  - callout is "Why Isn't It Doing Anything?" based on the suggestion by @Bioscibrarian and the material currently in episode6

Also part of my instructor checkout step!

### DIFF
--- a/_episodes/04-pipefilter.md
+++ b/_episodes/04-pipefilter.md
@@ -87,6 +87,27 @@ $ wc -l *.pdb
 ~~~
 {: .output}
 
+> ## Why Isn't It Doing Anything?
+>
+> What happens if a command is supposed to process a file, but we
+> don't give it a filename? For example, what if we type:
+>
+> ~~~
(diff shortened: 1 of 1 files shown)


gcapes:
Hi @cmentzel and thanks for the PR!

I like the place where this is introduced. I would suggest however that we remove the callout from episode 6 rather than duplicating what we have here.

Also, a couple of formatting niggles (see the [style guide](http://carpentries.github.io/lesson-example/06-style-guide/index.html#keyboard-key):
- capitalisation of letter keys
- rendering of keyboard keys

cmentzel:
Also, removed the similar section in Episode 6 as part of PR #976

gcapes:
Thanks! I'll merge these PRs, but a point to bear in mind is that it would have been easier if you made all the requested changes in this PR instead of creating a new one for a subset of the changes.

Issue #644, opened 2017-10-25
ls -F needs more explanation


weisi-chen:
ls -F (more explanation on why use -F is needed)

gcapes:
I'd say it's fine as it is. Have you seen this (from episode 2)?
> ls prints the names of the files and directories in the current directory in alphabetical order, arranged neatly into columns. We can make its output more comprehensible by using the flag -F (also known as a switch or an option) , which tells ls to add a trailing / to the names of directories:

colinmorris:
One thing that might make this a little confusing is that, in the example shown in the lesson, the pwd contains *only* directories. 

```
Applications/ Documents/    Library/      Music/        Public/
Desktop/      Downloads/    Movies/       Pictures/
```

If there were one or two files in there, it might make the usefulness of `-F` clearer (it lets you distinguish between directories and files at a glance):

```
Applications/ Documents/    Library/      mongoose.jpg    Music/    Public/ 
Desktop/      Downloads/    Movies/       Pictures/       Todo
```

Lyndon18:
It may also be useful to introduce flags formally, before adding flags into the examples.

loz-hurst:
Should it be clarified that '-F' is a general 'classify' flag - i.e. it does not *just* add '/' to directories (which is the implication of the lesson text) but also adds characters to other 'special' files (e.g. '*' to executable, '@' to symlinks, '=' to sockets, '|' to pipes etc.).

I'm not suggesting introducing all these concepts, just describing '-F' as an option that classifies files rather than 'adds / to directores' which could be mis-leading.

raynamharris:
I agree. This came up during my maintainer on-boarding class. I learned that it is really a quite useful tool for folks who don't use color coding. 

Also, I find the `ls -F` option quite useful option when combined with `-a` because it makes it more obvious that `./` and `../` refer to directories but agree with the folks above that it is unclear when the said directory contains only subdirectories.

gshamov:
These days, colour terminals show files, directories and links in colour. Isn't ls -F often redundant? (I have seen it first time in the lesson myself.) Would it be more natural to introduce options to ls by using options that are used most often (ls -la, ls -lrt)?

loz-hurst:
> These days, colour terminals show files, directories and links in colour. Many Linux distros actually alias ls to alias ls='ls --color=auto'

[...]

PR #972, opened 2019-05-17
Explain general command syntax after simple example

### LINKED ISSUE #846: Episode 1: `ls -F` never fully explained, and introduced before `ls`
Instead of jumping straight into "ls -F" command, it'd be good to demonstrate with "ls" first and then go to "ls -F".

gcapes:
Fix #846.

- Explain ls -F more completely
- Explain REPL and commands in episode 1
- Move options and arguments explanation to episode 2
- Show example where captialisation matters, regardless of OS (see #798)

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -82,8 +82,7 @@ Most importantly:
 when typing commands, either from these lessons or from other sources,
 *do not type the prompt*, only the commands that follow it.
 
-So let's try our first command, which will list the contents of the current directory
-(assuming that our current directory contains some files or subdirectories):
+So let's try our first command, which will list the contents of the current directory:
 
 ~~~
 $ ls
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -77,14 +77,10 @@ $
 {: .language-bash}
 
 The shell typically uses `$ ` as the prompt, but may use a different symbol.
+In the examples for this lesson, we'll show the prompt as `$ `.
 Most importantly:
 when typing commands, either from these lessons or from other sources,
 *do not type the prompt*, only the commands that follow it.
-
-In the examples for this lesson, we'll show the prompt as `$ `. You can make your
-prompt look the same by executing the command `PS1='$ '`. But you can also leave
-your prompt as it is - often the prompt includes useful information about who and where
-you are.
 
 So let's try our first command, which will list the contents of the current directory
 (assuming that our current directory contains some files or subdirectories):
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -45,7 +45,7 @@ line by line. Using the tradition GUI approach of clicks will take several hours
 
 This is where we take advange of the shell - a **command-line interface**
 to make such repetitive tasks automatic and fast. It can take a single instruction and
-repeat it as it is or with some modification as many times as we want. The task in
+repeat it as is or with some modification as many times as we want. The task in
 the example above can be accomplished in a few minutes at most.
 
 The heart of a command-line interface is a **read-evaluate-print loop** (REPL). It is called
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -40,12 +40,12 @@ a new folder/directory) with the convenience of a few mouse clicks. This way of 
 with a computer is intuitive and very easy to learn. But this way of giving instructions to
 a computer scales very poorly if we are to give a large stream of instructions even if they
 are similar or identical. For example if we have to copy the third line of each of a thousand
-text files stored in thousand different folders/directories and paste it into a single file
+text files stored in thousand different directories and paste it into a single file
 line by line. Using the tradition GUI approach of clicks will take several hours to do this.
 
 This is where we take advange of the shell - a **command-line interface**
 to make such repetitive tasks automatic and fast. It can take a single instruction and
-repeat it over as it is or with some modification as many times as we want. The task in
+repeat it as it is or with some modification as many times as we want. The task in
 the example above can be accomplished in a few minutes at most.
 
 The heart of a command-line interface is a **read-evaluate-print loop** (REPL). It is called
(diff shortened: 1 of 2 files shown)


_episodes/01-intro.md
```diff
@@ -59,98 +59,61 @@ loops back and waits for you to enter another command.
...
-doesn't exist. Also, capitalization matters: `LS` is different from `ls`.
-
-Next we see the output that our command produced. In this case it is a listing
-of files and folders in a location called `/` - we'll cover what all these mean
-later today. Those using a macOS might recognize the output in this example.
-
-Finally, the shell again prints the prompt and waits for you to type the next
-command.
-
 In the examples for this lesson, we'll show the prompt as `$ `. You can make your
```
colinmorris: Minor suggestion: maybe move this into the paragraph above, immediately before the last sentence ("Most importantly:..."). Alternatively, I feel like removing this part entirely might be an improvement. It doesn't seem like there's much upside to changing their prompt to `$ `, but there is a potential big downside if they typo the command somehow and mess up their prompt (this happened to one learner the last time I taught this lesson).

_episodes/01-intro.md
```diff
@@ -59,98 +59,61 @@ loops back and waits for you to enter another command.
...
 prompt look the same by executing the command `PS1='$ '`. But you can also leave
 your prompt as it is - often the prompt includes useful information about who and where
 you are.
 
-Open a shell window and try executing `ls -F /` for yourself (don't forget that spaces
-and capitalization are important!). You can change the prompt too, if you like.
-
-### How does the shell know what `ls` and its options mean?
+So let's try our first command, which will list the contents of the current directory
+(assuming that our current directory contains some files or subdirectories):
```
colinmorris: Nitpick: is the parenthetical necessary? I mean, if I run `ls` on an empty directory, does it not output the contents of that directory?

_episodes/02-filedir.md
```diff
@@ -157,12 +158,44 @@ Desktop/      Downloads/    Movies/       Pictures/
 
 Here,
 we can see that our home directory contains mostly **sub-directories**.
-Any names in your output that don't have trailing slashes,
+Any names in your output that don't have a classification symbol,
```
colinmorris: For lines 139-161, I somewhat prefer the old wording (modulo the fixes you made to the repetition of "option", and the no-longer-applicable "earlier example" callback), because:

* A typical learner is unlikely to be familiar with executable files (in the unix sense), and even less likely to be familiar with links. 
* Calling it the "classify option" seems a little confusing, because it's not obvious that `-F` would be an abbreviation for "classify". A reader might even think that "classify option" is some generic piece of shell jargon? You could try to introduce the `--classify` long form, though I don't know if that would be helpful at this point.

_episodes/01-intro.md
```diff
@@ -59,98 +59,61 @@ loops back and waits for you to enter another command.
...
-program that creates a new folder/directory. The simple programs which are used to perform
+Those programs can be as complicated as climate modeling software and as simple as a
+program that creates a new directory. The simple programs which are used to perform
 stand alone tasks are usually refered to as commands.
 The most popular Unix shell is Bash, (the Bourne Again SHell --- so-called because
 it's derived from a shell written by Stephen Bourne).
 Bash is the default shell on most modern implementations of Unix
 and in most packages that provide Unix-like tools for Windows.
 
-### What does it look like?
```
colinmorris: Not a big deal either way, but I kind of liked this section heading

colinmorris:
This is great stuff! Really like the way you've reorganized and reworded this content. And happy to see the "How does the shell know what `ls` and its options mean?" section go the way of the dodo. I think it was overly technical and low on practical value.

Left a few comments with nitpicks or suggestions - but feel free to take or leave them.

gdevenyi:
Love the addition of the structure of a command!

_episodes/01-intro.md
gcapes: Great suggestions!

_episodes/01-intro.md
gcapes: Yep, fair point!

_episodes/01-intro.md
gcapes: So did I, but given we're no longer jumping in at the deep end, the heading seems a bit OTT:
## What does it look like? 
A prompt.

_episodes/02-filedir.md
gcapes: Hmm.
I was trying to address [this comment](https://github.com/swcarpentry/shell-novice/issues/644#issuecomment-374961770). 

How about this?
> We can make its output more comprehensible by using the `-F` **option** 
(also known as a **switch** or a **flag**) ,

_episodes/01-intro.md
```diff
@@ -40,12 +40,12 @@ a new folder/directory) with the convenience of a few mouse clicks. This way of
...
 a computer scales very poorly if we are to give a large stream of instructions even if they
 are similar or identical. For example if we have to copy the third line of each of a thousand
-text files stored in thousand different folders/directories and paste it into a single file
+text files stored in thousand different directories and paste it into a single file
 line by line. Using the tradition GUI approach of clicks will take several hours to do this.
 
 This is where we take advange of the shell - a **command-line interface**
 to make such repetitive tasks automatic and fast. It can take a single instruction and
-repeat it over as it is or with some modification as many times as we want. The task in
+repeat it as it is or with some modification as many times as we want. The task in
```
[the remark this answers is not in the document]
gdevenyi: Sorry late on this, another instance of not "finishing" the review.

_episodes/01-intro.md
```diff
@@ -59,98 +59,56 @@ loops back and waits for you to enter another command.
...
+> will print an error message such as:
 >
 > ~~~
-> $ ls-F
+> $ ks
 > ~~~
 > {: .language-bash}
 > ~~~
-> -bash: ls-F: command not found
+> ks: command not found
```
gdevenyi: Depending on the operating system, the command-not-found bash hook might suggest a program to install to supply ks....

PR #980, opened 2019-06-05
Remove exercise on `uniq`

### LINKED ISSUE #876: Pipes and Filters hint on question about uniq command

---

In the  Pipes and Filters lesson, elaborate on the Hint to the question: Why Does uniq Only Remove Adjacent Duplicates?

There is a hint to think about very large data sets, but that's not enough for me to come up with the answer.

---

gcapes:
Fix #876.

### DIFF
--- a/_episodes/04-pipefilter.md
+++ b/_episodes/04-pipefilter.md
@@ -97,13 +97,13 @@ $ wc -l *.pdb
 > ~~~
 > {: .language-bash}
 >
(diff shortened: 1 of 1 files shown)


gcapes:
We do however need to consider later examples which use `uniq`
- Pipe construction
- Which pipe

The first of these would need an introduction to the command.

gcapes:
bump!

gdevenyi:
Looks fine. Thanks!.

PR #983, opened 2019-06-06
Remove excess of technical detail

### LINKED ISSUE #956: Excess of technical detail in episode 4 on processes
Quote from [episode 4 on pipes and filters](http://swcarpentry.github.io/shell-novice/04-pipefilter/index.html) (emphasis mine):

> The vertical bar, `|`, between the two commands is called a pipe. It tells the shell that we want to use the output of the command on the left as the input to the command on the right. **The computer might create a temporary file if it needs to, or copy data from one program to the other in memory, or something else entirely; we don’t have to know or care.**

I agree wholeheartedly! But later in the same lesson we do in fact get into some OS-level behind-the-scenes details that seem unnecessary. In particular, the stuff on processes, beginning with:

> Here’s what actually happens behind the scenes when we create a pipe. When a computer runs a program — any program — it creates a process in memory to hold the program’s software and its current state.

The term 'process' isn't used anywhere else in the lesson. For the purposes of this explanation, I feel like we could just continue to use either the term 'command' or 'program' (i.e. "the program's standard input"), since those terms are used throughout the lesson.

Also, I think this whole paragraph on the shell's implementation could be cut:

> The shell is actually just another program. Under normal circumstances, whatever we type on the keyboard is sent to the shell on its standard input, and whatever it produces on standard output is displayed on our screen. When we tell the shell to run a program, it creates a new process and temporarily sends whatever we type on our keyboard to that process’s standard input, and whatever the process sends to standard output to the screen.

I don't really need to know that the shell is itself a program with a stdout and a stdin, rather than, say, a book of spells animated by dark magic. Do I?

gcapes:
Fix #956.

Minor rewording and reordering of remaining text.

### DIFF
--- a/_episodes/04-pipefilter.md
+++ b/_episodes/04-pipefilter.md
@@ -99,7 +99,7 @@ $ wc -l *.pdb
 >
 > but don't type `*.pdb` (or anything else) after the command?
 > Since it doesn't have any filenames, `wc` assumes it is supposed to
-> process standard input, so it just sits there and waits for us to give
+> process input given at the command prompt, so it just sits there and waits for us to give
 > it some data interactively. From the outside, though, all we see is it
 > sitting there: the command doesn't appear to do anything.
 >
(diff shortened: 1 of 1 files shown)


gdevenyi:
This is probably sensible to do. The technical details aren't needed for into learners.

gcapes:
bump

gdevenyi:
hihi, I guess I was vague in my response, this looks fine. I'm okay with you merging your stuff directly too :)

gcapes:
>I'm okay with you merging your stuff directly too :) 

Thanks, I do this on lessons where I'm the sole maintainer, but for this lesson I'd prefer to avoid merging my own PRs - always good to get a second pair of eyes on things.

PR #985, opened 2019-06-06
Minor edits to episode 4. Closes #979.

### LINKED ISSUE #979: Episode 4, "Which Pipe?" options not previously introduced
In the "Which pipe?" exercise, several of the options include `uniq -c` and `grep` as candidate solutions. This is the first occurrence in the lesson of the `grep` command at all, and while `uniq` has been introduced, the `-c` option has not yet been introduced for this command.

This caused some disorientation in learners.

I think `uniq -c` is fine if the `-c` option is explained earlier, but that this is too early for `grep`.

reid-a:
Edits to the "Which Pipe?" exercise in episode 4. Removed the first candidate solution involving "grep" (not yet introduced at this point in the lesson), and added a sentence describing the "-c" option to the uniq command.

### DIFF
--- a/_episodes/04-pipefilter.md
+++ b/_episodes/04-pipefilter.md
@@ -618,19 +618,19 @@ so that you and other people can put those programs into pipes to multiply their
 > ~~~
 > {: .output}
 >
-> Assuming your current directory is `data-shell/data/`,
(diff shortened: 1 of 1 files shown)


gcapes:
Great stuff, thanks!

PR #986, opened 2019-06-13
Include definition of delimiter in episode 4

morganoneka:
In episode 4, the term "delimiter" was used without being defined previously. How the -d option is used in the cut command is described briefly, but the actual meaning of the word wasn't included. I added a definition of what a delimiter is and an explanation of what it means in the context of the cut command.


gdevenyi:
Good catch, thanks!

Issue #963, opened 2019-04-19
wc clarity


alouyakis:
The Unix Shell: Lesson 4. Pipes and Filters
Word count without flags is used to initially introduce the command and outputs three points of data - character, word, and line count. Then -l is added to show how line count is attained. It would be helpful to also include wc -cwl to show that the output is the same as default without flags. This would help by teaching about default flags and also increase exposure to varied usage by showing -c and -w in addition to -l. A possible addition (without showing the output):
```
$ cd molecules
$ wc *.pdb
```
The default flags for the wc command are -cwl. If we run wc -cwl instead of just wc, the output will be identical:
```
$ wc -cwl *.pdb
```
Each flag can also be used individually. By running wc -l instead of just wc, the output shows only the number of lines per file:
```
$ wc -l *.pdb
```

gdevenyi:
Thanks for the suggestion, a PR would be welcome to discuss default options.

jackVanish:
I'm not sure the Pipes and Filters place is the right lesson for this. I agree that teaching about default flags would be useful, but it seems like it's a little disjoint from the learning objectives of the Pipes and Filters lesson. 

What I think might be a better solution would be to take the 'General Syntax of a Shell Command', which is currently buried in the 'Navigating Files and Directories' lesson, and break it out to its own distinct lesson. This would tighten the focus of the Navigating Files and Directories lesson to only showcase cd, ls, and pwd, as well as lay the groundwork for later lessons with more clarity. This lesson could build on the material already in the 'General Syntax of a Shell Command', as well as incorporate the material on default flags using wc. That dovetails well, since the default flags are mentioned in the wc man page.

jackVanish:
I've pulled down the code and branched it, but I can't seem to push it up to open a pull request. Can anyone set me straight here?

gcapes:
@jackVanish
You would need to fork the repo first, clone your fork, then make changes in a new branch, push them to origin (your fork) and create a pull request from there.

I can help with more details if required.
Thanks

jackVanish:
Thanks very much. I've laid the base work for this in a pull request here: https://github.com/swcarpentry/shell-novice/pull/992. I think there are probably some minor tool-ups left, but I leave that to those more experienced than I to work out. I'm happy to take them on if someone identifies where I've fallen short.

gcapes:
Given #992 is closed, I'm going to close this issue. I guess the lesson is probably full enough already.

PR #992, opened 2019-06-28
Made changes and added General Syntax episode.

jackVanish:
Pursuant to discussion here (https://github.com/swcarpentry/shell-novice/issues/963), I've broken out the 'General Syntax' of a shell command to its own episode. This gives a stronger foundation to later lessons and cover basic concepts, including default flags and man pages, before they're introduced.

I've left the estimated time spots blank since I wasn't sure how best to estimate those. Also the lesson has few exercises, and I'm not sure if someone else better-equipped would want to build those in. Hopefully the syntax is all in order, though. I haven't changed much of the wording, but I did rewrite the questions and learning goals, as well as added the section on default flags.


_episodes/02-generalsyntax.md
```diff
@@ -0,0 +1,264 @@
...
+Applications/         System/
+Library/              Users/
+Network/              Volumes/
+~~~
+{: .output}
+
+> ## Default Flags
+>
+> Some commands, when run without any flags, will assume one or more by default.
+> A good example of this is the 'wc' command, which has the default flags -c, -l,
```
gcapes: Backticks for `wc` please.

Some questions learners might have at this point on seeing this new command:
- What is the wordcount command?
- What does it do?
- Can I run an example of it?

_episodes/02-generalsyntax.md
```diff
@@ -0,0 +1,264 @@
...
+> and -w. This means that running 'wc' is equivalent to running 'wc -clw'. If a
+> command has default flags, they will be listed in the manual, or 'man' page, usually under
+> the 'Description' section. We will cover man pages and how to read them in the
```
gcapes: Default options for `wc` are not listed anywhere in my man pages or `--help` output on Linux Mint. Which OS are you using?

_episodes/02-generalsyntax.md
gcapes: Backticks for `wc` please.

Some questions learners might have at this point on seeing this new command:
- What is the wordcount command?
- What does it do?
- Can I run an example of it?

_episodes/02-generalsyntax.md
jackVanish: I'm using Mac OSX. The part I'm referring to is at the bottom of the Description section (see attached screenshot) that reads 'The default action is equivalent to specifying the -c, -l and -w options." 

Screen Shot 2019-07-02 at 9 08 18 AM

_episodes/02-generalsyntax.md
jackVanish: Thinking this over, to really test wc you'd need to make a file or have one handy to run wc on. Making files seems like a lot for this early lesson- what do you think a good way to deal with this would be?

_episodes/02-generalsyntax.md
gcapes: Ah, I see. This isn't given in my man pages on Linux. We should be trying to avoid material where the output is very different depending on platform. I don't know what the man pages look like on Git Bash.

```
SYNOPSIS
       wc [OPTION]... [FILE]...
       wc [OPTION]... --files0-from=F

DESCRIPTION
       Print  newline,  word,  and byte counts for each FILE, and a total line if more than one FILE is specified.  A
       word is a non-zero-length sequence of characters delimited by white space.

       With no FILE, or when FILE is -, read standard input.

       The options below may be used to select which counts are printed, always  in  the  following  order:  newline,
       word, character, byte, maximum line length.

       -c, --bytes
              print the byte counts

       -m, --chars
              print the character counts

       -l, --lines
              print the newline counts

       --files0-from=F
              read  input  from the files specified by NUL-terminated names in file F; If F is - then read names from
              standard input

       -L, --max-line-length
              print the maximum display width

       -w, --words
              print the word counts

       --help display this help and exit

       --version
              output version information and exit

AUTHOR
       Written by Paul Rubin and David MacKenzie.

REPORTING BUGS
       GNU coreutils online help: <http://www.gnu.org/software/coreutils/>
       Report wc translation bugs to <http://translationproject.org/team/>

COPYRIGHT
       Copyright  ©  2016  Free  Software  Foundation,  Inc.   License  GPLv3+:  GNU   GPL   version   3   or   later
       <http://gnu.org/licenses/gpl.html>.
```

_episodes/02-generalsyntax.md
gcapes: I would drop this section.  Not sure what others think?

jackVanish: The impetus for creating the PR was better clarity around WC and its default flags but I'm not married to it being there. It doesn't seem like there's a good place to fit it in.

jackVanish: I've taken it out, which should address both of the two comments above.

_episodes/02-generalsyntax.md
gdevenyi: Implicit on the manpage, but its there:
> Print  newline,  word,  and byte counts for each FILE

_episodes/02-generalsyntax.md
```diff
@@ -0,0 +1,264 @@
+---
+title: "General Syntax of a Shell Command"
+teaching: ??
+exercises: ??
+questions:
+- "How do I tell the shell to do something?"
+- "How do flags and options modify a shell program?"
```
uoa-noel: Perhaps this question could be rephrased as "How does using flags and options change the behaviour of a shell program?"
Flags and options do not change a shell program in terms of its content. By specifying "behaviour", we could reduce the confusion a learner may have over what flags and options do.

ElnazAmanzadeh:
Hi. The basic lesson is very important as people wan to make it perfect. But in many cases they just prefer to use command prompt in windows, so as it is the lesson should be general to cover both Windows and Linux commands. I think his lesson is great. I also have a suggestion that if it is possible to go ahead more using Jupyter Notebook in the lessons, which means anaconda lesson comes before bash, then we go to the shell. It may be craze a little bit but since typing a command and seeing the result is almost direct maybe it helps more.

colinmorris:
I'm concerned that reorganizing the lesson content in this way may make it harder to follow. The current episode 2 (navigating files and directories), has a nice gradual progression where it starts by giving some practice with plain commands, then introduces a modified form that uses a flag, and finally shows an example with a flag and an argument. If we instead jump straight into the most complicated version (command with flag and argument), I think it may be disorienting.

Also, all the episodes except the first are structured around the theme of how to perform some task. By comparison, an episode on "General syntax of a shell command" feels a bit abstract, which might make it a harder sell.

gdevenyi:
I think I agree with @colinmorris that the current state of the lesson is targeted at a "how to get things done" flow. As such while this contribution is excellent, I don't think it works well with the current aims of the lesson. In addition, the adding another episode will increase the time load of this lesson which we can't do in light of the time balance against the other lessons for a two-day workshop.

I would however be very interested in adding this to https://github.com/carpentries-incubator/shell-extras as its great work!

PR #994, opened 2019-07-10
Update 04-pipefilter.md

terryn:
I thought to add some clarification about what concatenation actually means; it's not something I ever remember hearing before I took computer science and it may be unclear.

### DIFF
--- a/_episodes/04-pipefilter.md
+++ b/_episodes/04-pipefilter.md
@@ -141,10 +141,11 @@ lengths.txt
 {: .output}
 
 We can now send the content of `lengths.txt` to the screen using `cat lengths.txt`.
-`cat` stands for "concatenate":
-it prints the contents of files one after another.
-There's only one file in this case,
-so `cat` just shows us what it contains:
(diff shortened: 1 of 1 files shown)


slel:
How about something like
```
We can now print the content of `lengths.txt` to the screen using `cat lengths.txt`,
where `cat` is a program whose name is short for "concatenate", which means combine
lists or strings together into a single one.

The content of each file given to `cat` as an argument is sent to standard output,
i.e., printed to the screen. In the case of a single argument, as here, there is
not much concatenation happening but the content of the file is displayed.

This is a common way of looking at the content of a file, especially a short one,
when we do not intend to, or want to, modify it.
```

gcapes:
Thanks for the PR. I think the original explanation is clearer, and certainly more concise. 

We've recently reduced the references to standard input and output, in an attempt to reduce unnecessary technical complexity, so I think your explanation (while largely correct) is more difficult to understand.

I'll put in a PR to further explain concatenate - see what you think.

gdevenyi:
Thanks @gcapes merged at #995

PR #1003, opened 2019-08-06
Add setup instructions for Ubuntu on Windows

mclarade:
… windows tools that are compatible with git. It should provide a better user experience, and be more easily transferable between Linux and windows.


gcapes:
Hi @mclarade 

Thanks for the PR! 

Have you seen the link under this section?
>For advanced users, you may choose one of the following alternatives:

I think the consensus is that Git Bash provides the easiest setup and a satisfactory environment for this lesson as well as Git, so I would say the existing linked instructions are the way to go. Advanced Windows users can go ahead and follow them, but it doesn't distract from the main recommendation.

See #271 for further discussion.

Issue #1004, opened 2019-08-07
Discrepancy between text and figure in loop explanation


HoWol76:
Just going through the Loop-lesson and noticed that the loop is mainly explained by looping over the files basilisk.dat and unicorn.dat. But in the figure, it uses files called input-1.dat, input-2.dat, and input-3.dat.

I'm wondering whether learners would find this confusing.

Also, I'm not sure about the best way to fix this. I think having 3 iterations of the loop helps a bit more with the understanding, so just limiting oneself to 2 creatures could be detrimental.

I think the 'proper' way would be to add a 3rd creature to the curriculum (maybe a minotaur) and then make the figure match these three creatures?

PR #1005, opened 2019-08-08
add minotaur.dat. 

HoWol76:
The creatures (basilisk and unicorn) are used extensively to demonstrate how a loop works. The relevant figure instead uses 3 files: input-1.dat, input-2.dat, and input-3.dat.

This merge request adds a third creature file, the minotaur, to the list, and updates both the figure and the relevant sections to incorporate this. 

I have created an issue, https://github.com/swcarpentry/shell-novice/issues/1004, to explain my concerns. This is an idea how to solve this.

Considering that this is a very broad change for such a minor issue (including having to change the zip file), I'm not sure it needs to be incorporated. But if you want to, you can do this here.


gcapes:
Thanks for the PR @HoWol76!

Great attention to detail.

PR #1006, opened 2019-08-08
Minor fix to 05-loop.md after minotaur added

slel:
One extra instance of "two" should now be "three".

This is a follow-up to #1005.


gcapes:
Great, thanks!

PR #1013, opened 2019-08-14
Clarify learning objective in 07-find.md

monaghaa:
What changed?: Revised Learning Objective #2 to reflect that the "find" command can also be used for directories (one of the examples in the lesson uses the "$ find . -type d" command)

What specifically changed?:
Revised:
"Use `find` to find files whose names match simple patterns." 
to:
"Use `find` to find files and directories whose names match simple patterns."


gcapes:
Great, thanks!

Issue #1015, opened 2019-08-17
Objectives in episode 4: Pipes and Filters: `<` character


kairstenfay:
In the Pipes and Filters episode of the Unix Shell intro material, there is an objective listed that is not covered in the actual material. 

At the bottom of [the lesson page](http://swcarpentry.github.io/shell-novice/04-pipefilter/index.html), under Key Notes, the following is listed:
* `<` operator redirects input to a command

The same objective is also listed on [the reference page](http://swcarpentry.github.io/shell-novice/reference/).

If you do a Control-F (find) on the page, you will see that the `<` character is not mentioned in the lesson plan. All the other commands/characters in the objectives are mentioned many times. It seems this `<` command is either a typo or was neglected to be written into the lesson.

gcapes:
Thanks for the issue! This material was removed a while ago... I guess the objectives were overlooked.

A PR to remove these objectives would be welcomed!

gcapes:
Fixed in #1016

PR #1016, opened 2019-08-19
Remove `<` operator from Episode 4 objectives

kairstenfay:
The material for the `<` operator was removed awhile ago from the
episode 4 lesson plan, but the objectives were overlooked. Remove
the bullet point for the `<` operator from the objectives so that
it no longer appears in the Key Points section for Pipes and
Filters.

Please see issue #1015 for more context.


gcapes:
Great, thanks!

Issue #991, opened 2019-06-28
 Clarification for 'Loops' lesson in the Unix shell 


lshapiro31:
The first time I went through this lesson I was a bit confused by the rationale for the first loop at the top of the lesson. I am wondering if it might be clearer to show the field in the data files we are looking to isolate before running the loop? 

So for example, after this block of text:

Suppose we have several hundred genome data files named basilisk.dat, unicorn.dat, and so on. For this example, we’ll use the creatures directory which only has two example files, but the principles can be applied to many many more files at once. We would like to print out the classification for each species, which is given on the second line of the file. 

[ _Here add the following command to execute before running the loop_]

$ head -n 5 basilisk.dat unicorn.dat 
==> basilisk.dat <==
COMMON NAME: basilisk
CLASSIFICATION: basiliscus vulgaris
UPDATED: 1745-05-02
CCCCAACGAG
GAAACAGATC

==> unicorn.dat <==
COMMON NAME: unicorn
CLASSIFICATION: equus monoceros
UPDATED: 1738-11-24
AGCCGGGTCG
CTTTACCTTA

---

gcapes:
Hi @lshapiro31 

Is the confusion that the content of the files hasn't been shown, which means you're not sure what to expect as the output of the loop? Because the lesson does already explain that we want to print out the second line of the file. We could `cat` one of the files to show an example (which I think I usually do when teaching this), then move on to the rest of the explanation.

Would that address the problem?

Thanks

mgibbswalton:
Yes!  So I just finished demo teaching this exact lesson, and I think a  cat command prior to running the loop to "know your data" would make a lot of sense.  It does say "We would like to print out the classification for each species, which is given on the second line of the file", but a I do think a cat would be good to do within this context:

Change We would like to print out the classification for each species, which is given on the second line of the file.

to

We would like to print out the classification for each species, which is given on the second line of the file as shown in the command below (for the basilisk.dat file)
cat basilisk.dat
 1  COMMON NAME: basilisk
     **2  CLASSIFICATION: basiliscus vulgaris**
     3  UPDATED: 1745-05-02
....

kathrynnapier:
I have just been running through this lesson in preparation for a demo and found myself providing some context on the structure of the file by viewing it first and explaining it really helped me with the flow of the lesson. I agree with the suggestion by @mgibbswalton on changing the text slightly to make this a bit clearer.

[...]

PR #1019, opened 2019-08-22
Text amended to clarify genome file structure

kathrynnapier:
Minor amendments to the text to describe the structure of the genome files by first viewing them using head- this makes it a bit clearer why we want to progress to using a loop to print out the second line of each file. See issue 991 for more discussion.

https://github.com/swcarpentry/shell-novice/issues/991

### DIFF
--- a/_episodes/05-loop.md
+++ b/_episodes/05-loop.md
@@ -28,13 +28,14 @@ As such they are key to productivity improvements through automation.
 Similar to wildcards and tab completion, using loops also reduces the
 amount of typing required (and hence reduces the number of typing mistakes).
 
-Suppose we have several hundred genome data files named `basilisk.dat`, `minotaur.dat`, `unicorn.dat`, and so on.
-For this example,
-we'll use the `creatures` directory which only has three example files,
-but the principles can be applied to many many more files at once.
-We would like to print out the classification for each species, which is given on the second line of the file.
-For each file, we would need to execute the command `head -n 2` and pipe this to `tail -n 1`.
-We'll use a loop to solve this problem, but first let's look at the general form of a loop:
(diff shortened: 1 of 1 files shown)


_episodes/05-loop.md
```diff
@@ -28,13 +28,14 @@ As such they are key to productivity improvements through automation.
...
-For each file, we would need to execute the command `head -n 2` and pipe this to `tail -n 1`.
-We'll use a loop to solve this problem, but first let's look at the general form of a loop:
+In the `creatures` directory, we have three genome data files named `basilisk.dat`, `minotaur.dat`, and `unicorn.dat`. The structure of these files is the same: the common name, classification, and updated date are presented on the first three lines, with DNA sequences on the following lines. Let's look at the files:
+
+```
+head -n 5 basilisk.dat minotaur.dat unicorn.dat
+```
+{: .language-bash}
+
+We would like to print out the classification for each species, which is given on the second line of each file. For each file, we would need to execute the command `head -n 2` and pipe this to `tail -n 1`. In this particular example we only have three files, but imagine if we had hundreds of files- we can therefore use a loop to perform the same action on many different files. Firstly, let's look at the general form of a loop:
```
gcapes: I think we can keep the original wording for the initial explanation:
>Suppose we have several hundred genome data files named `basilisk.dat`, `minotaur.dat`, `unicorn.dat`, and so on.
For this example, 
we'll use the `creatures` directory which only has three example files,
but the principles can be applied to many many more files at once.

Then your new wording
> The structure of these files is the same: the common name, classification, and updated date 
are presented on the first three lines, with DNA sequences on the following lines. Let's look at the files:
```
head -n 5 basilisk.dat minotaur.dat unicorn.dat
```

> We would like to print out the classification for each species, which is given on the second line of the 
file.

Then back to the original wording
> We'll use a loop to solve this problem, but first let's look at the general form of a loop:

kathrynnapier:
Thanks for the feedback! Have made the changes as suggested, b=please let me know if there are any other issues. Much appreciated!

gcapes:
Thanks!

Issue #1009, opened 2019-08-08
Standardize use of quotation marks in lesson


mimishu:
Hi there, in reviewing the lesson of http://swcarpentry.github.io/shell-novice/07-find/index.html I got really confused with the uses of single quote vs. double quotes in the command examples, for example: 

$ grep -w "is not" haiku.txt 
$ grep -E '^.o' haiku.txt

I imagine double quotes are used around words and single quotes are used around characters? There is no explanation on differences of their uses anywhere in this lesson and I think it will be useful for novice learners. I suggest adding a sentence explaining their uses, for example (from https://stackoverflow.com/questions/6697753/difference-between-single-and-double-quotes-in-bash):

"Single quotes won't interpolate anything, but double quotes will. For example: variables, backticks, certain \ escapes, etc. Enclosing characters in single quotes ( ' ) preserves the literal value of each character within the quotes. A single quote may not occur between single quotes, even when preceded by a backslash."

I hope this helps! Best, Shu (shu.liu@uci.edu)

gcapes:
Thanks for bringing this up! 
While there is this difference in behaviour between single and double quotes, I don't think it actually makes a difference for any of the `grep` or `find` examples in the lesson - indeed I think most of them run just fine with no quotes at all.

I agree that using different quote types for different examples, and even within the same example does rather raise the question as to why. Perhaps it would be good to stick with double quotes throughout for consistency and to reduce this distraction?

gdevenyi:
Good catch on these small details.

For the purposes of these lessons, we should unify for a single type. I don't believe we cover the difference and I think that would be out-of-scope of this novice-level lesson.

gcapes:
Let's stay with double quotes - I think they are the first type introduced, and are more frequently used over the episodes.

jttkim:
The difference is that single quotes preserve the literal value of each character while there are exceptions for double quotes, most notably `$` which retains its special role as a variable de-referencing operator.

I think it makes sense to mention this briefly in the episode. This is mainly because in my experience, learners are very likely to find examples using single and double quotes somewhere in the web, and in the majority of workshops (Carpentry or otherwise), questions / remarks along the lines "I found out that I can also use single / double quotes and it seems to work just as well" will surface.

[...]

PR #1020, opened 2019-08-28
Use single quotes to quote in text

### LINKED ISSUE #787: Single quote vs Double quote
The lesson has

> Let’s go into that directory with cd and run the command wc *.pdb. wc is the “word count” command: it counts the number of lines, words, and characters in files.

and

> On the other hand, p*.pdb only matches pentane.pdb and propane.pdb, because the ‘p’ at the front only matches filenames that begin with the letter ‘p’.

The "New Oxford Style Manual" says 

> People writing for the Internet should note that single quotation marks are regarded as easier to read on a screen than double ones. British practice is normally to enclose quotation marks, and to use double quotation marks for a quotation within a quotation:
>
> > 'Have you any idea,' he said, 'what "red mercury" is?'
>
> The order is often reversed in newspapers, and uniformly in US practice:
>
> > "Have you any idea," he sad, "what 'red mercury' is?"

@swcarpentry/shell-maintainers Can we define the style that we will going to use?

gcapes:
Fix #787.

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -51,7 +51,7 @@ the example above can be accomplished in a few minutes at most.
 The heart of a command-line interface is a **read-evaluate-print loop** (REPL). It is called
 so because when you type a command and press <kbd>Return</kbd> (also known as <kbd>Enter</kbd>) the shell
 reads your command,
-evaluates (or "executes") it,
+evaluates (or 'executes') it,
 prints the output of your command,
 loops back and waits for you to enter another command.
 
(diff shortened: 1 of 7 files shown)


gdevenyi:
This modifies all the quotes including those that are just textual, rather than the ones in code. I don't think that was the point of #787

gcapes:
If I've done it correctly, it should *only* modify textual quotes - conforming to a consistent style guide is the point of #787. Consistency of single-vs-double quotes in code are a separate issue #1009.

gdevenyi:
Oh wow I completely missed there was a separate issue. I guess we're good.

PR #1025, opened 2019-09-02
Fixing some broken html in the pipe exercise

Denubis:
One of my students noticed the exercise having some broken html.

(They would also like some more discussion of delimiters and that sort of capability, but that's outside the scope of this pull request)

### DIFF
--- a/_episodes/04-pipefilter.md
+++ b/_episodes/04-pipefilter.md
@@ -501,7 +501,8 @@ so that you and other people can put those programs into pipes to multiply their
 >
 > The `cut` command is used to remove or "cut out" certain sections of each line in the file,
 > and `cut` expects the lines to be separated into columns by a <kdb>Tab</kbd> character.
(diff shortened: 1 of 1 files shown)


Denubis:
I'm not quite sure why it isn't indicating the <kbd>Tab</kbd> and the "<code>" to ` fix in the commit, but that's the actual significant part. The bold to italics and new paragraph are less relevant.

gcapes:
Thanks for the PR! I guess you've got unstaged changes, or haven't pushed them yet which is why the diff doesn't show them. I'll make the fix for you - thanks again for pointing it out.

I don't feel strongly about bold/italics, but here's the style guide  on that: http://carpentries.github.io/lesson-example/06-style-guide/index.html#strong-emphasis 
I think we're consistent with it - we're introducing a new term here. What do you think?

Denubis:
Oh, sorry. Yeah, it should be bold.

PR #1026, opened 2019-09-02
Remove demotivating use of just

### LINKED ISSUE #1023: auditing the lesson for demotivating use of j (just) word
According to the teaching tips and the train-the-trainer workshop, it is recommended that we should avoid using the 'J' (just) word. However, from the lesson materials, there are a few of them. Would it be good if we can avoid this word in the lesson? I understand that in some context, this word can be used as well.

gcapes:
And a few that weren't demotivating, but unnecessary.

Fix #1023.

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -794,7 +794,7 @@ Run `pwd` and `ls -F` to ensure that we're in the directory we expect.
 
 ### Nelle's Pipeline: Organizing Files
 
-Knowing just this much about files and directories,
+Knowing this much about files and directories,
 Nelle is ready to organize the files that the protein assay machine will create.
 First,
 she creates a directory called `north-pacific-gyre`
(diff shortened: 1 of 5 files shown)


_episodes/02-filedir.md
```diff
@@ -101,7 +101,7 @@ because its name begins with `/`.
 > Notice that there are two meanings for the `/` character.
 > When it appears at the front of a file or directory name,
 > it refers to the root directory. When it appears *inside* a name,
-> it's just a separator.
```
gdevenyi: I not sure this removal is appropriate.

gdevenyi:
Most of these instances are not "demotivating" usage, but that's fine. I'm happy with the grammar adjustment.

_episodes/02-filedir.md
gcapes: Yes there were a few I wasn't sure about. I'll fix this one.

PR #1030, opened 2019-09-05
change find-file-tree image to larger png

hoytpr:
The find-file-tree.svg image is too small to read. While using this lesson, the .png image in this PR was much easier to read. You might prefer .svg images, but this image is essentially the same and works well.


gcapes:
Hi Peter,
Thanks for bringing this up. I think svg files are preferred because they can be zoomed easily, whereas png cannot. I've just changed the default size of the svg file using draw.io and it looks to work well, retaining the ability to zoom the image easily in a browser. 

I'll merge that change and close this PR.
Thanks again for bringing this up!

hoytpr:
Hi Gerald, the image is much easier to read now! But two things to mention:
1. The "o" in "one.txt" is cut off now.
2. I notice that the "LittleWomen.txt" file isn't present in the image.
(FYI in Firefox at least, the png files seem to zoom with the browser just fine, using ctl_+ and ctl_-)
Thanks for your hard work and good luck!

gcapes:
1. Hmm. I have no idea about this one. The image displays correctly for me in Chrome, Opera, and Firefox.
2. `LittleWomen.txt` wasn't present in the previous version of the file.

hoytpr:
That's odd. here's a link to a screenshot from Firefox. 
https://drive.google.com/open?id=1cYbQFc4wOhthe6BdZNsBMgLQnyZk_SBn
Tried right-click-view-image, and "o" still cut off. Zoomed way in and way out... no change. 

But you're right, it **works in Chrome** (although the lettering is much smaller). I'll see if I can figure out what needs to be changed in my Firefox.

hoytpr:
Okay it's working now. Not sure exactly what happened, but the new version of Firefox (64-bit) may have been conflicting with an older install that wasn't cleaned out. 

Looks good!
-p

gcapes:
Great!

PR #1045, opened 2019-10-17
Simpler example to mv files to a new directory

ErinBecker:
The example of moving files to the current directory could seem overly complicated to someone who is new to the shell.  I felt like a better way to solve a problem like this would be to show you can cd into the directory you misplaced the files in and using the mv command put the multiple files on one line pointing to the new directory.   So the solution for getting the files into the raw/ folder would be: 
$ cd analyzed
$ mv sucrose.dat maltose.dat ../raw

Seems much simpler this way.  I would also reinforce to students the importance of tab completion.   At the very least, it'd be good to show this method as an alternative to the solution provided.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -336,30 +336,30 @@ quotes.txt
 ~~~
 {: .output}
 
-> ## Moving to the Current Folder
+> ## Moving Files to a new folder
 >
 > After running the following commands,
-> Jamie realizes that she put the files `sucrose.dat` and `maltose.dat` into the wrong folder:
+> Jamie realizes that she put the files `sucrose.dat` and `maltose.dat` into the wrong folder.  The files should have been placed in the raw folder.
 >
 > ~~~
 > $ ls -F
 >  analyzed/ raw/
 > $ ls -F analyzed
 > fructose.dat glucose.dat maltose.dat sucrose.dat
-> $ cd raw/
+> $ cd analyzed
 > ~~~
 > {: .language-bash}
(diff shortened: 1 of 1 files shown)


_episodes/03-create.md
```diff
@@ -336,30 +336,30 @@ quotes.txt
 ~~~
 {: .output}
 
-> ## Moving to the Current Folder
+> ## Moving Files to a new folder
 >
 > After running the following commands,
-> Jamie realizes that she put the files `sucrose.dat` and `maltose.dat` into the wrong folder:
+> Jamie realizes that she put the files `sucrose.dat` and `maltose.dat` into the wrong folder.  The files should have been placed in the raw folder.
```
gcapes: Please split this long line by adding a new line after the first sentence.
Also, backticks for the `raw` folder please.

_episodes/03-create.md
```diff
@@ -336,30 +336,30 @@ quotes.txt
...
 > $ ls -F analyzed
 > fructose.dat glucose.dat maltose.dat sucrose.dat
-> $ cd raw/
+> $ cd analyzed
 > ~~~
 > {: .language-bash}
 >
-> Fill in the blanks to move these files to the current folder
-> (i.e., the one she is currently in):
+> Fill in the blanks to move these files to the raw/ folder
```
gcapes: backticks for `raw` please.

gdevenyi:
+1 a great idea just need the adjustments suggested by @gcapes

PR #1048, opened 2019-10-23
Update 02-filedir.md

ajstewartlang:
It seems perhaps a little unclear to write that 'our home directory contains mostly **sub-directories**' given that they are all actually sub-directories. I'm proposing changing 'mostly' to 'only' as that seems to better match the situation.


gdevenyi:
Cheers thanks!

Issue #1033, opened 2019-09-12
Making sure you are on a Bash shell


mnGopher:
Hi,

in your 'introduction the shell' unit, there is a section called "The Shell" where you mention that the lesson applies to the Bash  shell. The assumption is that the students will be using a Bash shell, but why don't you use the command **_ps_** to confirm that the actual shell they are using is Bash. thanks, yectli

colinmorris:
This is briefly alluded to in the [Setup](http://swcarpentry.github.io/shell-novice/setup.html) section:

> The default Unix Shell for Linux operating systems is usually Bash. On most versions of Linux, it is accessible by running the (Gnome) Terminal or (KDE) Konsole or xterm, which can be found via the applications menu or the search bar. If your machine is set up to use something other than bash, you can run it by opening a terminal and typing `bash`.

Maybe I'm wrong, but it seems like it would be a very rare scenario for a learner to be using a unix shell other than bash. It might be better for an instructor/helper to deal with such cases individually if they come up.

A downside of having everyone do this check as part of episode 1 would be that the output of `ps` could be intimidatingly opaque to novices.

mikerenfro:
This topic came up during the October 24-25 instructor training sessions. MacOS Catalina has a default shell of `zsh`, while earlier versions have a default shell of `bash`. If we assume that students will bring their own devices to the workshop, `zsh` will become increasingly common for students using MacOS.

But it should be easy enough to check the output of `echo $SHELL` instead of using `ps`. And I need to put together a contribution for my checkout process anyway.

nsoranzo:
See also issue https://github.com/swcarpentry/shell-novice/issues/978

mikerenfro:
Resolved in 8b4908d ?

gdevenyi:
Fixed by #1049 

Thanks!

wirawan0:
Pardon me for commenting on an old, closed issue. I just spotted this one and want to weigh on.

While we mentioned everywhere that we are using "bash" shell, I don't recall any bash-specific feature actually being used in our lesson. We don't use thing such as:

* Array variables, like ${VAR[index]}
* The `[[ ]]` , `(( ))` , `==`, `<()` and other funky operators
* Regular expression matching  (Bash >= 4)
* Extended matching operator like `@()` , `+()`
* Bash-specific commands (`shopt`, custom completion)
* and many other Bash-specific syntax, variables, features, or semantics

[...]

PR #1049, opened 2019-10-29
Adding simple checks for default shell, adjusting Mac instructions to…

mikerenfro:
… account for zsh.

Attempt at resolving #1033.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -27,18 +27,15 @@ In the lesson, you will find out how to access the data in this folder.
 > or [(KDE) Konsole](https://konsole.kde.org/)
 > or [xterm](https://en.wikipedia.org/wiki/Xterm),
 > which can be found via the applications menu or the search bar.
-> 
-> To check if your machine is set up to use something other than Bash, type `echo $SHELL` in your terminal window.
->
 > If your machine is set up to use something other than Bash, you can run it by opening a terminal and typing `bash`.
->
-> ### Mac OS
-> For a Mac computer running Mac OS Mojave or earlier releases, the default Unix Shell is Bash.
-> For a Mac computer running Mac OS Catalina or later releases, the default Unix Shell is Zsh.
-> Your default shell is available via the Terminal Utilities program within your Applications folder.
+
+> ### macOS
+> For a Mac computer running macOS Mojave or earlier releases, the default Unix Shell is Bash.
+> For a Mac computer running macOS Catalina or later releases, the default Unix Shell is Zsh.
+> Your default shell is available via the Terminal program within your Utilities folder.
 >
 > To open Terminal, try one or both of the following:
-> * Go to your Applications. Within Applications, open the Utilities folder. Locate Terminal in the Utilities folder and open it.
+> * In Finder, select the Go Menu, then select Utilities. Locate Terminal in the Utilities folder and open it.
 > * Use the Mac ‘Spotlight’ computer search function. Search for: `Terminal` and press <kbd>Return</kbd>.
 >
 > To check if your machine is set up to use something other than Bash, type `echo $SHELL` in your terminal window.
--- a/setup.md
+++ b/setup.md
@@ -27,15 +27,18 @@ In the lesson, you will find out how to access the data in this folder.
 > or [(KDE) Konsole](https://konsole.kde.org/)
 > or [xterm](https://en.wikipedia.org/wiki/Xterm),
 > which can be found via the applications menu or the search bar.
+> 
+> To check if your machine is set up to use something other than Bash, type `echo $SHELL` in your terminal window.
+>
 > If your machine is set up to use something other than Bash, you can run it by opening a terminal and typing `bash`.
-
-> ### macOS
-> For a Mac computer running macOS Mojave or earlier releases, the default Unix Shell is Bash.
-> For a Mac computer running macOS Catalina or later releases, the default Unix Shell is Zsh.
-> Your default shell is available via the Terminal program within your Utilities folder.
+>
+> ### Mac OS
+> For a Mac computer running Mac OS Mojave or earlier releases, the default Unix Shell is Bash.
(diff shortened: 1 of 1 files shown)


gcapes:
Thanks for the PR.

What do we think about checking the default shell on Linux? My thoughts are that it is so unlikely to anything other than Bash that we could just amend the Mac instructions.

mikerenfro:
A quick web search didn't return any Linux distribution using anything other than bash for the default shell. So I'm good with just restricting it to macOS if others agree.

gdevenyi:
+1 for adjusting to be macOS only. Please conform to #1050 naming :)

mikerenfro:
> +1 for adjusting to be macOS only. Please conform to #1050 naming :)

Done, and I also wrote up what should be a simpler way to get Mac users to their Terminal through Finder.

setup.md
```diff
@@ -27,16 +27,24 @@ In the lesson, you will find out how to access the data in this folder.
...
-> If your machine is set up to use something other than bash, you can run it by opening a terminal and typing `bash`.
+> 
+> To check if your machine is set up to use something other than Bash, type `echo $SHELL` in your terminal window.
+>
+> If your machine is set up to use something other than Bash, you can run it by opening a terminal and typing `bash`.
 >
 > ### macOS
-> For a Mac computer, the default Unix Shell is Bash,
-> and it is available via the Terminal Utilities program within your Applications folder.
+> For a Mac computer running Mac OS Mojave or earlier releases, the default Unix Shell is Bash.
```
gdevenyi: Can you please adjust naming to confirm with #1050

setup.md
```diff
@@ -27,16 +27,24 @@ In the lesson, you will find out how to access the data in this folder.
...
+> For a Mac computer running Mac OS Catalina or later releases, the default Unix Shell is Zsh.
```
gdevenyi: Can you please adjust naming to conform with #1050

setup.md
```diff
@@ -27,16 +27,24 @@ In the lesson, you will find out how to access the data in this folder.
...
 >
 > To open Terminal, try one or both of the following:
-> * Go to your Applications. Within Applications, open the Utilities folder. Locate Terminal in the Utilities folder and open it.
+> * In Finder, select the Go menu, then select Utilities. Locate Terminal in the Utilities folder and open it.
 > * Use the Mac ‘Spotlight’ computer search function. Search for: `Terminal` and press <kbd>Return</kbd>.
 >
+> To check if your machine is set up to use something other than Bash, type `echo $SHELL` in your terminal window.
+>
+> If your machine is set up to use something other than Bash, you can run it by opening a terminal and typing `bash`.
+>
```
gdevenyi: This section is a duplication of the section above, is that intentional?

setup.md
mikerenfro: Adjusted.

setup.md
mikerenfro: My assumption was that Linux users would use the Linux section, and macOS users would  use the macOS section. Though the instruction could be factored out and placed at the end, it might be more trouble to indicate that the instruction applies to both Linux and macOS users.

gcapes: While true, I thought we [agreed that this PR would address Macs only ](https://github.com/swcarpentry/shell-novice/pull/1049#issuecomment-547906483)

mikerenfro: Correct, and I may have mixed up a couple of separate but related issues. The current setup.md for Linux users tells them to type `bash` if their default shell isn't set to Bash, but doesn't explain how to check what their default shell is.

Though it's quite unlikely a Linux system will use something other than Bash, I wasn't comfortable with removing existing instructions.

mikerenfro:
Checking back in. The original Linux wording for "If your machine is set up to use something other than Bash" dates to [October 2017](https://github.com/swcarpentry/shell-novice/commit/c3f17144156b2bee27f452fb24297670f224a37e#diff-2cc0b07f42930e4520820be1fe66f913R31), and my commit adds a check for both Mac and Linux users to verify if their machine is set up that way.

I'm good to do any of:

- leave this PR as-is, which is a mix of mostly Mac instructions, with one Linux addition to parallel the new Mac language

- remove the Linux part of this PR, and submit a second PR just for Linux that will use the same language for checking your shell as this PR does for Mac

- remove the Linux part of this PR, keeping this as a change for just Mac users, and leave the Linux language untouched with no additional PRs

- remove any reference to alternative shells in Linux, as they're quite unlikely to occur (either in this PR or another one)

Just let me know, thanks.

gdevenyi:
I think we can safely remove the linux "check your shell" stuff as bash is the universal default for anything the learners will ever see.

Thanks!

setup.md
```diff
@@ -27,9 +27,6 @@ In the lesson, you will find out how to access the data in this folder.
 > or [(KDE) Konsole](https://konsole.kde.org/)
 > or [xterm](https://en.wikipedia.org/wiki/Xterm),
 > which can be found via the applications menu or the search bar.
-> 
-> To check if your machine is set up to use something other than Bash, type `echo $SHELL` in your terminal window.
->
 > If your machine is set up to use something other than Bash, you can run it by opening a terminal and typing `bash`.
```
gdevenyi: This too please :)

mikerenfro: I'm seeing two different contexts for this comment. One shows the sentences for checking your shell on line 31. Another shows that line removed, and the "If your machine is set up to use something other than Bash" line is line 30, a sentence that dates back 2 years. [My current fix-1033 branch](https://github.com/mikerenfro/shell-novice/blame/fix-1033/setup.md#L30) shows the latter.

Are you suggesting removing the "If your machine is setup to use something other than Bash" line, or were you referring to the line showing how to check if the shell was set to Bash?

gcapes:
If anyone feels strongly about amending the Linux instructions, please create a new issue/PR. Thanks.

PR #1050, opened 2019-10-30
Updating Apple operating system references to "macOS"

mikerenfro:
"macOS" has been the name since 2016, and [Wikipedia](https://en.wikipedia.org/wiki/MacOS_version_history) doesn't capitalize it at the start of a sentence. I think I caught all the variants of OSX, OS X, Mac OS, etc.


gdevenyi:
Thanks for the update, and providing a reference to the lowercasedness :)

PR #1051, opened 2019-11-01
Update template

fmichonneau:
as we are preparing the [transition to using bundler](https://github.com/carpentries/maintainer-RFCs/issues/1), @gcapes [found an issue](https://github.com/carpentries/lesson-example/pull/276#issuecomment-538317326) that can be solved by upgrading to the latest lesson template. This PR brings it in.


gdevenyi:
Will leave this for @gcapes to address since he caught the bug.

gcapes:
I've no time to test this, but trust that you've fixed the problem :)

PR #1064, opened 2019-11-19
Update 04-pipefilter.md

mianasbat:
I added CTRL + D option to get the output of `wc -l` to get the result on the fly as an option. Which is the purpose of the command when no input is specified.

Please delete this line and the text below before submitting your contribution.

---

---


gdevenyi:
Hi.

Thanks for the contribution. I think this might be outside the scope of the lesson, since we never use it anywhere afterwards and it will undoubtedly bring up questions about control sequences and other things which will increase the cognitive load on learners. Thoughts @gcapes?

Issue #1044, opened 2019-10-10
Ep.2  ls capitalisation example


flacchy:
I would like to make a suggestion for the episode 2 Navigate files and directories. Under the session "General syntax of a shell command" there is a mention of capitalisation but not clear example, if a student is looking at the material they might try 

`ls -r `
`ls -R`

and this can cause panic if.

Might be an option to add an example after the text 

> Each part is separated by spaces: if you omit the space between ls and -F the shell will look for a command called ls-F, which doesn’t exist. Also, capitalization can be important: ls -r is different to ls -R.

and say: 

We can try and type the following commands: 

`ls -F`
`ls -f`

what is the difference between the two? 

answer: 
by using the uppercase (-F) `ls -F` this  allow us to display the content in a more clear way by highlighting files and directories; if we use the lowercase (-f) we are viewing an unsorted output and also switching on a different flag (-a) without needing to call this specifically.

gdevenyi:
Fixed by #1066

PR #1066, opened 2019-11-29
Update 02-filedir.md

edanscriven:
Added an example of options to 'ls' that differ by capitalisation, as suggested in Issue #1044 .


gdevenyi:
Great addition, thanks!

PR #1070, opened 2019-12-06
Fix formatting of keyboard key combinations

### LINKED ISSUE #1069: Audit use of keyboard combinations in lesson
PR #1068 made me realize that we're not conforming everywhere to the style guide for keyboard combos. We should audit and fix any instances which are wrong.

gcapes:
Use keyboard symbols for keys users should press,
and upper case when referring to letter keys.

Fix #1069.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -153,22 +153,23 @@ $ nano draft.txt
 {: .callout}
 
 Let's type in a few lines of text.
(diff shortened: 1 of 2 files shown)


slel:
In the file `_episodes/05-loop.md`, why not
change to `<kbd>Ctrl</kbd>+<kbd>A</kbd>`
and ` <kbd>Ctrl</kbd>+<kbd>E</kbd>`?

gcapes:
Indeed, why not! Well spotted.

gdevenyi:
Wow, I'm surprised that we only had these small number sneak in. LGTM.

PR #1071, opened 2019-12-07
small change to specify that the "molecules" directory is in the main data-shell directory

slacalle:
small change to specify that the "molecules" directory is in the main "data-shell" directory. This is in line with the examples at the end of the previous lesson. It is aimed to minimize confusion for students.


gdevenyi:
Thanks for the small focused change and excellent description!

PR #1072, opened 2019-12-07
Added image of shell command anatomy.

slacalle:
I believe this helps students in the creation of mental concept maps and better absorb the following descriptions of each component parts.


dpshelio:
Cool!!! Thanks @slacalle

gcapes:
Thanks @slacalle 
This is a nice idea, but the image is too big. Could you please resize it so it's a comparable size to the other images in the episode?

deppen8:
I like this image, so I made the small resize change in a new PR #1250. (I don't think there is a way to do it here because I can't push to the original fork. Correct me if I am wrong there or if there is a better flow.)

PR #1074, opened 2019-12-12
Standardize use of quotes

### LINKED ISSUE #1009: Standardize use of quotation marks in lesson

gcapes:
Fix #1009.

### DIFF
--- a/_episodes/06-script.md
+++ b/_episodes/06-script.md
@@ -240,7 +240,8 @@ which means,
 'All of the command-line arguments to the shell script'.
 We also should put `$@` inside double-quotes
 to handle the case of arguments containing spaces
-(`"$@"` is equivalent to `"$1"` `"$2"` ...)
+(`"$@"` is special syntax and is equivalent to `"$1"` `"$2"` ...).
+
 Here's an example:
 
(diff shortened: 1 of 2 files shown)


gdevenyi:
LGTM

Issue #1073, opened 2019-12-11
error in commands available for wc in pipes and filters lesson


scastanedabarba:
The lesson for pipes and filters has the options for `wc` wrong for printing the number of lines and characters (this may be a system specific problem that I am unaware of).
The commands in the lesson are:
-1    number of lines
-c    number of characters

The correct commands are:
-l    number of lines
-m  number of characters

gcapes:
Well spotted! Pull Request anyone?

PR #1079, opened 2019-12-16
Change wc option -c (byte count) to -m (char count)

mattipt:
Closes swcarpentry/shell-novice#1073

The wc command option -c counts bytes, not characters. The correct option for character count is -m.


gcapes:
Great, thanks!

PR #1081, opened 2020-01-13
Edit in command not found message, in case the program corresponding to that command is not installed or mis-typed.

TSL-RamKrishna:
Typing a misspelled command gives "Command not found error", however, in cases where tool is not installed, you may still get "Command not found error" and I think the beginners should know this as well.

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -101,7 +101,7 @@ Documents   Library     Music       Public
 > ~~~
 > {: .output}
 >
-> Usually this means that you have mis-typed the command.
+> Usually this means that you have mis-typed the command. You get the same error, when you type the tool command but the tool has not been installed.
 {: .callout}
 
 


_episodes/01-intro.md
```diff
@@ -101,7 +101,7 @@ Documents   Library     Music       Public
 > ~~~
 > {: .output}
 >
-> Usually this means that you have mis-typed the command.
+> Usually this means that you have mis-typed the command. You get the same error, when you type the tool command but the tool has not been installed.
```
gdevenyi: Hi, I don't really like the term "tool".

Perhaps reword to "You can get the same error if the command you want is not installed"

TSL-RamKrishna: Hi, Thanks for the suggestion. I will edit it.

gcapes: To add my 2p:
Sure it's *programs* which are installed? They have command line interfaces, or graphical user interfaces, but the command is what you issue from the shell in order to run the program. So you don't install a *command*. Or have I missed something?

slel:
Maybe phrase like this:

> This might mean you mis-typed the command, or that the
> program corresponding to that command is not installed.

Issue #1084, opened 2020-01-23
Add Header to Episode 02 


catherine-martlin:
I noticed that in Episode 2 we move straight from the command help section into learning about `cd` (around line 458) and that felt a bit rushed to me. I think it may help users to break up those two sections with a header/visual clue that we are changing topics.

catherine-martlin:
This was addressed in #1085 and merged in!

PR #1085, opened 2020-01-23
Added new header 

catherine-martlin:
To the portion where we switch to talking about moving through directories to separate it from the discussion on help commands


catherine-martlin:
See Issue #1084

gcapes:
A nice addition, thanks!

PR #1093, opened 2020-02-25
Modified redirects and pipes image to SVG format

nmckinlay:
Changed the code to reflect the redirects-and-pipes.svg instead of the original redirects-and-pipes.png.


gdevenyi:
Hi @nmckinlay thanks for your contribution.

Can you please check that this figure renders approximately the same size as the png on the page?

Also, could you please remove the png since it isn't needed. Thanks!

Issue #1099, opened 2020-03-30
Episode 2 - clarify wording around Desktop directory


jdusenberry:
In Navigating Files and Directories, I found the wording potentially confusing in places.  

For example, on line 527, the cd command is discussed as moving "us from our home directory onto our Desktop, then into the 'data-shell' directory,...".  I think of "onto the Desktop" as more of a GUI concept, and would suggest changing this to "...into the Desktop directory, then into...".  I think this would keep it consistent with the terminology used elsewhere for changing directories on the command line.

That said, I did not see anywhere in the lesson that explained that files in the Desktop directory (as a subdirectory of the home directory) show up on the GUI desktop.  If it doesn't further confuse the issue, I would suggest changing lines 477 to 480 to something like:

Your output should be a list of all the files and sub-directories in your
Desktop directory, including the `data-shell` directory you downloaded at
the [setup for this lesson]({{ page.root }}{% link setup.md %}).  On many systems, the contents of the Desktop directory are the same as your GUI Desktop.  Take a look at your Desktop to confirm that your output is accurate. 

Also in the paragraph starting on line 527, I would recommend changing 'ls' in line 530 to 'ls -F' to be consistent with the command shown on line 545.  This change will also help emphasize the difference between an option and an argument.

I've never done a pull request, but happy to learn how to do that if these changes sound reasonable.

gcapes:
Thanks for the issue(s)! 

PRs welcome - one per issue ideally, so I think there are potentially 3 here:
1. wording change 'into desktop directory'
2. GUI desktop = /home/Desktop
3. add option to `ls`

PR #1100, opened 2020-03-31
ls -F edits

jdusenberry:
This is for one of the topics raised in #1099.

The pull request is to make descriptive text consistent with the code blocks for `ls` vs `ls -F`.

This is the first time I've submitted a pull request, and for some reason it looks like my editor made changes (whitespace or CR/LF changes?) that I did not intend to make.  It is not obvious to me how to prevent that from happening.


gcapes:
Great, thanks!
It looks like your editor has stripped trailing whitespace. This is nearly always a good thing, but can most likely be controlled in your editor settings/preferences somewhere.

PR #1103, opened 2020-04-02
Explain connection between Desktop directory and the GUI desktop

jdusenberry:
Explain connection between the Desktop directory and the GUI desktop. 

I'm uncertain if learners will be familiar with the term "GUI" but this seemed like the simplest way to make the connection.

Part of #1099.


slel:
Maybe spell it out:

On many systems, the command line Desktop directory
is the same as the Desktop visible in the graphical user
interface (GUI) of the operating system.

PR #1104, opened 2020-04-15
1101 ls abbreviation on first use

deeppatelcs:
Moved **ls** abbreviation  explanation from episode 2 to episode 1.

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -75,7 +75,8 @@ Most importantly:
 when typing commands, either from these lessons or from other sources,
 *do not type the prompt*, only the commands that follow it.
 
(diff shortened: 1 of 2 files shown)


slel:
Please add a comma and a newline after "which stands for 'listing'".

gcapes:
Hi @deeppatelcs,  and thanks for the PR.

The line in ep1 is now a bit long and difficult to read. I would rephrase like this:
> So let's try our first command, `ls` which is short for listing. 
This command will list the contents of the current directory:

deeppatelcs:
Hi @slel and @gcapes I have rephrased it suggested by @gcapes in above comment.

PR #1109, opened 2020-05-18
Clarify directory navigation terminology

### LINKED ISSUE #1108: Entering and leaving directories
It looks like there has been discussion on the orientation of directory trees (I'm old so for me the root is at the bottom but I get that others prefer to consider it like a family tree) but I wonder if we could avoid enforcing the orientation by the use of up and down and rather talk about entering and exiting directories? That's what I do when teaching people to navigate the file system, treating directories like rooms that you enter and leave. That way you can describe `cd ..` as leaving the current directory and `.` as staying in the current directory which I think is more helpful to learners. It also helps when describing relative paths I find.

gcapes:
Fix #1108.

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -560,8 +560,9 @@ animals.txt       morse.txt     planets.txt     sunspot.txt
 ~~~
 {: .output}
 
-We now know how to go down the directory tree, but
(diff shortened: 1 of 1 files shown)


gdevenyi:
Looks good to me :+1:

Issue #840, opened 2018-08-29
Episode 5: Include explanation of "datafile"


mleverance:
In the Loops episode, the "Variables in Loops" exercise early in the episode introduces the term 'datafile' as follows:
What is the output of the following code?
>$ for datafile in *.pdb
> do
>    ls *.pdb
> done

Now, what is the output of the following code?
>$ for datafile in *.pdb
> do
>	ls $datafile
> done

This is the first instance of datafile. In the lead-up to this exercise, 'filename' is used in conjunction with the loop for *.dat files. It would be helpful to include an explanation of how datafile is different and why it's used in the exercise. Information about whether it's connected to any specific file extension types, or if there's a way to know when to use datafile vs. other terms when writing code, would be helpful for learners.

Datafile shows up again later in the lesson under "Nelle's Pipeline: Processing Files" so learners will see it again, but it isn't expanded upon in that part either.

gcapes:
Thanks for raising this issue. 
While the specific choice of variable name isn't explicitly introduced, immediately before this exercise is a section which explains that
> We have called the variable in this loop `filename` in order to make its purpose clearer to human readers. The shell itself doesn’t care what the variable is called; if we wrote this loop as: ...

and then it gives two other examples of the same loop using a different variable name.

My feeling is that this is a 'sticky note' question to be addressed by an instructor if it comes up. What do others think?

gdevenyi:
I agree with @gcapes that this concept was introduced and explained immediately before this step and doesn't need an explicit callout.

jasonnstanley:
i think i understand the issue. On a first reading a novice may get confused. For you old hands the change of variable is obvious. For a new learner however it may seem that this is a new control rather than a simple variable change. So the proposed (#1110) annunciation removes any doubt for the novice learner.

colinmorris:
I also agree with gcapes on this. I'll add that the text immediately before the exercise talking about variable names was added since this issue was originally filed. If you look at [the state of the episode](https://github.com/jwarthur/shell-novice/blob/ceb3117a2bea985b1ab8481a14463c6ad0cc0b9a/_episodes/05-loop.md) at that time, mleverance's commentary is spot on. But I would say the issue has since been addressed.

gcapes:
Thanks for that link @colinmorris - really useful. 
@mleverance thanks for the original PR all that time ago! I'll close given this has been addressed now.

PR #1110, opened 2020-05-26
fix-issue-840 instructor

jasonnstanley:
Let users know a new variable is being employed when looping through file names.

So earlier it was filename and $filename, then later ti becomes datafile and $datafile.

So just let users know its a simply use of variables that suit the example...


gcapes:
Merge/close decision pending further discussion on #840

PR #1117, opened 2020-06-25
Add Chromebook as option in setup.md

ambrown77:
Add a mention to Chomebook OS
Variety in OS systems and users, important to make sure all included and can participate in workshop. Should only add 30 seconds of discussion to the lesson.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -61,6 +61,10 @@ In the lesson, you will find out how to access the data in this folder.
 > See the reference below for information on installing and using PuTTY, 
 > using the Windows 10 command-line tool, or installing and using a Unix/Linux emulator.
 >
+>
+>
+> ### Chromebook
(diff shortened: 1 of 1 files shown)


gcapes:
Thanks for the PR! I have zero experience with Chrome OS. Can you follow the whole course using crosh/shell?

gdevenyi:
It's definitely possible if the chromebook supports android apps, because you can get "termux" which has a full apt repository and all the proper commands. I can't find enough details on Crosh to see what's installed by default to tell, but if the files are already download and unzipped, I suspect it would work.

ndporter:
It's worth considering here not only whether the Shell course can be done on ChromeOS but also the Git and Python/R - since they're often offered as a unit and it might be confusing if people started with ChromeOS then found out they had to use Win/Mac/Linux. It might be easier to point to web apps instead for all of these if possible - since that would allow all.

bkmgit:
This is one of several options.  Termux can also be used on Android.  It is also possible to use cloud environments.  While these may be mentioned, unless there is much demand, it is probably best not to focus on them until they have been more carefully checked.  As such, closing this, but thank you for the suggestion. Raised https://github.com/swcarpentry/shell-novice/issues/1377

PR #1123, opened 2020-07-19
change luck-based explanation to software-based

wiringa:
The "if you're lucky" text in [Navigating Files and Directories](http://swcarpentry.github.io/shell-novice/02-filedir/index.html) doesn't offer a fact-based explanation for cases when `man ls` doesn't provide usage examples.

> This will turn your terminal into a page with a description of
> the `ls` command and its options and, if you’re lucky, some
> examples of how to use it.

If the `man` page for `ls` on a learner's device doesn't include usage examples they aren't unlucky, they've found a `man` page entry that lacks any examples.

Some possible replacements for "if you're lucky" are

* "depending on your version of `ls` and its documentation",
* "depending on your operating system,"
* "depending on the computer," and
* "depending on your system."

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -401,8 +401,8 @@ $ man ls
 {: .language-bash}
 
 This will turn your terminal into a page with a description
-of the `ls` command and its options and, if you're lucky, some examples
-of how to use it.
+of the `ls` command and its options and, depending on your version of
+`ls` and its documentation, some examples of how to use it.
 
 To navigate through the `man` pages,
 you may use <kbd>↑</kbd> and <kbd>↓</kbd> to move line-by-line,
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -401,8 +401,9 @@ $ man ls
 {: .language-bash}
 
 This will turn your terminal into a page with a description
-of the `ls` command and its options and, depending on your version of
-`ls` and its documentation, some examples of how to use it.
+of the `ls` command and its options and, depending on your variant of the
+terminal software, or your operating system, supplying `ls` and its
+documentation, some examples of how to use it.
 
 To navigate through the `man` pages,
 you may use <kbd>↑</kbd> and <kbd>↓</kbd> to move line-by-line,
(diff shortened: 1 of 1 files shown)


_episodes/02-filedir.md
```diff
@@ -401,8 +401,8 @@ $ man ls
 {: .language-bash}
 
 This will turn your terminal into a page with a description
-of the `ls` command and its options and, if you're lucky, some examples
-of how to use it.
+of the `ls` command and its options and, depending on your version of
+`ls` and its documentation, some examples of how to use it.
```
gdevenyi, line 405: Thanks for the suggestion here.

I think I'd like something like "depending on your variant of the terminal software supplying `ls` and its documentation'

wiringa: Would a mix of terminal software and operating system be more appropriate?

"depending on your variant of the terminal software, or your operating system, supplying `ls` and its documentation”

I’d consider Git Bash as terminal software and shell. Using macOS for the operating system, using either Terminal or iTerm2, and either bash or zsh, the same ls binary and man page content are used for `ls`.

gdevenyi: Sounds fine with me.

gcapes: I would be in favour of a very short explanation here (or even no explanation because it's not a key point). All we want to convey to learners is that the man page they see for `ls` won't necessarily have any examples. Could we not just change 'if you're lucky' to 'maybe'? Or maybe just cut the current sentence short
> This will turn your terminal into a page with a description of the ls command and its options.

gdevenyi (approved):

gcapes:
Despite the approval from @gdevenyi, I think this change makes for a sentence which is needlessly complicated and difficult to follow. I would favour simplifying it as detailed in my previous comment:
https://github.com/swcarpentry/shell-novice/pull/1123#discussion_r500902312

wiringa:
I agree with @gcapes on simplification, my original suggestion and the modifications do complicate the material. The simplification does address my original concern over "lucky." I'm happy to add another commit with the previously suggested change if that's agreeable.

> This will turn your terminal into a page with a description of the ls command and its options.

gcapes:
Great, let's do that! Thanks.

wiringa:
Added commit with shortened sentence.

PR #1127, opened 2020-08-06
Added shebang callout for shell scripting episode

dallen1:
I added a callout explaining shebangs in the shell scripting episode because they are a necessary best practice when trying to ensure that code is repeatable across environments.

### DIFF
--- a/_episodes/06-script.md
+++ b/_episodes/06-script.md
@@ -391,9 +391,19 @@ $ cd ../north-pacific-gyre/2012-07-03/
 ```
 {: .language-bash}
 
+> ## Shebangs
+> Take a look at the first line of Nelle's script and see how it starts with the '#!' pattern?
+> This character sequence is called a shebang and it's used to declare the interpretter for the script to use.
+> It's best practice to declare your interpretter because there are many ways to run a shell script and this will ensure 
+> that the correct interpreter is used regardless of how the script was initiated.
+>
+> Nelle uses the /usr/bin/env bash' path to ensure that it will run using bash on any UNIX environment regardless of where bash is installed on that system. 
+{: .callout}
+
 She runs the editor and writes the following:
 
 ~~~
+#!/usr/bin/env bash
 # Calculate stats for data files.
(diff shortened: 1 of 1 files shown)


_episodes/06-script.md
```diff
@@ -391,9 +391,19 @@ $ cd ../north-pacific-gyre/2012-07-03/
 ```
 {: .language-bash}
 
+> ## Shebangs
+> Take a look at the first line of Nelle's script and see how it starts with the '#!' pattern?
+> This character sequence is called a shebang and it's used to declare the interpretter for the script to use.
```
gdevenyi, line 396: interpretter -> interpreter

gdevenyi:
Thanks for the addition, we should definitely callout the usage of the shebang in the example scripts.

_episodes/06-script.md
gcapes: What's an interpreter? I don't think this is covered anywhere in the lesson.

gdevenyi: Good point! We should use a term already in the lesson. Can we just use "program"?

gcapes:
I think we should consider the teaching notes in the instructor guide before going ahead with this. The bullet on why we don't teach shebang makes some very good points: http://swcarpentry.github.io/shell-novice/guide/#teaching-notes

gdevenyi:
Should the shebang just be removed from the scripts instead? I figured we should avoid the shebang because of the reasons cited.

gcapes:
Which script(s) contain a shebang? `goodiff` and `goostats` don't, and we've not introduced the shebang in the scripts we write in this episode as far as I can tell?

gdevenyi:
@gcapes good question, a quick grep shows there aren't any. I was primed to think there was by the first patch line of the PR
>  Take a look at the first line of Nelle's script and see how it starts with the '#!' pattern?

I think since there are no shebangs, and we don't cover PATH and executable or such, we can close this.

Thanks for your contribution, but for novice users this topic is out of scope.

PR #1128, opened 2020-08-12
Update 03-create.md

mathbionerd:
Update the "Tips" about not using spaces for file or directory names by using the same example in the Unix tutorial, and encourage students to test out for themselves to see what happens.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -87,14 +87,15 @@ creatures/  data/  molecules/  north-pacific-gyre/  notes.txt  pizza.cfg  solar.
 >
 > Complicated names of files and directories can make your life painful
 > when working on the command line. Here we provide a few useful
-> tips for the names of your files.
+> tips for the names of your files and directories.
 >
 > 1. Don't use spaces.
 >
 >    Spaces can make a name more meaningful,
 >    but since spaces are used to separate arguments on the command line
 >    it is better to avoid them in names of files and directories.
->    You can use `-` or `_` instead (e.g. `north-pacific-gyre/` rather than `north pacific gyre/`).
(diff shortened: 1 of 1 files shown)


gdevenyi:
Thanks for the suggestion!

Currently, the example refers to the contents of data-shell, which are mentioned immediately previously, I so I don't quite get the change regarding that. I do however like the suggestion to test out spaces, and the addition of directories to the naming explicitly.

Can you explain some more why data-shell is a better choice than the current wording?

mathbionerd:
Thank you for the response. I'm sorry about being confused by your comment though. The previous example was not to use "data-shell". Rather it was: 

"You can use `-` or `_` instead (e.g. `north-pacific-gyre/` rather than `north pacific gyre/`)."

I suggested to change this to referring to data-shell: 

"You can use `-` or `_` instead (e.g. `data-shell/` rather than `data shell/`)."

gdevenyi:
> using the same example in the Unix tutorial

I'm not sure what you mean here. I don't see any reason why we can't use the north-pacific-gyre, so I'm asking why you want to change it.

mathbionerd:
You certainly can use north-pacific-gyre. 

I proposed using "data-shell/" as the example because the directory name is already used in the lesson, and there is a really clear mistake that happens when the user types the commend without the hyphen, "mkdir data shell". This uses the exact example from the lesson to show why it is best practice to use a hyphen, underscore, or CamelCase, without adding as much to the cognitive load of the learner. 

It worked really nicely in my teaching demo. 

Feel free to ignore the suggestion.

gdevenyi:
I'm fine with this change, any comments @gcapes ?

gcapes:
The current directory at this point is `data-shell`. So there's already a `data` directory  - there will just be a new `shell` directory created. That's fine, just checking that's what you're expecting?

PR #1131, opened 2020-08-18
Remove difficult exercise

### LINKED ISSUE #893: Episode 7 "Finding Files With Different Properties" exercise too difficult/obscure?
For reference, here's the question:

> The `find` command can be given several other criteria known as “tests” to locate files with specific attributes, such as creation time, size, permissions, or ownership. Use `man find` to explore these, and then write a single command to find all files in or below the current directory that were modified by the user `ahmed` in the last 24 hours.  
Hint 1: you will need to use three tests: `-type`, `-mtime`, and `-user`.  
Hint 2: The value for `-mtime` will need to be negative—why?

### Issue 1: I think this is too hard

I timed myself solving this problem, and it took me about 6 minutes. (For reference, the total exercise time for episode 7 is given as 20 minutes, and it has 6 exercises)

Most of that time was spent trying to figure out the modification time test through reading the man page and trial and error. `find`'s time-based tests are quite subtle, and figuring out the solution requires more than just reading the `-mtime` entry in the man page. That entry directs us to "See the comments for -atime", and from there, if you're lucky, you'll notice the explanation of the syntax/semantics of numeric arguments a few lines above.

### Issue 2: Niche-y

I'm sure I've written thousands of `find` commands in my life, but I don't think I've ever had reason to use the `-user` flag. In fact I can probably count on one hand the number of times I've used tests other than `-name` or `-type`.

In the amount of time this exercise takes, we could probably give the learner a brief introduction to the `-exec` flag which I think is an extremely practical feature by comparison. (Though I'd be just as happy simply removing this exercise without adding a replacement. See #133 and #179 for discussions of whether to include `-exec`.)

gcapes:
Fix #893.

### DIFF
--- a/_episodes/07-find.md
+++ b/_episodes/07-find.md
@@ -660,24 +660,3 @@ about them.'
 > {: .solution}
 {: .challenge}
 
-> ## Finding Files With Different Properties
->
-> The `find` command can be given several other criteria known as "tests"
-> to locate files with specific attributes, such as creation time, size,
-> permissions, or ownership.  Use `man find` to explore these, and then
-> write a single command to find all files in or below the current directory
-> that are owned by the user `ahmed` and were modified in the last 24 hours.
->
-> Hint 1: you will need to use three tests: `-type`, `-mtime`, and `-user`.
->
-> Hint 2: The value for `-mtime` will need to be negative---why?
->
-> > ## Solution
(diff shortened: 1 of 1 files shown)


gdevenyi:
I agree with this removal. By the time someone has a question such as this there's lots of answers of stackoverflow.

PR #1132, opened 2020-08-24
Minor text clarifications, added MobaXterm as an option, provided direct links for PuTTY, MobaXterm, Cygwin

8cat:
Minor clarifications in the language of the "Windows" and "For advanced users..." sections of the page, added MobaXterm as a free SSH client option, and provided direct links for PuTTY, MobaXterm, and Cygwin. These changes are intended to provide additional options and more useful URLs (direct links to the named tools) for those wishing to access a Unix shell environment from a Windows system.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -18,7 +18,7 @@ In the lesson, you will find out how to access the data in this folder.
 >
 > Some computers include a default Unix Shell program.
 > The steps below describe some methods for identifying and opening a Unix Shell program if you already have one installed.
-> There are also options for identifying and downloading a Unix Shell program, a Linux/UNIX emulator, or a program to access a Unix Shell on a server.
+> There are also options for identifying and downloading a Unix Shell program, a Linux/Unix emulator, or a program to access a Unix Shell on a server.
 >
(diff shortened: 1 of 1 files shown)


gcapes:
Thanks for the PR. This section could certainly use some tidying.

I would be keen to simplify further, giving one (main) recommended option per OS. So for Windows: Git for Windows. This will help to keep the lesson easier to teach, and to follow, because there'll be less of a range of 'it's not exactly like that on my machine' type questions.

As such I'm not sure what the benefit is of giving *four* other options (WSL, Putty, Cygwin, and now MobaXterm).
Git for Windows is the simplest option for workshops because it also covers the Git lesson, and WSL is probably the best option for following the shell course. 

I would move all the extra options that we recommend under the 'Advanced users' section. But perhaps just WSL and MobaXterm. WSL because it gives the closest approximation to the course notes, and MobaXterm because it's easy to install and use.

Thoughts?

gdevenyi:
I think all the options presented here will work for shell-novice, but we need to make sure they interact properly with the other lessons which are in a standard workshop.

I agree with @gcapes that we should make one clear supported recommendation for starting off.

gcapes:
I think this has mostly been superseded by #1152?

PR #1134, opened 2020-08-26
Add more realistic regex example

gcapes:
Fix #930.


gdevenyi:
I'm neural on this. This regex is far more complex, and the modification also includes the addition of a "-r" option

deppen8:
Reviving this PR... 

I don't feel super strongly that the haiku example needs replaced, but if that is the route you want to go, how about something that searches the DNA sequences in `creatures/`? I am not a geneticist, but I could imagine you might want to find patterns in the sequences. Here is one that got three hits right away. (It also gets hits if you drop the `.{3}` for less complexity.)

```bash
$ cd creatures/
$ grep -E "^.{3}TTC" basilisk.dat
CATTTCGCGC
TGATTCCAGA
GCCTTCCTAT
```

PR #1135, opened 2020-08-26
Use POSIX convention for abstract examples

### LINKED ISSUE #1120: Using common conventions in lesson materials when providing abstract or incomplete code examples.
During most of our lesson materials and essentially all of our live coding, we are dealing with very literal instructions.  By that I mean ls -r means type exactly that.  However, occasionally, we wade into the more abstract voice.  An example of this is found in the "Key points" section of the [Unix Shell Navigating Files and Directories](http://swcarpentry.github.io/shell-novice/02-filedir/index.html) episode.  The text reads:  

**cd path changes the current working directory.
ls path prints a listing of a specific file or directory; ls on its own lists the current working directory.**

Here, we don't actually intend for the learner to type "ls path", but instead the word path is a parameter or argument to the command that is a placeholder for variable text or input.  I'd suggest we use common convention here to indicate as much.  example text changed is below:

**cd [path] changes the current working directory.
ls [path] prints a listing of a specific file or directory; ls on its own lists the current working directory.**

There are likely other examples, but this is the one that jumped out at me.  Thoughts?

---

If this issue is about a specific episode within a lesson, please provide its link or filename.

---

gcapes:
Fix #1120.


gdevenyi:
Looks great!

PR #1139, opened 2020-09-24
Update setup.md

### LINKED ISSUE #1137: Switch order of setup page
Currently the data download section, which has instructions that you should open your terminal, comes before the section that actually includes how to install terminal.  I suggest we move the install directions up.  

Could also maybe add headers to differentiate the install and the download sections.

froesch-21:
fixes #1137 

I think that the suggestion by sstevens2 is a good and useful, so I reorganized the setup page as suggested. 

This is my first contribution and I found a small mistake just before commiting. Because of that there are two commits. I could not find a way to merge these two commits.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -4,18 +4,10 @@ title: Setup
 root: .
 ---
 
-You need to download some files to follow this lesson.
+To follow this lesson you need to have access to the shell and have to download some files.
 
-1. Download [data-shell.zip][zip-file] and move the file to your Desktop.
-2. Unzip/extract the file.
-   **Let your instructor know if you need help with this step**.
-   You should end up with a new folder called **`data-shell`** on your Desktop.
-3. Open a terminal.
-   If you're not sure how to open a terminal in your operating system, see the instructions below.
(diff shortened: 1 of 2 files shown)


gdevenyi:
Thanks for the contribution. Should you be interested in learning a bit more git, here's some details on how to squash your commits together https://medium.com/@slamflipstrom/a-beginners-guide-to-squashing-commits-with-git-rebase-8185cf6e62ec

Otherwise, we can do it here during merge.

setup.md
```diff
@@ -113,3 +105,17 @@ you can run it by opening a terminal and typing `bash`.
 [gnome-terminal]: https://help.gnome.org/users/gnome-terminal/stable/
 [kde-konsole]: https://konsole.kde.org/
 [xterm]: https://en.wikipedia.org/wiki/Xterm
+
+
+>## Download and extract the data-shell.zip-file:
```
gdevenyi, line 110: I think the formatting is bad here? This is both a quote and a second-level title.

PR #1140, opened 2020-09-25
Rename recombine to recombined in chapter 3

PabRod:
In [chapter 3](https://swcarpentry.github.io/shell-novice/03-create/index.html) (Working With Files and Directories) we create a folder named `recombine`.

Folders named after imperative verbs can be mistaken for commands. Using `recombined` instead solves this minor issue.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -459,30 +459,31 @@ quotations.txt
 > ~~~
 > {: .output}
 > ~~~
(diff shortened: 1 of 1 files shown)


gdevenyi:
I like this, its something I've always implicitly done when working but didn't have any concept to explain.

gdevenyi:
Thanks!

PR #1146, opened 2020-10-08
remove custom permalinks from Extras pages

tobyhodges:
See discussion here for more background: https://github.com/datacarpentry/datacarpentry.github.io/issues/542

In short, some Carpentries lessons use `/guide/` for their Instructor Notes page and others use `/guide.html`. This PR removes the `permalink` field from YAML front matter of the Instructor Notes (`_extras/guide.md`) and other Extras files, to make established lessons consistent with new lessons created with [the lesson template](https://github.com/carpentries/styles/). Keeping the paths to these files consistent will help us avoid broken links on the [Software Carpentry Lessons page](https://software-carpentry.org/lessons/), and ensure that equivalent paths in new lessons created with the template are consistent with previously-developed lessons like this one.

If and when this PR is merged, I'll update the link on https://software-carpentry.org/lessons/ to the Instructor Notes page for this lesson to avoid a broken link there. If you know of anywhere else that would need updating to avoid a broken link, please let me know!

### DIFF
--- a/_extras/about.md
+++ b/_extras/about.md
@@ -1,6 +1,5 @@
 ---
 layout: page
 title: About
-permalink: /about/
 ---
 {% include carpentries.html %}
--- a/_extras/discuss.md
+++ b/_extras/discuss.md
@@ -1,7 +1,6 @@
 ---
 layout: page
 title: "Discussion"
-permalink: /discuss/
 ---
 ## Alphabet Soup
 
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -1,7 +1,6 @@
 ---
 layout: page
 title: "Instructor Notes"
-permalink: /guide/
 ---
 *   Why do we learn to use the shell?
     *   Allows users to automate repetitive tasks
--- a/reference.md
+++ b/reference.md
@@ -1,7 +1,5 @@
 ---
 layout: reference
-permalink: /reference/
-root: ..
 ---
 
 ## Summary of Basic Commands


tobyhodges:
Update: looks like this is the wrong approach after all. Closing now

tobyhodges:
And reopening after [further consultation with @fmichonneau](https://github.com/datacarpentry/datacarpentry.github.io/issues/542#issuecomment-705450136). I was incorrect in my explanation above: the benefit of these changes is that it will allow the locally-built lesson pages to be used offline e.g. at workshops in locations with limited/unreliable Internet access.

gcapes:
HI @tobyhodges 
It looks more like your original explanation was correct to me. The lesson can currently be build locally with jekyll and it looks like all the pages work as expected. Some have URLs ending in /index.html and others don't, but they all render. So that benefit doesn't appear to be correct.

I see that this PR changes the URLs to have an extra /index.html for some pages, which supports your initial reasoning. *If* that is the motivation, I would think a similar change should be made to `reference.md`?

tobyhodges:
Without the Jekyll server running i.e. run `make site` and open `_site/index.md` in your browser, does the Instructor Notes page render correctly when you open it from the Extras dropdown? Here's what it looks like for me...

Screenshot 2020-10-08 at 12 18 51

tobyhodges:
The idea is that the `_site` folder could be built and taken to a workshop on a USB stick. This PR should solve this issue for locally-built pages.

gcapes:
Ah yes, I see the purpose now. I think `reference.md` needs updating in the same way, doesn't it?

tobyhodges:
Well spotted!

PR #1136, opened 2020-09-08
Update 03-create.md

d-ave-g:
Added lines 77 - 99 to introduce mkdir -p, along with a sample, as well as the introduction of the -R flag to ls to recursively list the newly created directory hierarchy beneath the thesis directory.  Added lines 877 - 879 to introduce an exercise scenario to highlight mkdir -p.  Added "the default behavior of" to line 894 to avoid leaving students with the impression that mkdir cannot create subdirectories.  Added lines 897-899 to provide a solution explanation for the mkdir -p exercise scenario.

Please delete this line and the text below before submitting your contribution.

---

Thanks for contributing! If this contribution is for instructor training, please send an email to checkout@carpentries.org with a link to this contribution so we can record your progress. You’ve completed your contribution step for instructor checkout just by submitting this contribution.  

Please keep in mind that lesson maintainers are volunteers and it may be some time before they can respond to your contribution. Although not all contributions can be incorporated into the lesson materials, we appreciate your time and effort to improve the curriculum.  If you have any questions about the lesson maintenance process or would like to volunteer your time as a contribution reviewer, please contact The Carpentries Team at team@carpentries.org.  

---

PR #1151, opened 2020-10-12
03-create: Check emptiness before creating subdir

encukou:
The `mkdir -p` example creates directories in `thesis/`.
It might be meant as a "what if?" section, rather than hands-on exercise, but if people follow it, `thesis/` won't be empty at this point.

I suggest moving the `ls -F` that checks if the dir is empty above this section.

There are further `ls thesis` commands in the file that don't list `chapter_1/`, but I think people can figure out what's going on.
The change will work even if the `mkdir -p` section is removed (a possibility I see mentioned in some past discussion).

---

It's just a suggestion; feel free to change/close the PR as you see fit :)

While I'm here: ♥ Thank you for making the materials available! I am translating and adapting them for [a Czech-language course about Linux basics](https://naucse.python.cz/2020/linux-admin/carpentry-unix/filedir/) – with attribution and thanks, of course – and I find them remarkably well written & thought out.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -75,6 +75,13 @@ creatures/  data/  molecules/  north-pacific-gyre/  notes.txt  pizza.cfg  solar.
 ~~~
 {: .output}
 
+Since we've just created the `thesis` directory, there's nothing in it yet:
+
+~~~
+$ ls -F thesis
+~~~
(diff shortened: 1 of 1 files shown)


gcapes:
Great, thanks for picking up on this. It looks like this error was introduced when merging #1136.

PR #1153, opened 2020-10-12
Update 07-find.md

cmroberts-como:
Should be in a yellow "try me" box, and not to be necessarily covered during lesson.


gcapes:
Thanks for the PR.
The `tree` command won't be available by default on all versions of Bash: it wasn't for me with Linux Mint. So while it's a nice tool to know about, I think it's beyond this novice lesson.

PR #1155, opened 2020-10-18
looping exercise based on the *.txt files

Vilma-Agalioti:
Topic: Exercise on looping that uses the files and commands that were demonstrated in the previous lessons.
Note: I used this when I taught bash about a month ago, feel free to incorporate, provide me with feedback to improve, or reject.

---

### DIFF
--- a/_episodes/05-loop.md
+++ b/_episodes/05-loop.md
@@ -771,24 +771,25 @@ so she decides to get some coffee and catch up on her reading.
 >
 > ~~~
 > # navigate to the folder data 
-> > my_counter=0 # set up a counter, avoid spaces, the name of the counter can be anything but needs to be consistent across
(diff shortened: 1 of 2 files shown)


gcapes:
Thanks for the PR.
This episode already has a lot of exercises, but none of them ask the learner to write a loop from scratch. I think it might be a good idea to replace one of the existing exercises, with one where they have to write a simple loop from scratch like your PR suggests. This might help the material stick a bit better.

I'm not sure about including a loop counter example too.

deppen8:
This one seems to have gone cold, so I am going to close. Happy to re-open if we want to work on it more.

For what it's worth, I agree with @gcapes that if we want to add more exercises, we should propose some others that we can drop.

Issue #1126, opened 2020-08-03
Add file extension to goostats and goodiff


drjonathanbradley:
Is there a reason the goostats and goodiff files are not named goostats.sh and goodiff.sh? Earlier in the lesson, we talk about the importance of having file extensions, and in the scripts section we emphasize using .sh for bash scripts, but we don't do that in the files we give learners. 

We've currently run into issues where learners think that goostats is a bash command itself and not a filename because it doesn't have a file extension.

I'm happy to put in a pull request to fix the files, but I wanted to make sure there wasn't a reason for doing it the current way before I did.

gcapes:
Thanks for the issue. I'm not sure of the answer. I can't see a good reason not to rename them. Those files are executable so that might be the reason, but given the lesson uses them like `bash goostats` and not `./goostats`, I guess they should have the executable permission removed and a .sh extension added.

gdevenyi:
I agree with @gcapes that since we don't touch executable bits or PATH and we emphasize good naming conventions, we should probably practice this properly.

tlestang:
> Is there a reason the goostats and goodiff files are not named goostats.sh and goodiff.sh?

Usually executable files don't have an extension, since it is irrelevant to the way they are executed. 
I thought that the point of having `goodiff` and `goodstats` as executables without an extension was to mimic the general case of processing data using a external executable, whether it be binary, bash, python.. Learners _can_ realize that this particular executable is, in fact, a text file, but are not meant to. From that perspective I would actually go the other way: keep them executable, add a shebang line `#!/bin/bash` and execute them with `./`.

Turning both goostats and goodiff into non-executable bash scripts is perfectly valid, but would perhaps be less illustrative of  interfacing a small bit of bash (looping over data files) with a black-box executable (some program written by the supervisor) which I think is the point of the activity _Nelle’s Pipeline: Processing Files_ in episode 5.

gcapes:
@tlestang 
Thanks for that. You make some good points. There's advantages to both options but running them as executables will likely throw up some questions whose answers would involve permissions and PATH which might derail the lesson. I guess the main point is to address the inconsistency in naming/permissions/method of running by choosing one of the two suggestions above. I'm leaning towards choosing the simple option of adding file extensions so that we can avoid having to cover more concepts.

Related to this, I think the current explanation of how to run a shell script with `bash my-script.sh` is a bit vague and could be improved by way of analogy to running a python script with `python my-script.py`.

[...]

PR #1156, opened 2020-11-23
Add file extension to goostats and goodiff

drjonathanbradley:
This pull request is in response to issue #1126 . I have added the .sh file extension to goostats and goodiff and replaced the zip file with one containing the updated files.


gcapes:
Thanks, this is a good start to fixing the issue. However, if we add a file extension, we would also need all the references to `goostats` and `goodiff` in the lesson notes (ep 1, 5, 6) to be changed to `goostats.sh` and `goodiff.sh`.

PR #1160, opened 2020-12-07
Specify the current directory for `grep -r` example, prevent hang.

nkicg6:
`grep` reads from stdin if no directory is specified. Running `grep -r Yesterday` hangs with the message: `grep: warning: recursive search of stdin`. Specifying `.` fixes. 
Run on macOS 10.14.6 GNU Bash 3.2.57(1)

### DIFF
--- a/_episodes/07-find.md
+++ b/_episodes/07-find.md
@@ -208,7 +208,7 @@ If we use the `-r` (recursive) option,
 Let's search recursively for `Yesterday` in the `data-shell/writing` directory:
 
 ```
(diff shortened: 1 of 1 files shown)


gdevenyi:
Thanks. To be clear, this is one of the differences that crops up between OSX's bsd-based utils, and the GNU utils that Linux uses. Linux assumes the `.` when using `-r`

We'll merge this since it makes things uniform interface-wise and learners have already heard about `.`

PR #1162, opened 2020-12-14
Note about the text cursor, and about pressing the enter key

chuseq:
Added a short note about the text cursor after the prompt. Sometimes we have found that some new learners don't infer what that rectangle is (maybe worth mentioning in the Background paragraph that in a GUI the graphical cursor is the mouse pointer? i don't know). Also added that they have to press the enter key to execute a command. 
Maybe it's obvious but for people really new to a shell can be helpful.

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -72,10 +72,10 @@ In the examples for this lesson, we'll show the prompt as `$ `.
 Most importantly:
 when typing commands, either from these lessons or from other sources,
 *do not type the prompt*, only the commands that follow it.
-Also note that after you type a command, you have to press the *Enter* key yo execute it.
+Also note that after you type a command, you have to press the <kbd>Enter</kbd> key to execute it.
 
 The prompt is followed by a **text cursor**, a character that indicates the position where your typing will appear. 
-The cursor is usually a flashing solid rectangle, but it can also be an underscore or a pipe. 
+The cursor is usually a flashing or solid block, but it can also be an underscore or a pipe. 
 You may have seen it in a text editor program, for example.
 
 So let's try our first command, `ls` which is short for listing.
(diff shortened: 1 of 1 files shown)


_episodes/01-intro.md
```diff
@@ -72,6 +72,11 @@ In the examples for this lesson, we'll show the prompt as `$ `.
 Most importantly:
 when typing commands, either from these lessons or from other sources,
 *do not type the prompt*, only the commands that follow it.
+Also note that after you type a command, you have to press the *Enter* key yo execute it.
```
gdevenyi, line 75: Can you please convert the Enter code to use the `<kbd>` syntax.

_episodes/01-intro.md
```diff
@@ -72,6 +72,11 @@ In the examples for this lesson, we'll show the prompt as `$ `.
...
+
+The prompt is followed by a **text cursor**, a character that indicates the position where your typing will appear. 
+The cursor is usually a flashing solid rectangle, but it can also be an underscore or a pipe.
```
gdevenyi, line 78: Perhaps here a "flashing or solid block".

I wonder if we can actually put in the actual text characters instead?

gdevenyi (changes requested):

gdevenyi:
Thanks for the suggestions, I like this addition, but I have a few changes to request.

_episodes/01-intro.md
chuseq: Done. Also corrected s/yo/to/.

_episodes/01-intro.md
chuseq: Done. I was looking for an ascii code or something for the block cursor but could not find any.

gdevenyi:
Thanks!

PR #1163, opened 2020-12-15
Mention bash author

chuseq:
Just mention bash author, Brian Fox
---


gdevenyi:
Thanks for the suggestion, but I don't think this falls into the scope of needed information for learners of shell-novice.

PR #1164, opened 2020-12-15
man pager key shortcuts case

chuseq:
Mention that the man command goes through a pager program, and change case of key shortcuts to avoid confusion


gdevenyi:
Thanks for your suggestion, however I don't think the information you've added is needed for learners to achieve the objectives in this section.

In addition, our style guide uses uppercase letters to reflect they keys on a standard keyboard, which are also uppercase.

Thanks again.

PR #1165, opened 2020-12-23
Fix #1116: Updated explanation and clarified the usage of `ls` with filename or directory as arguments

### LINKED ISSUE #1116: Ep3. mv ls output clarification
In the [Moving files and directories section](http://swcarpentry.github.io/shell-novice/03-create/index.html#moving-files-and-directories), `mv thesis/quotes.txt . ` moves this file to the current working directory. `ls thesis` would then have no output as the file was moved out of that directory. 
> Further, `ls` with a filename or directory name as an argument only lists that file or directory. We can use this to see that quotes.txt is still in our current directory: 
Bash ls quotes.txt 
Output quotes.txt"

As a new user, this was not quite clear to me. Does this mean that `quotes.txt` would not show up in the output if it was not in the current working directory? Maybe it would help to clarify this if this is the case. Possibly, `ls` with a filename or directory name as an argument only lists that file or directory in the output if it is contained in the current working directory. The sentence alone was a bit confusing.

niketagrawal:
Based on the suggestions made in the comments of #1116 , the steps to understand and verify the operation of moving a file between directories are updated to convey a better clarity in using `ls` with filenames or directories as arguments.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -346,9 +346,20 @@ $ ls thesis
 ~~~
 {: .language-bash}
 
-Further,
-`ls` with a filename or directory name as an argument only lists that file or directory.
-We can use this to see that `quotes.txt` is still in our current directory:
+Alternatively, we can confirm the file `quotes.txt` is no longer present in the `thesis` directory:
+
+~~~
+$ ls thesis/quotes.txt
+~~~
+{: .language-bash}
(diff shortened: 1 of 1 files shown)


gcapes:
Merged with edits in 1817437164afecac291e53575dd9a3b146ca8b0b

PR #1167, opened 2021-01-05
Add note to navigate to home directory

wrightrc:
I've had a learner whose shell's don't default to their home directory. We didn't catch this until `ls -F Desktop`, so I think they missed some of the material, or didn't grasp some of the details/useage of relative paths. 

I think it would be useful to mention explicitly that learners should navigate to their home directory if `pwd` returns something other than their home directory, both in "Home Directory Variation" and prior to the `ls -F Desktop` bash code chunk.

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -69,7 +69,7 @@ which is Nelle's **home directory**:
 {: .callout}
 >  
 > We will also assume that your `pwd` command returns your users home directory. 
-> If `pwd` returns something different you may need to navigate there using `cd ~` 
+> If `pwd` returns something different you may need to navigate there using `cd` 
 > or some commands in this lesson will not work as written. 
 > See [Exploring Other Directories](exploring-other-directories) for more details 
 > on the `cd` command.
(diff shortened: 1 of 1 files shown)


_episodes/02-filedir.md
```diff
@@ -67,6 +67,12 @@ which is Nelle's **home directory**:
 > In future examples, we've used Mac output as the default - Linux and Windows
 > output may differ slightly, but should be generally similar.
 {: .callout}
+>  
+> We will also assume that your `pwd` command returns your users home directory. 
+> If `pwd` returns something different you may need to navigate there using `cd ~`
```
gdevenyi, line 72: Hi, the bare "cd" command also navigates to the home directory, so there's no need for "~"

gdevenyi:
Good catch, I haven't seen that before (how did it happen?!). I think we can drop the tilde from the suggested command for simplicity.

_episodes/02-filedir.md
wrightrc: I agree, save any details here for after cd is introduces.

gdevenyi:
Thanks!

PR #1169, opened 2021-02-10
Fix different indentation length

ashki23:
There were different indentation styles in page 05-loop. I did cosmetic changes to make them uniform.


gdevenyi:
Great eye! Thanks!

PR #1173, opened 2021-02-13
Add more extensive alt tags to images

mpolitze:
Hi everyone!

While I went through a couple of issues for my checkout procedure I noticed from the issues in the python repository that more extensive "alt" descriptions for the images might also be helpful for this lesson; especially where folder structures are needed within the challenges.


gdevenyi:
Thanks very much for the contribution!

PR #1176, opened 2021-02-16
Added missing prompt character in Ep. 5

piperfw:


gdevenyi:
Thanks for the fix. I see that the very next block of code is missing the proper formatting for a for loop which the subsequent block gets. Can you fix that too?

piperfw:
Hi @gdevenyi, I did notice this but understood that the prompt (and `>` characters) had been _purposefully omitted_ from that block, to indicate that it was not code that should be explicitly run but pseudo code to explain the concept of a loop. Similar lack of formatting is used in the exercises Saving to a File in a Loop - Part One/Two as well as when explaining the `cp` command further down the page.

gdevenyi:
That makes sense, there should remain a distinction for pseudocode.

Issue #1175, opened 2021-02-14
2. Navigating Files and Directories: Truncate ls --help output


piperfw:
Is there a good reason for including the full output of `ls --help` in episode 2? 

I see little educational value of this and it makes the page considerably longer (I doubt the expectation is for learners to read the entire output; if they wanted to it would be better for them to run the command in their own shell where it can be combined with grep, less etc.).

I propose including the first few lines and then ellipses `⋮` or `...`

gdevenyi:
We would welcome a PR proposing a fix for this,

gcapes:
Resolved in #1178

PR #1178, opened 2021-02-18
02: Truncate ls --help output

piperfw:
RE issue  #1175

Truncate output of `ls --help` in episode 2, mirroring how `grep --help` is truncated in episode 7.

The choice of where to cut the output was made to (just) include the `-F` option, which learners will be familiar with from the previous episode.

This saves 91 lines.


gdevenyi:
Thanks for this!

Issue #1166, opened 2021-01-05
Lesson contribution – instructor checkout – Introducing the shell


Upe-ndra:
Hi, I have a suggestion about the key points for [Introducing the Shell](http://swcarpentry.github.io/shell-novice/01-intro/index.html) lesson.

I think some of the key points like: explaining read-run-print cycle, discussion about options (flags), demonstration of tab completion, identifying the command, options and filenames, mentioned at the end of this lesson are mostly covered in the next lesson [Navigating files and directories](http://swcarpentry.github.io/shell-novice/02-filedir/index.html). So maybe it will be better to move them there?

Thank you,
Upendra

piperfw:
I agree Upendra (and came here to raise an issue on the matter). Specifically, the **first four points should be removed** as they refer to material in later episodes (perhaps at one stage this content was in the first episode).

If an additional key point was added for make up for the removal of the first four, it could describe the role of scripts in reproducing sequences of commands, Bash as a particular example of a shell (and the de facto Unix shell), or simply what the prompt is.

gdevenyi:
We would welcome a PR demonstrating the proposed changes. Thanks.

PR #1179, opened 2021-02-18
01. Update key points

piperfw:
RE #1166.

Removed four key points that referred to material no longer covered in this episode.

Added two new key points reflecting the actual episode content - feedback on these would be welcome.


gcapes:
Merged with edits in 26b36731

PR #1183, opened 2021-02-24
Update 01-intro.md

tvsingh:
The definition of a shell program needed to be more clear.


gcapes:
Thanks for your PR.
Sometimes the clarity of an explanation beats accuracy. I'm afraid I think this is one such occasion.

PR #1193, opened 2021-04-10
Update 02-filedir.md

rahman9909:
added Linux


gcapes:
Thanks for the thought. However the point here is that both work for Linux, but only one or other on macOS or git bash

PR #1195, opened 2021-04-19
Remove mkdir -p example

### LINKED ISSUE #1187: Ep03: ls output incorrect
Another recent change that hasn't been checked thoroughly:

The `mkdir -p` example creates a load of subdirectories in the 'thesis' directory. This output isn't shown later on in the notes when listing the contents of the `thesis` directory. The notes show we just have `draft.txt`. This error propagates through to the 'Moving files and directories' section.

It's hardly the end of the world to have these extra subdirectories there, but the notes no longer match the output you get from running the commands, and they probably mainly serve to confuse the picture. Extraneous load and all that.

gcapes:
The subsequent output of `ls` should now be correct.

Fix #1187.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -82,28 +82,6 @@ $ ls -F thesis
 ~~~
 {: .language-bash}
 
-Note that `mkdir` is not limited to creating single directories one at a time. The `-p` option allows `mkdir` to create a directory with any number of nested subdirectories in a single operation: 
-
-~~~
-$ mkdir -p thesis/chapter_1/section_1/subsection_1
-~~~
-{: .language-bash}
-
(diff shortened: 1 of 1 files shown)


gremau:
An issue with this pull request is that, by removing the `mkdir -p` example, that option remains un-introduced for an exercise that comes later in the episode. Looking at the "Reproduce a folder structure" exercise (around line 876), it brings up the `mkdir -p` option again as the solution to the exercise. So, removing the example earlier in the episode, which is what this pull request does, means the exercise needs to be dropped or changed. I think maybe some wording changes earlier in the episode might solve the problem.

gcapes:
This is a good point, which I'm  going to address in a new PR.

PR #1197, opened 2021-05-10
Edit Ep3 to make directory contents consistent

### LINKED ISSUE #1187: Ep03: ls output incorrect

gremau:
* Output from the `ls` commands in the lesson did not show the
`chapter_1` directory or `my_file.txt` contents that were created in
`thesis/` earlier in the episode. Now they do.
* Edited some of the lesson text to clarify what is happening.
* fixes #1187

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -215,7 +215,7 @@ $ ls
 {: .language-bash}
 
 ~~~
-draft.txt
+chapter_1 draft.txt
 ~~~
 {: .output}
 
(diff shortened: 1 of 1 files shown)


gcapes:
Thanks for the PR. I agree it makes sense to keep an example of `mkdir -p`, but I'm not convinced by the thesis example.

I think the sub-directories are an unrealistic way to organise a thesis directory, and the extra directories clutter the following examples.
Perhaps it would be better to show an entirely separate example such as a new project directory which contains:
`data  results  src`

This then covers both objectives of introducing `mkdir -p` and not cluttering the output.

I could incorporate this idea into #1195

gcapes:
Incorporated into #1203

PR #1208, opened 2021-06-18
Including `cd -` as hint

### LINKED ISSUE #1154: Include an example for using 'cd -' for going to the previous directory
http://swcarpentry.github.io/shell-novice/02-filedir/index.html

Relevant section `Two More Shortcuts` shoutout

The explanation that exists on the page is correct but might not illustrate that it only keeps track of the previous one so running it twice doesn't go through your history, but instead puts you back where you were.

helenehoffmann:
fix #1154

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -516,6 +516,12 @@ equivalent to `ls -Fa`.
 > is used.
 {: .callout}
 
+> ## Returning to previous directory
+>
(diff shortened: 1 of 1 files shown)


gcapes:
Hi @helenehoffmann and thanks for the PR!
I like the idea of giving an example, but I think it should be an example that learners can run. We also need to improve the clarity of the original explanation as suggested in #1154.
I've just submitted a PR for discussion: #1209

gcapes:
Fixed in #1209

PR #1209, opened 2021-06-21
Explain `cd -` more thoroughly and add runnable example

### LINKED ISSUE #1154: Include an example for using 'cd -' for going to the previous directory

gcapes:
Fix #1154


gcapes:
Building on the ideas in #1208, I'm wondering if the example should live where I've put it, or as an exercise?  Given the callout material is optional I thought best to include in the callout itself.

adrem1 (approved):
I think this is a great addition to the lesson!

Issue #1202, opened 2021-06-11
add another solution to exercise 


rksin8:
Exercise [Tracking a Species ](http://swcarpentry.github.io/shell-novice/07-find/index.html) can have two solutions:

`grep -w $1 -r $2 | cut -d : -f 2 | cut -d , -f 1,3  > $1.txt`    (Given)

`grep -w $1 -r $2 | cut -d , -f 1,3  | cut -d : -f 2 > $1.txt`

As this is a rearranging commands exercise, a user may come with either of the solutions. It will be good if we could either add both the solutions or mention in the text that there could be more than one possible solution.

gcapes:
Looks good. PR welcome!

PR #1210, opened 2021-06-25
Mention alternative tracking a species solution

tcouch:
Suggested solution for #1202 . Leaves figuring out why it works either way as an exercise for the reader, rather than giving a potentially long winded explanation.


gcapes:
Great, thanks for the PR!

PR #1217, opened 2021-07-15
Updated keypoints 03-create.md

rajeevraz:
added about option i with the rm command in order to safely delete content.


gdevenyi:
Thanks for the suggested contribution!

I'm not sure that the `-i` option for rm is a key point in terms of how we think about key points in a lesson. Thoughts @gcapes ?

rajeevraz:
Hi Gabriel,

I could have actually put that in a different way. 
PFB the pull-request which actually speaks for my contribution.

https://github.com/swcarpentry/shell-novice/pull/1220

gcapes:
@gdevenyi 
I agree - this isn't a key point.

PR #1220, opened 2021-07-17
Updated a key point in 03-create.md

rajeevraz:
Added 'permanently' just to emphasize that anything deleted using rm in Linux deleted forever.


gcapes:
Thanks for the suggestion. However, this detail already has its own key point so I don't think we need to add it again here.

> - "The shell does not have a trash bin: once something is deleted, it's really gone."

gdevenyi:
I agree with @gcapes here, this topic is already covered.

Thanks for the suggestion!

PR #1223, opened 2021-07-28
fix errors in second episode

zkamvar:
- keypoints yaml item was missing because of a malformed tag
- A git merge conflict sign was present in a header


gcapes:
Thanks for this! I think these are both my screw ups from a recent merge.

zkamvar:
> Thanks for this! I think these are both my screw ups from a recent merge.

Hey now! There are no screw ups, only opportunities. I'm pretty dang sure that I've merged conflict signs and broken yaml before, so absolutely no biggie :smile:

gcapes:
> Hey now! There are no screw ups, only opportunities.

LOL! How very 'Carpentries' of  you :)

I guess this happens much more easily when I use the GitHub online editor. Lesson learned :)

PR #1224, opened 2021-07-29
Replaced species with alkanes.

Kauthar-Omar:
Updated a variable to a more suitable informative name.


gcapes:
Thanks for taking the time to contribute to the lesson. I'm not sure whether there's anything wrong with using `species` as the loop variable, but certainly `alkanes` is a downgrade in clarity. `alkane` however would be fine.

gdevenyi:
A `chemical species` is a thing, but `alkane` in the singular would also be correct but also more specific. I'm ambivalent about the change overall.

PR #1226, opened 2021-08-04
Add file to make sort example runnable

### LINKED ISSUE #810: Ep4 `sort -n` has different behaviour on Mac and git bash vs linux
The `-n` flag only works as expected on Linux. The other OSes give the same output as `sort`

We noticed this when @mawds was teaching using git bash.

gcapes:
- Update shell-lesson-data and zip file for download
- Link to file in episode 04

Fix #810.


deppen8 (approved):
Looks good to go. This will close quite a few issues 🎉 .

PR #1228, opened 2021-08-20
Change female pronouns to gender neutral pronouns

karenwill95:


deppen8:
Thanks for the PR, @karenwill95. GitHub tells me it is your first PR, so congrats! 🎉 🥳  

As I'm guessing you know, adopting inclusive language is an important part of the Carpentries pedagogy, so I think the PR looks good. I'll add @swcarpentry/shell-novice-maintainers here too for their input. 

Before we merge it though, if we are going to change the pronouns in episode 1, we should also change them throughout the lesson. A quick search suggests there are pronouns in episodes 1-6. That is a pretty big job, but happy to review your PR if you want to tackle it.

gdevenyi:
I agree with @deppen8 that if we're going to undertake this, we should do so all at once in one big change to maintain consistency.

milodubois:
I saw this was closed but thought I'd chime in (since I follow all convos here) and say this could be performed with a custom loop in Python or Java, passing the text files in and running through a REGEX processor. Additionally, we could scan for contextual language usage and dependencies to confirm that a proposed change is consistent with English grammar. Following, we could assign a score to the most easily fixed (the ones in which our algorithm does not find much ambiguity nor dependencies with other pronouns, such as possessive ones, and then approve the changes in batches from those that don't need review (easiest) to those that need human review. In this way, perhaps there may be four or five groups. If we assume a normal distribution, 20-25% of all the algorithm's suggestions for changes should theoretically be able to be automated and pushed without extensive human review. Simply a suggestion, sorry if this comment is not in the right place, but I want to second the importance of inclusivity and also applaud swcarpentry and contributors! In response to closed issue #1228 and in response to @deppen8, @gdevenyi, et al.

PR #1234, opened 2021-09-07
Avoid using bracket wildcard notation

### LINKED ISSUE #615: Better solution to "Wildcard expressions"

gcapes:
Fix #615.

### DIFF
--- a/_episodes/04-pipefilter.md
+++ b/_episodes/04-pipefilter.md
@@ -682,49 +682,9 @@ Since it's too late to get the information any other way,
 she must exclude those two files from her analysis.
 She could delete them using `rm`,
 but there are actually some analyses she might do later where depth doesn't matter,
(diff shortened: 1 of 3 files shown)


gdevenyi:
Definitely agree here. Cuts down the lesson and this is more complex than it needs to be.

gdevenyi (approved):

deppen8 (approved):

PR #1235, opened 2021-09-09
Remove mentions of orthogonality

### LINKED ISSUE #541: examples of "orthogonality" questionable
@gvwilson  A quibble - I don't believe that the examples of `.` and `..` [given here in lesson 2](https://github.com/swcarpentry/shell-novice/blame/8a135d15821510648314f90183ef7e650a04e0f0/_episodes/02-filedir.md#L578) and wildcard expansion [given later in lesson 4](https://github.com/swcarpentry/shell-novice/blame/e2f1725fe47172e5779f9ae1a7af3adda75fa1d8/_episodes/04-pipefilter.md#L98) are really examples of orthogonal design;  `.` and `..` are simply conveniences provided by the filesystem, and wildcards are a convenience provided by the shell. Orthogonal design is design for composability and lack of side-effects. I think the line "[they are interpreted the same way by every program](https://github.com/swcarpentry/shell-novice/blame/8a135d15821510648314f90183ef7e650a04e0f0/_episodes/02-filedir.md#L579)" is particularly problematic because it might imply that if you were to write a new program, you should handle `.` and `..` or `*` and `?` specially in your cli args, to maintain "orthogonal design". If you want to introduce the concept of orthogonality I believe it would be better off in the lesson on pipes, with the example of coreutils (mostly) being designed to use stdin and stdout streams in the same way, allowing pluggable combinations of things.

deppen8:
This PR closes #541 by removing mentions of "orthogonal design" in the lesson. The reasons for removing are:

- "orthogonal design" as an idea is not fleshed out
- Recognizing and giving names to software design patterns is beyond the scope of this lesson
- There is question about whether the examples are even _good_ examples of "orthogonal design" cf. #541 

(Edit: sorry, just now realizing I should have probably done this from a fork.)

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -516,19 +516,6 @@ equivalent to `ls -Fa`.
 > is used.
 {: .callout}
 
-> ## Orthogonality
->
-> The special names `.` and `..` don't belong to `cd`;
-> they are interpreted the same way by every program.
-> For example,
-> if we are in `/Users/nelle/data`,
-> the command `ls ..` will give us a listing of `/Users/nelle`.
-> When the meanings of the parts are the same no matter how they're combined,
-> programmers say they are **orthogonal**:
-> Orthogonal systems tend to be easier for people to learn
-> because there are fewer special cases and exceptions to keep track of.
-{: .callout}
-
 These three commands are the basic commands for navigating the filesystem on your computer:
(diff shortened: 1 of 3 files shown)


gcapes:
Great stuff, thanks! Yes, maybe from a fork is the 'ideal' method, but the main thing is to create a new branch as you did.

PR #1242, opened 2021-09-12
Setup suggestions: introduce Cygwin; introduce keyboard shortcut (checkout)

chenchenzi:
Hi lesson maintainers,

I made some small suggestions/changes for the setup.md page. 
1) I introduced Cygwin as an alternative UNIX emulator on Windows (which is extendable and versatile), under advanced users.
2) I intriduced a keyboard shortcut for instant access to Terminal on Ubuntu.

Best,
Chenzi.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -60,6 +60,8 @@ menu.
 
 **For advanced users:**
 
+Another good emulator that run in a Windows environment is [Cygwin][https://cygwin.com/install.html], which is free to install, relatively easy to maintain, flexible, and extendable.
+
 As an alternative to Git for Windows you may wish to [Install the Windows Subsystem for Linux][wsl]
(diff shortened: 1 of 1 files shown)


deppen8:
Thanks for these suggestions, @chenchenzi. 

On the topic of Cygwin, I think we want to keep the options limited in the setup. If we include Cygwin, I think that sets up some degree of expectation that instructors will be able to support it and that the lesson will be compatible. However, I don't think that is a promise we can make at this point.

On a similar note, as someone who doesn't use Linux regularly, I am wondering if that keyboard shortcut is widely applicable. If it only applies to Ubuntu, perhaps it is a bit too niche to include?

Happy to hear others' thoughts.

gcapes:
Thanks for the PR!
However, I agree with @deppen8 on both.

Issue #1238, opened 2021-09-10
Modernize README


deppen8:
The README is pretty sparse. Some other lessons have more details in their README that we could use for inspiration. 

In particular, I think it would be good to call out how folks can contribute and where to find good first issues. Happy to get other ideas too.

gcapes:
I looked at the python lesson's README. I think they've gone a bit overboard, but we should probably steal the section on  [using GitHub](https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/CONTRIBUTING.md#using-github).

I think this is something a lot of new contributors struggle with.

PR #1243, opened 2021-09-13
add section linking to CONTRIBUTING page

deppen8:
A quick PR to work towards #1238 (and test new GitHub Actions).


gcapes:
Looks good. Could you put in some line breaks please? Line 23 is really long.

Also, I find this construction rather clunky
> contribute to the continued development

Perhaps we could just drop 'continued'?

PR #1247, opened 2021-09-14
Make lesson check fail on error

### LINKED ISSUE #1245: Make lesson check stricter
There's little point in running GitHub actions to run `lesson-check-all` if it allows errors through (which it does). Mostly the output from this is fairly trivial, but at some point, an error was introduced which messes up the formatting:
```
./_episodes/02-filedir.md:71: Unknown or missing blockquote type None
```

I'll have a little look at the Makefile and report back. But I think we should fix all the little problems then we can keep on top of things. I'm happy to do this.

gcapes:
Fix #1245.


gdevenyi (approved):
After #1248

PR #1250, opened 2021-09-22
New image resized

deppen8:
A small PR to resize the image contributed in #1072


gcapes:
Hi @deppen8 
The image size looks ok, but there's loads of whitespace around it now. If you can fix that I think this is good to go!
image
Thanks!

deppen8:
@gcapes, I think this should be fixed now. Can you take a look and make sure it is fixed on your end too?

PR #1262, opened 2021-11-16
explaining environment dependency for getting help earlier

LunaSare:
I suggest explaining that the `--help` option and `man` are environment dependent earlier, so instructors and learners know what to expect from the beginning. 
It caught me off guard during teaching, and I did not realize that the explanation was given a bit further down until a few days later. During class I am usually too nervous and focused on keeping the class going, so I just told students I would investigate it further and talk about it during the next session, and we skipped to the next topic. 

Thanks!


gcapes:
Thanks for this PR. I'm making a few tweaks and will merge it shortly. :)

Issue #1264, opened 2021-11-17
 ls --help  is not available on Mac OS:   make it more visible by bold text


dhanyapushpadas:
[02-filedir.md](https://github.com/swcarpentry/shell-novice/edit/gh-pages/_episodes/02-filedir.md)
line 190 and line 196 , make the text in parentheses bold so that it is more visible to learners that` ls --help `is not available on macOS and `man ls`  is not available in Git Bash

mszarko:
In a recent Carpentries workshop I observed, this section still generated a lot of confusion for the Mac users (they either didn't see the bolded section, or didn't understand it). Could a note be added to explain what users on the respective operating systems should do instead? For example, line 189 could read something like:
 
1. We can pass a `--help` option to the command (not available on macOS; use man -ls instead), such as:

I definitely understand (and appreciate!) the effort to keep the script as simple as possible to avoid constant updating for every operating system. But I did make a note of this one for my own separate teaching notes, since it obviously still caused some confusion for several workshop participants.

gcapes:
I'd encourage this discussion as part of #798

PR #1265, opened 2021-11-17
Update 02-filedir.md

dhanyapushpadas:
Fixed issue #1264 [1264](https://github.com/swcarpentry/shell-novice/issues/1264)

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -187,13 +187,13 @@ are plain old **files**.
 to use a command and what options it accepts ---
 **depending on your environment, you might find that only one of these ways works:**
 
-1. We can pass a `--help` option to the command (not available on macOS), such as:
+1. We can pass a `--help` option to the command **(not available on macOS)**, such as:
     ~~~
     $ ls --help
     ~~~
     {: .language-bash}
 
-2. We can read its manual with `man` (not available in Git Bash), such as:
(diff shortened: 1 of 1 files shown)


gcapes:
Thanks for the PR. 

The thing with bold is that it's easily overdone. I think that with the warning above the bullets already bold, this extra bold isn't needed. 

Any thoughts?

deppen8:
@dhanyapushpadas, GitHub tells me it is your first GitHub pull request. Congrats! 🎉 And thanks for contributing to this lesson!

On the content of the lesson, I agree with @gcapes: the disclaimer is there in bold above and repeated again in normal text, which should be sufficient. 

Since you brought it to our attention though, I do find the current wording a bit awkward with the parenthetical breaking up the text in an odd place. Perhaps there is some other way to clean up the text and raise the disclaimer in a smoother fashion. I'd be happy to look at some suggestions, if you have any ideas.

dhanyapushpadas:
@gcapes @deppen8 Thank you for the comments, with further thinking I too agree with your points. I don't have any better suggestion,  but I have experienced that this is not visible enough to some learners. Probably by linking the text to the reason why it is not available or making a note , we can differntiate this?

deppen8:
I am closing this PR because of inactivity, but happy to re-open if there is further appetite for making changes here.

PR #1267, opened 2021-11-26
Consistently use option vice switch/flag

tbert:
It may be better to stick with a single term
consistently throughout the text, and mention
that alternate terms may be used by other
shell users in real-world settings.

Also use alternate wording to avoid using "options"
in two different contexts to further avoid confusion.

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -20,7 +20,7 @@ keypoints:
 - "`ls [path]` prints a listing of a specific file or directory; `ls` on its own lists the current working directory."
 - "`pwd` prints the user's current working directory."
 - "`/` on its own is the root directory of the whole file system."
-- "Most commands take options (flags) that begin with a `-`."
+- "Most commands take options that begin with a `-`."
 - "A relative path specifies a location starting from the current location."
 - "An absolute path specifies a location from the root of the file system."
 - "Directory names in a path are separated with `/` on Unix, but `\\` on Windows."
@@ -147,14 +147,13 @@ system and how you have customized your filesystem.)
 
 `ls` prints the names of the files and directories in the current directory.
 We can make its output more comprehensible by using the `-F` **option**
-(also known as a **switch** or a **flag**),
 which tells `ls` to classify the output
 by adding a marker to file and directory names to indicate what they are:
 - a trailing `/` indicates that this is a directory
 - `@` indicates a link
 - `*` indicates an executable
 
-Depending on your default options,
+Depending on your shell's default settings,
 the shell might also use colors to indicate whether each entry is a file or
 directory.
 
@@ -242,7 +241,7 @@ Mandatory arguments to long options are mandatory for short options, too.
 {: .output}
 
 > ## Unsupported command-line options
-> If you try to use an option (flag) that is not supported, `ls` and other commands
+> If you try to use an option that is not supported, `ls` and other commands
 > will usually print an error message similar to:
 >
 > ~~~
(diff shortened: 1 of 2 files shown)


_episodes/02-filedir.md
```diff
@@ -734,14 +733,18 @@ $ ls -F /
...
-We've already encountered options (also called **switches** or **flags**) which
+We've already encountered options  which
 either start with a single dash (`-`) or two dashes (`--`),
 and they change the behavior of a command.
 [Arguments] tell the command what to operate on (e.g. files and directories).
 Sometimes options and arguments are referred to as **parameters**.
 A command can be called with more than one option and more than one argument, but a
 command doesn't always require an argument or an option.
 
+Note that options are also called **switches** or **flags**, especially for options
```
gcapes, line 744: This wording implies that the terms switch and flag are on equal footing with option. I think we're trying to present 'option' as the preferred term to use (because that's what the man pages use), so I would reword to reflect this. Something like
> You might sometimes see  *options* being referred to as **switches** or **flags**, especially for options that take no argument. In this lesson we will stick with using the term *option*.

reference.md
```diff
@@ -120,6 +112,14 @@ operating system
 :   Software that manages interactions between users, hardware, and software [processes](#process).
     Common examples are Linux, macOS, and Windows.
 
+option
+:   A way to specify an argument or setting to a command-line program.
+    By convention Unix applications use a dash followed by a single letter,
+    such as `-v`, or two dashes followed by a word, such as `--verbose`,
+    while DOS applications use a slash, such as `/V`.
```
gcapes, line 119: I'm not sure we need to go into DOS in the Unix shell lesson.

reference.md
```diff
@@ -82,14 +82,6 @@ filter
     they read data from [standard input](#standard-input),
     process it, and write the result to [standard output](#standard-output).
 
-flag
-:   A terse way to specify an option or setting to a command-line program.
```
gcapes, line 86: This is mostly repeated from the option entry, but without a definition of the difference. I would probably drop this entry, and maybe mention in the option entry this alternative terminology.

gcapes (changes requested):
Thanks for the PR! I'm always keen on contributions which add clarity :)
This PR is good stuff, but I've a few small comments that need addressing please.

reference.md
tbert: This entry was renamed to 'option' and slightly cleaned of subjective ("terse way") wording.

reference.md
tbert: I more or less just moved the existing text; that change is, IMO, orthogonal to this diff.

_episodes/02-filedir.md
tbert: I've no objection to this change being made before merge, but doing so via the pull request resolution interface on github is, at present, beyond my ken.

reference.md
gcapes: Sorry, I was reading this with tired eyes and got confused, thinking there were not yet entries.

reference.md
gcapes: Yes, fair point!

_episodes/02-filedir.md
gcapes: Let me know if you'd like me to make that change, but the normal approach would be for you to make another commit on your feature branch and push it. That will then your the PR.

tbert: Follow-up commit pushed to branch in forked repo. Was afraid my reliance on command-line tools would force me to learn yet another web interface for a single action.

reference.md
tbert: As a follow-up, this may be a good opportunity to flag something for "good first commit" status for purposes of onboarding.

_episodes/02-filedir.md
gcapes: I've squashed and merge this PR now. Thanks again for the contribution!

PR #1272, opened 2021-12-15
Command expansion in loops callback

jtfield:
Hello Carpentry Maintainer Team!

I am making this PR to suggest adding a small code block at the end of the Shell-Novice lesson (07 Find). This is for the Carpentry checkout process. However, I thought it would be interesting to explain to new users that command expansions can be used in loops. I look forward to comments on if this suggestion is too complex for new users. Thanks for your time!

Cheers,
Jasper

### DIFF
--- a/_episodes/07-find.md
+++ b/_episodes/07-find.md
@@ -625,13 +625,13 @@ $ grep "FE" $(find .. -name "*.pdb")
 Packaging multiple commands inside a `$()` can also be used in loops.
 This might be useful if you have to perform many searches and want to automate the process.
 
-```
+~~~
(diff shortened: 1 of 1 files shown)


gcapes:
Thanks for the PR!

This looks like a nice example but as you noted, this is a novice lesson and it is already pushed for time. As such I'm not sure whether this is something we should add.

@swcarpentry/shell-novice-maintainers ?

gdevenyi:
I think I agree with @gcapes that we don't really have space for this. In addition, this kind of generation of looping is not best practice (you should use find's exec function for example)

PR #1275, opened 2022-01-06
Improve formatting, update paths/output for latest data

alex-ball:
This PR contains formatting fixes and improvements, and updates content to account for changes in the `shell-lesson-data` directory. I have tried to avoid any controversial changes. The specific things changed here are as follows:

- The paths given in the challenge "Find the Longest File With a Given Extension" in Episode 6 were out of date and have been updated.
- The outputs of `ls` have been updated to match what a learner would see after doing everything in order from a fresh copy of `shell-lesson-data`. The one exception is that I have ignored the `touch my_file.txt` command in Episode 3: that's a tricky one.
- In certain cases, it looked like one-line `ls` output had been manually split over two lines (entries ordered across then down); I have changed these cases to proper two-line `ls` output, that is, ordered down then across.
- I have added `$ ` at the start of a few lines that were meant to be run at the prompt, and removed it from an output line.
- I have cleaned the indentation of `for` loops so they all use four spaces as standard, as some tabs had crept in.
- Where an output block contained only errors, I have changed the format directive to `{ .error}`.
- Where bash script lines had been formatted as `{ .output}`, I have changed the directive to `{ .source}`. There is one case where I have changed `{ .language-bash}` to `{ .source}` (at the start of Episode 6, to match the examples that follow immediately after) but otherwise I have left those alone.
- I corrected some typos in "Reproduce a folder structure" in Episode 3 and added a missing space a bit further up (under "Moving files and directories").

Sorry for bundling all these changes up, but it seemed more efficient than submitting each change separately.

### DIFF
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -753,7 +753,8 @@ For example, `ls -s` will display the size of files and directories alongside th
 while `ls -S` will sort the files and directories by size, as shown below:
 
 ~~~
-$ ls -s Desktop/shell-lesson-data/data
+$ cd ~/Desktop/shell-lesson-data
+$ ls -s data
 ~~~
 {: .language-bash}
 ~~~
@@ -763,7 +764,7 @@ total 116
 ~~~
 {: .output}
 ~~~
-$ ls -S Desktop/shell-lesson-data/data
+$ ls -S data
 ~~~
 {: .language-bash}
 ~~~
--- a/_episodes/02-filedir.md
+++ b/_episodes/02-filedir.md
@@ -754,10 +754,19 @@ while `ls -S` will sort the files and directories by size, as shown below:
 
 ~~~
 $ ls -s Desktop/shell-lesson-data/data
+~~~
+{: .language-bash}
+~~~
 total 116
  4 amino-acids.txt   4 animals.txt   4 morse.txt  12 planets.txt  76 sunspot.txt
  4 animal-counts     4 elements      4 pdb         4 salmon.txt
+~~~
+{: .output}
+~~~
 $ ls -S Desktop/shell-lesson-data/data
+~~~
+{: .language-bash}
+~~~
 sunspot.txt  animal-counts  pdb        amino-acids.txt  salmon.txt
 planets.txt  elements       morse.txt  animals.txt
 ~~~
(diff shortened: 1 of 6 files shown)


_episodes/02-filedir.md
```diff
@@ -752,11 +753,11 @@ For example, `ls -s` will display the size of files and directories alongside th
 while `ls -S` will sort the files and directories by size, as shown below:
 
 ~~~
-$ ls -s Desktop/shell-lesson-data/data
+$ ls -s ~/Desktop/shell-lesson-data/data
```
gcapes, line 756: This should work either with or without the tilde. I don't feel particularly strongly one way or the other, but this code block mixes input and output so it would be good to separate the commands from the output.

gcapes:
This is great work, thanks! Some of the incorrect paths are my fault :/

gcapes (changes requested):
This is good work - one further improvement I think would sit nicely in this PR.

_episodes/02-filedir.md
alex-ball: Sorry, a controversial one found its way in.

From what I can see, the learner is in `~/Desktop/shell-lesson-data` at this point, or possibly `~/Desktop/shell-lesson-data/creatures`. The `ls Desktop/shell-lesson-data/data` command only works if the learner is in their home directory, so I added the tilde so it didn't matter where they were.

But since the learner isn't actually asked to run those commands themselves, I'm happy to revert these changes to keep the code block clean.

alex-ball: I must be having a dim day, I just saw what you meant. I can split the code up, no problem.

gcapes: Yes, you're right. My bad.

gcapes: No worries - I've had a few of those lately :)
After a quick review, it's not certain which directory the learner will be in at this point, so maybe a `cd` would be helpful. Either here, or before the next `Nelle` section.
But yes the main point I wanted to make was about splitting input and output blocks.

alex-ball: How would you feel about this?

```bash
$ cd ~/Desktop/shell-lesson-data
$ ls -s data
```
```bash
$ ls -S data
```

If they have gone off somewhere else in the meantime and run this, they are back in the right place for the `Nelle` section.

gcapes: Looks great!

gcapes (approved):

PR #1277, opened 2022-01-11
Setup.md: clarification about the Git Bash name

wmotion:
A novice approaching both Bash and Git during the Software Carpentries may think that the name 'Git Bash' means that Bash is a feature or a flavour of Git.
A comment at the point in which Windows users install the Git-based Bash emulator clarifies the relationship between these two pieces of software.
An actual question during a Software Carpentries session stimulated this edit.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -61,6 +61,10 @@ which gives you access to both Bash shell commands and Git.
 Once installed, you can open a terminal by running the program Git Bash from the Windows start
 menu.
 
+To avoid that the name 'Git Bash' results in confusion during the Software Carpentries, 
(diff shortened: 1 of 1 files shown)


wmotion:
Editing the name of the branch at the fork side caused the closure of this pull request.
Not as intended. The branch has been restored and the issue has been reopened.

gcapes:
Thanks for the contribution. This however feels more like a 'red sticky note' clarification than something we need in the setup. I think there's value in keeping the setup concise.

wmotion:
Point of conciseness taken. The same change is now proposed for the first episode in the passage where the Bash shell is first introduced. The new pull request is #1278.

PR #1295, opened 2022-04-29
Remove misplaced number

cknuepfer:
The number 358242 doesn't make any sense here.


gcapes:
Great catch! This one is my fault.

PR #1298, opened 2022-05-05
Remove note on WSL for Windows

dicksonlaw583:
Grace Fishbein and I were helping out at a Unix Shell workshop today, and the students using WSL bash were having a very hard time following along. Here are a number of incompatibilities that we noticed:

- The desktop directory is completely different
- Tab completion didn't work for everyone
- `nano` is not available
- `man` is not available

I would like to remove the note about using WSL as an alternative, in order to avoid such incompatibilities during lessons. By the virtue of the lesson, we cannot assume that they are the "advanced user" that the note assumes.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -60,14 +60,6 @@ which gives you access to both Bash shell commands and Git.
 
 Once installed, you can open a terminal by running the program Git Bash from the Windows start
 menu.
-
-**For advanced users:**
-
-As an alternative to Git for Windows you may wish to [Install the Windows Subsystem for Linux][wsl]
-which gives access to a Bash shell command-line tool in Windows 10.
-
-Please note that commands in the Windows Subsystem for Linux (WSL) may differ slightly
-from those shown in the lesson or presented in the workshop.
 </article>
 
 <article role="tabpanel" class="tab-pane" id="macos">


deppen8:
Thanks for the PR. It is always useful to get direct feedback from lessons when they are taught.

Before we drop the WSL option completely, I think this needs further exploration. I don't have access to a Windows machine, but I had a friend test with WSL2. `nano` and `man` were available there, so I think there is more complexity to it.

gcapes:
I think the problems reported might be a WSL vs WSL2 issue. If anything I'd encourage a move towards WSL because in doing do the lesson would avoid many of the problems associated with using Git Bash.

AndreasBilke:
In my experience a wsl(2) is way more comfortable for Windows user since this environment mirrors a real Unix (which is close to the lesson material).

milodubois:
Chiming in to gently nudge and agree with comment from @gcapes, after looking into WSL vs WLS2. Also, please excuse the perhaps uninformed or naive question, but why is Oracle VBox not suggested as the primary environment? I ask this as a data scientist (computational and molecular biology, host-pathogen interactions) and as someone who dabbles in developmental personal projects in the biopharma/genomics arena. Certain FOSS applications critical for comp & mol bio are only available for Linux systems and given the questionable reliability of Windows for this sort of work, perhaps others may have noticed this. I have personally benefited from a course, btw. 

Thank you all for the work you do!

Cheers!
@milodubois

bkmgit:
People using WSL should be able to use man and nano since the default installed Linux is Ubuntu https://docs.microsoft.com/en-us/windows/wsl/setup/environment  - nano may need a separate install.  Have not tested it myself though.  A virtual machine has more overhead than WSL so needs a reasonable high spec computer, especially if one is also doing video conferencing.  If it mostly causes confusion, it is ok to drop it and remain with GitBash which does not have `man` but does have `--help`.

@dicksonlaw583 What versions of Windows were used?

bkmgit:
Maybe more explanation similar to https://docs.pace.gatech.edu/gettingStarted/windows_subsystem_linux/ would be helpful to have.

dicksonlaw583:
@bkmgit 

They are all Windows 10, though I did not have time during the sessions in May to ask them which subversion it is.

One persistent issue is the desktop directory on Windows, which for many people with institutional machines (but not all) are completely different from the Unix-standard `~/Desktop` and the two Window-standards `/c/Documents and Settings/USERNAME/Desktop` (XP) and `/c/Users/USERNAME/Desktop` (7, 8, 10, 11) expected in [unit 2](https://swcarpentry.github.io/shell-novice/02-filedir/index.html). Desktops mounted on network drives are a frequent sighting, and there would be an institution-specific prefix that must be found for each participant in this situation and what it maps to in the shell before being able to proceed. Sometimes we are able to get by if it's Git Bash (create a directory, double-click it, right-click > Git Bash, then `pwd`), but we didn't have established routines for easily determining it on WSL as none of the instructors have used it.

bkmgit:
I think this section should be update to encourage use of WSL2, and encourage instructors or those setting up the environment to test things out beforehand.

bkmgit:
Closing this, will revisit later. https://github.com/swcarpentry/shell-novice/issues/1379

bkmgit:
Need to close unresolved pull requests to update to new format.

PR #1300, opened 2022-05-24
Remove extra ')' in 07-find.md

### LINKED ISSUE #1294: A small change in the episode "Finding Things"
This correction is for the content in Episode 07 - https://swcarpentry.github.io/shell-novice/07-find/index.html

In the  "Matching and Subtracting" section, the following command:
  `grep -v "unicorn" $(find creatures -name "*.dat"))`
produces an error:
   `-bash: syntax error near unexpected token `)'`

As a lesson contribution for instructor training, I would like to edit it to remove the extra ')' so the answer provided in the Solution area will match the expected output.

</details>

Trombach:
closes #1294


deppen8:
Thanks @Trombach !

PR #1301, opened 2022-05-25
Add shortcut key to open a Terminal in Linux

MGomezN:
Add shortcut key to quickly open a terminal in Linux Ctrl+Alt+T
<details>
<summary><strong>Instructions</strong></summary>

Thanks for contributing! :heart:

If this contribution is for instructor training, please email the link to this contribution to
checkout@carpentries.org so we can record your progress. You've completed your contribution
step for instructor checkout by submitting this contribution!

Keep in mind that **lesson maintainers are volunteers** and it may take them some time to
respond to your contribution. Although not all contributions can be incorporated into the lesson
materials, we appreciate your time and effort to improve the curriculum. If you have any questions
about the lesson maintenance process or would like to volunteer your time as a contribution
reviewer, please contact The Carpentries Team at team@carpentries.org.

You may delete these instructions from your comment.

\- The Carpentries
</details>

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -95,6 +95,7 @@ The default Unix Shell for Linux operating systems is usually Bash.
 On most versions of Linux, it is accessible by running the
 [Gnome Terminal][gnome-terminal] or [KDE Konsole][kde-konsole] or [xterm][xterm],
 which can be found via the applications menu or the search bar.
+By default, to quickly open a Terminal in Ubuntu and Linux Mint, press Ctrl+Alt+T
 If your machine is set up to use something other than Bash,
 you can run it by opening a terminal and typing `bash`.
 </article>


deppen8:
Hi @MGomezN , thanks for contributing to the lesson. 

I think Linux users make up a very small proportion of the learners for this lesson. As such, I am afraid that adding this shortcut is likely to add more confusion for people who are just getting started.

deppen8:
It can be difficult, but we have to try to think of it from the perspective of an absolute beginner. Many of the learners will not be familiar with the terms "Ubuntu" and "Linux" and most won'thave a clear mental model of what they are. So when they encounter them in the setup instructions, they will wonder if those instructions apply to them and might try to use that shortcut. When it doesn't work for the vast majority of people, they could get discouraged, frustrated, or confused.

This is not what we want right as learners first encounter the lesson. And since this is in the setup instructions, there won't be an instructor on hand to help them through it. All of that leads me to believe that the benefit of the shortcut does not outweigh the possible confusion it could spark.

In more general terms, this lesson has been around for a long time and that means it has a lot of content. We have to have a pretty strict threshold for adding new pieces to the lesson.

MGomezN:
Thanks for the explanation, @deppen8. I will consider your observations in my further contributions.

PR #1305, opened 2022-06-01
07 consistency; Leah's project / Tracking a Species

nbehrnd:
This is to i) increase consistency between the test data provided, and the
representation of typical data when introducing the project.  For ii), to
put more emphasis on that individual commands presented to construct the
command in question are still to be rearranged by the reader.


gcapes (changes requested):
I like this idea - thanks for the PR.
However, the result for rabbit would now give three lines: 
```
2012-11-05,22
2012-11-06,19
2012-11-07,16
```

While we're improving clarity, I think that the 'species.txt' explanation is a bit lacking and would be better as  `<species>.txt`. Would you like to make that change too?

nbehrnd:
@gcapes Indeed, there was an accidental omission by mine (the line).  Your suggest to highlight the flexibility of the permanent record written indeed equally is included now.

PR #1307, opened 2022-06-03
Fix solution code in episode Pipes and Filters > Pipe Reading Comprehension

adogb:
This PR fixes an error in the solution code in the section on Pipe Reading Comprehension in the Pipes and Filters episode. Since the last command sorts the data in animals.csv in reverse order, it should be deer, then rabbit, then racoon. Not rabbit, deer, racoon.


gcapes:
Thanks for the PR!
Does your machine give a different answer to the one provided? When I run the example, the I get the same as given in the solution.

gcapes:
No worries :)

PR #1313, opened 2022-06-22
Fixes #1312: adding bash programming howto as follow up reference

### LINKED ISSUE #1312: Extras: Add manual link for advanced bash scripting guide
Hello

I suggest under the "Manuals" section of Extras to add a link to the ["Advanced Bash Scripting Guide" ](https://tldp.org/LDP/abs/html/) from [The Linux Documentation Project](https://tldp.org/), which I have suggested to people for many years and is always well received as a reference. I think it's probably the most comprehensive guide for aspiring shell programmers and, by linking to the project site in general, it'll open doors for exploration of further materials. 

Of course, absolutely happy to raise the PR should people be in agreement 😄 

Thanks!

JimCircadian:
Will email to checkout@carpentries.org, but will also look through the issues too :)

### DIFF
--- a/reference.md
+++ b/reference.md
@@ -242,6 +242,7 @@ wildcard
 ### Manuals
 * [GNU manuals](http://www.gnu.org/manual/manual.html)
 * [Core GNU utilities](http://www.gnu.org/software/coreutils/manual/coreutils.html)
+* [BASH Programming - Introduction HOW-TO](https://tldp.org/HOWTO/Bash-Prog-Intro-HOWTO.html)
 
 ### Miscellaneous
 * [North Pacific Gyre](http://en.wikipedia.org/wiki/North_Pacific_Gyre)


bkmgit (approved):
Thanks.

deppen8:
It is new to me so I just skimmed it, but I like that guide, @bkmgit . 

My one hesitation: Is there any chance it is out of date compared to the lesson? Unless I am reading it wrong, it was last updated in 2000.

JimCircadian:
@deppen8 by comparison to the advanced bash scripting guide it's out of date, but the coverage of this simpler guide and the lessons is such that there won't have been breaking changes in the last 22 years. The subject matter for bash users presented in both is fundamental functionality that hasn't changed. 

That said, it could be an argument for putting both in, and certainly it won't cover recent features of bash! Could be worth having a look at #1312 for context...

gcapes (commented):
A few thoughts:
1. This isn't a link to a manual, so it shouldn't go in a 'manuals' section.
2. The GNU manuals link might be better swapped out for the bash manual link: https://www.gnu.org/software/bash/manual/bash.html
3. What is the thought behind this new reference? Is it an alternative approach to the same concepts covered in the SWC course? Or to go beyond? There's a good bit of overlap, but there are some things presented but not explained, like the shebang in a script: ttps://tldp.org/HOWTO/Bash-Prog-Intro-HOWTO-2.html, and of course the script file must be executable to run in the way shown, but this isn't covered in either our lesson or the reference. So my first impressions are that it's of questionable value.

JimCircadian:
@gcapes 

1. I'm not sure I agree that this isn't a manual, but that's an argument based on semantics. 😉 It's aimed at providing information on "how to write intermediate shell scripts", which is what a manual is by definition. Originally I was after the advanced shell scripting guide which is better maintained and a very useful reference manual, but was deemed too much extra information. All told, these (or rather the ABS I originally proposed) have been useful references I've used over a number of years, so I thought they'd be helpful: certainly users should become reasonably able over time to use references such as this, hence proposing to put them under extras.
2. That is also a very good reference, I've not used it but certainly makes more sense. Happy to swap it out.
3. My thought is that the "Extras" section is a place to point students to further materials, so adding in useful sources of material  past the end of the course would be welcome.

I very much accept that the references I proposed might not be the best, so happy to amend to the GNU manual. Please let me know if you'd prefer this.

Thanks

JimCircadian:
Hello @bkmgit to avoid delaying any other work I've closed this, it was only a small change.

bkmgit:
Ok. Will open a new issue to resolve this after migration to the new
format. Thanks for the contribution.

PR #1322, opened 2022-07-31
Corrected typo in 02-filedir.md that led to `No such file or directory`

majona-garidi:
Corrected typo and clarified formulation of the lesson a bit. Before, output would show
`ls: north-pacific-gyre/G: No such file or directory`


gcapes:
Hi, thanks for the PR. 

The letters on a physical keyboard are always labelled as capitals, so we render them as such in the lessons. A capital letter would be shown as `shift` + `G`.

Issue #736, opened 2018-03-09
Episode 3: No mention of letter case for file names and directories


jfessic:
Hello,
In "Working with Files and Directories" under "Good names for files and directories" I think it would be good advice to suggest that file names contain only lower case letters.  If nothing else, it would be worthwhile to emphasize the importance of being consistent with letter case when choosing file names. 
 My understanding is that Unix is always case sensitive, while Windows is not, and macs depend on how the OS was configured when installed.  It can be very puzzling for someone who is not aware of this behavior and doesn't realize the different effects it can have on their files.  It is also good advice to develop a habit for how you will name files and stick with it.

shwina:
I am +1 for this addition. Just a comment: some "built-in" directories like `Desktop` and `Documents` do *not* follow these conventions. Perhaps any explanation should also account for this.

loz-hurst:
> My understanding is that Unix is always case sensitive

This depends on the filesystem being used.  By default Mac OSX is NOT case sensitive, despite being an actual Unix (bearing in mind Linux is not a Unix but a clone).

Richard-Regan:
OK, I will. Is it good practise for me to fork the repo first, and then clone my copy before I start?

bkmgit:
For text changes that are not large, forking the repository, making a branch in your fork, adding your changes to the branch and making a pull request are fine.  You can clone to your fork to your local machine or use the web interface.

Richard-Regan:
Thanks

Richard-Regan:
Hi, Just created a pull request. If any issues please contact me.

PR #1323, opened 2022-08-10
Added a tip to use lowercase letters

Serpens:
I added a new tip, as suggested in this issue: https://github.com/swcarpentry/shell-novice/issues/736

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -156,6 +156,13 @@ data/  results/
 > If you need to refer to names of files or directories that have spaces
 > or other special characters, you should surround the name in quotes (`""`).
 {: .callout}
+>
+> 4. Use only lowercase letters.
+>
+>    Depending on your operating system and file system, upper- and lowercase letters can be
+>    distinguished or treated as the same (Linux in general is case-sensitive, but macOS and
+>    Windows are not). The easiest way to avoid confusion is to stick to lowercase letters.
+
 
 ### Create a text file
 Let's change our working directory to `thesis` using `cd`,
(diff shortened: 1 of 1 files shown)


bkmgit (commented):
Maybe this could be given as a more general warning.  Typically, it is good to create a separate directory to work in, and keep other existing directories the same.

Serpens:
I can add that, but wouldn’t that be too much? The box is about names, not directory structure.

knarrff:
I don't quite agree. Being consistent with lower/upper case is what is key IMHO. Don't name a file Readme.md and later refer to it as readme.md or README.md, even if your OS allows for that. Other than that I don't think we should tell people to limit their choices here. Not only do most OS environments already create content that does not follow this, even the popular git hosters do, if you let them, pre-populate new repositories with Readme.md or README.md files, and files like LICENSE.md and CONTRIBUTING.md that the Carpentries use themselfes are case-senstitive.

You even get into problems when you think about programming languages that connect internal data structure with file names. A prime example is Java. In Java, file names of files containing classes should match the class name, and at classes should commonly use camel case naming, in particular should start with capital letters. This puts a requirement on capital letters also within file names, at least if you want the code to work on OSs that respect lower/upper case.

gcapes:
My preference is to use lowercase as much as possible, mainly because it makes the typing easier, so I like this tip. However, I don't think it's universal enough that we can suggest it without the caveats in https://github.com/swcarpentry/shell-novice/pull/1323#issuecomment-1354328164, which would then undermine the advice, and make the entry rather too long.

Issue #1320, opened 2022-07-28
Update maintainer list


gcapes:
Remove previous maintainers from README

PR #1324, opened 2022-08-10
Update maintainers

bkmgit:
https://github.com/swcarpentry/shell-novice/issues/1320


gcapes (changes requested):
We don't list previous maintainers, so 'current' is redundant.
Otherwise, all good - thanks!

bkmgit:
Should we list previous maintainers somewhere?

gcapes (approved):

Issue #1270, opened 2021-12-06
Ep05_"general form of a loop"_won't_give_output


dechavezv:
The Unix Shell, Episode 05 line 48 (https://swcarpentry.github.io/shell-novice/05-loop/index.html).

I just think that It may be useful to add a note explaning that the general for loop (see below) should give an erro as "thing" and "ist_of_things" do not exist in our working directory. 

for thing in list_of_things
do
    operation_using $thing    # Indentation within the loop is not required, but aids legibility
done

Best,

kieren-j-mitchell:
I was also momentarily confused about this when preparing to teach this episode. I'd also add that "operation_using" is not a valid command (causes "-bash: operation_using: command not found" error in terminal on MacOS).

For me, the main issue causing confusion was that the example of the general form of a loop was provided in a purple "Bash" box, which implies that it's meant to be entered and run (in pretty much every other case throughout the lesson the content of the Bash boxes are meant to be entered). I think it might be useful to either add a red "Error" box afterward (as dechavezv suggested) or alternatively present the general form as an image (or otherwise format it differently somehow) to make the distinction clear.

gcapes:
Thanks for the issue. I think this explanation can be clarified further by amending the introductory wording to make it more explicit that this is pseudocode, rather than a working Bash example. While the intention isn't for learners to attempt to run this pseudocode, I think the bash syntax highlighting is the most appropriate for the code block. Perhaps something like this:

> We’ll use a loop to solve this problem, but first let’s look at the general form of a loop **using the pseudo-code below**.

PR #1325, opened 2022-08-15
Explicitly state psuedo-code in ep 05

jamesacris:
In [Ep. 05: Loops](https://swcarpentry.github.io/shell-novice/05-loop/index.html), added the text 'using the pseudo-code below' to clarify that the code block is not expected to be run, rather to illustrate the concept of a loop.


jamesacris:
Addressing #1270

gcapes:
Great stuff, thanks!

PR #1335, opened 2022-10-17
Grammar Correction on Line 126

cgmerrick:
Lines 126-127 previously read, "Because Nelle is the user in our examples here, therefore we get '/Users/Nelle' as our home directory. Grammatically, it is better to either remove "Because" or "therefore." I removed "Because."


bkmgit:
Thanks.

bkmgit:
Pipeline failed because an older version of the template was used.

PR #1363, opened 2023-02-13
Do not overwrite a file that is used later

bkmgit:
Make a copy of the file.

### DIFF
--- a/_episodes/04-pipefilter.md
+++ b/_episodes/04-pipefilter.md
@@ -286,7 +286,8 @@ the output of `head` must be the file with the fewest lines.
 > to the same file. For example:
 >
 > ~~~
-> $ sort -n lengths.txt > lengths.txt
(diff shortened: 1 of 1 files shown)


gcapes:
Thanks for the PR. 
This is a callout showing what not to do. So I'm afraid that changing the example to a better practice negates the point of the callout.

emcaulay (commented):
This edit to the episode 4 addresses a concern raised and discussed in issue #1375 . I recommend accepted and merging these changes in. 

I just realized this PR is already closed, so my comment here is after the fact!

emcaulay:
Now, I'm seeing that this PR was not accepted, but I do think it is a good revision and addresses the concern discussed in #1375. I think it seems a little cruel to set up a learner to destroy a file by mistake when we expect them to have the intact file for continuing the lesson.

PR #1364, opened 2023-02-13
Fix typo and history error

bkmgit:
- use up instead of left arrow key to get previous command
- use history that is similar to what one would obtain based on following these instructions

### DIFF
--- a/_episodes/05-loop.md
+++ b/_episodes/05-loop.md
@@ -630,7 +630,7 @@ $ for datafile in NENE*A.txt NENE*B.txt; do echo $datafile stats-$datafile; done
 ~~~
 {: .language-bash}
 
-Using the left arrow key,
+Using the up arrow key,
 Nelle backs up and changes the command `echo` to `bash goostats.sh`:
 
 ~~~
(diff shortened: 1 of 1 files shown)


_episodes/05-loop.md
```diff
@@ -630,7 +630,7 @@ $ for datafile in NENE*A.txt NENE*B.txt; do echo $datafile stats-$datafile; done
 ~~~
 {: .language-bash}
 
-Using the left arrow key,
+Using the up arrow key,
```
gcapes, line 633: She's already pressed the up arrow key. At this point, she's making edits, using the left arrow key.

gcapes (changes requested):
History edits look fine, but the arrow key presses were already correct.

_episodes/05-loop.md
bkmgit: Thanks. Updated it, tried to clarify.

gcapes (changes requested):
Perhaps this would read more easily as
> Nelle navigates to the `echo` command and changes it to `bash goostats.sh`

Issue #1360, opened 2023-02-09
Protiens directory not accurately named


shaw2thefloor:
I don't know if this is just me being a pedant (very likely), but it has always bothered me, that the directory is named "protiens", but contains files of molecules, not proteins.

gcapes:
A PR would be welcome to fix this!

gcapes:
Fixed in #1367

PR #1367, opened 2023-02-16
Use alkanes instead of proteins

bkmgit:
@shaw2thefloor Addressing https://github.com/swcarpentry/shell-novice/issues/1360


gcapes:
The zip file of data which learners download in the setup section would need to be updated too. The last target in the Makefile will help with this.

gcapes (changes requested):
This looks ok but still needs the zip file updating (see last rule in the Makefile)

bkmgit:
Thanks. Updated zip file.

Issue #1356, opened 2023-01-23
[Navigating Files and Directories] small issue with wording 


AvRaJya:
Small issue with wording or maybe placement of the command thing.
The command is below ,but the sentence says that it is above.

![Screenshot_2023-01-23_00-28-27](https://user-images.githubusercontent.com/123229902/213973726-66799684-a2ed-427c-a861-fa56403df1eb.PNG)

https://swcarpentry.github.io/shell-novice/02-filedir/index.html#two-more-shortcuts

Not sure if this is something too minor to matter/bring up here. It's my first "contribution" so I'm not familiar with what types of problem/issues I should be bringing attention to.

bkmgit:
Thanks. Nothing is too minor. Can you make a pull request with a fix?

dimitrasal:
@bkmgit @AvRaJya I believe the word "above" in this case refers to use of the command higher in the text, however it can be a bit confusing. I have created a [PR](https://github.com/swcarpentry/shell-novice/pull/1373) to clarify better.

bkmgit:
@dimitrasal Thanks. @AvRaJya please reopen if further improvements are needed.

PR #1373, opened 2023-03-02
explains better word 'above' in episode 02

dimitrasal:
This fix explains better the phrasing in episode 02 as referred to in #1356


bkmgit:
Thanks.

PR #1383, opened 2023-04-24
fix HTML figure to use markdown syntax

zkamvar:
This will help with the transition to the workbench.

See https://github.com/carpentries/lesson-transition/issues/80


bkmgit:
Merging, but may need to resize figure once in new format.

PR #1389, opened 2023-05-03
Misspelled recursive flag

isepec:
In section "Copying files and directories" flag for recursive option was misspelled in two places. It said "-r" instead of "-R"!


isepec:
Lapsus calami! Closed the PR because my correction was, in fact, incorrect.

Recursive option for "cp" command can actually be both "-r" and "-R".
"ls" command is where "-r" and "-R" mean different things (reverse and recursive, respectively).

Note to myself: When in doubt, use "--help" option for checking possible options for a specific command!

PR #1392, opened 2023-05-09
Add `-n` switch to grep command

jsteyn:
Without the -n switch on grep to include line numbers on the result, the `cut -d : -f 2` of the script serves no purpose

### DIFF
--- a/episodes/07-find.md
+++ b/episodes/07-find.md
@@ -331,7 +331,7 @@ sequence in one command in order to achieve Leah's goal:
 cut -d : -f 2
 >
 |
-grep -w $1 -r $2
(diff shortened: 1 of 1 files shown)


bkmgit:
Thanks for the pull request. It does not seem to match the output shown in https://github.com/swcarpentry/shell-novice/blob/main/episodes/07-find.md?plain=1#L321-L324 the `-n` gives line numbers in which the text string is found. Maybe something else is needed? If so, one would also need to update https://github.com/swcarpentry/shell-novice/blob/main/episodes/07-find.md?plain=1#L350-L352

PR #1393, opened 2023-05-16
Update 03-create.md - Case sensitivity

Richard-Regan:
Added a Call out section on case sensitivity.

### DIFF
--- a/episodes/03-create.md
+++ b/episodes/03-create.md
@@ -148,19 +148,10 @@ expected and can even result in data loss.
 If you need to refer to names of files or directories that have spaces
 or other special characters, you should surround the name in quotes (`""`).
 
-::::::::::::::::::::::::::::::::::::::::::::::::::
+Whether or not your Bash shell is case-sensitive is normally determined by the underlying operating system. In most implementations of Bash it is case sensitive. On MacOS, however, it is not case sensitive.
 
-:::::::::::::::::::::::::::::::::::::::::  callout
+It can be a good practice to use all lowercase for files and directories, but this is not a rule; you are likely to see uppercase used in practice. This effect is negligible if the letter isn't the first letter in the name, so it's not an issue if you use naming standards such as camelCase.
 
-## Case Sensitivity
-
-This is normally driven by the underlying OS. In most implementations of BASH it is case sensitive, with the exception of MacOS which is case insensitive. Some OSs have predefined directories that use a capital letter at the start, such as Download, and Desktop. This is used to make these directors stand out.
-
-It is standard practice to use lowercase for your files and directories, but this is only a guideline.
-
-If you choose to use capital letters at the start of your file names, you should bear in mind the fact that sorting order in Unix is (sometimes) case sensitive, and some commands such as grep you may need additional options to achieve the desired result.
-
-Also this effect is negligible if the letter isn't the first letter in the name, so it's not an issue if you use naming standards such as camelCase.
 
 
 ::::::::::::::::::::::::::::::::::::::::::::::::::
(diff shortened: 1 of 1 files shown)


bkmgit:
Thanks. It is a little long and there may be differences between different versions of operating systems. Mac OS has recently switched to zsh as default, though the lesson mostly works ok.  Is it possible to reduce it to one or two sentences at most without its own section.  The lesson is already quite long and many instructors have trouble finishing.

episodes/03-create.md
```diff
@@ -148,6 +148,21 @@ expected and can even result in data loss.
 If you need to refer to names of files or directories that have spaces
 or other special characters, you should surround the name in quotes (`""`).
 
+::::::::::::::::::::::::::::::::::::::::::::::::::
+
+:::::::::::::::::::::::::::::::::::::::::  callout
+
+## Case Sensitivity
+
+This is normally driven by the underlying OS. In most implementations of BASH it is case sensitive, with the exception of MacOS which is case insensitive. Some OSs have predefined directories that use a capital letter at the start, such as Download, and Desktop. This is used to make these directors stand out.
```
deppen8, line 157:
```suggestion
Whether or not your Bash shell is case-sensitive is normally determined by the underlying operating system (Windows, MacOS, Linux, etc.). In most implementations of Bash, it _is_ case sensitive. On MacOS, however, it is not case sensitive.
```

Richard-Regan:
Hi,

Not a problem. Would you prefer it to be just an addition the "Good names for files and directories"?

Best Wishes
Richard
________________________________

episodes/03-create.md
```diff
@@ -148,6 +148,21 @@ expected and can even result in data loss.
...
+
+It is standard practice to use lowercase for your files and directories, but this is only a guideline.
+
+If you choose to use capital letters at the start of your file names, you should bear in mind the fact that sorting order in Unix is (sometimes) case sensitive, and some commands such as grep you may need additional options to achieve the desired result.
+
+Also this effect is negligible if the letter isn't the first letter in the name, so it's not an issue if you use naming standards such as camelCase.
+
+
```
deppen8, line 165:
```suggestion
```

Some of this is rolled into my previous suggestion and the camelCase bit is probably too much for this early in the lesson.

bkmgit:
That would be really great.  Thanks for your patience and perseverance.

deppen8 (changes requested):
I think this addition is a bit too much for this early part of the lesson. I made some suggestions that might help it be a little more succinct.

Richard-Regan:
Hi I have amended this addition, by removing the section and adding the reduced text to the section on "Good names"

Richard-Regan:
Hi All,

I am concerned I have missed a step with the revised pull request. Can someone confirm they see the updated lesson with the text about case sensitivity added to the section above about naming files and directories, and not in its own section. As this was done over 24h ago and I have not seen any comments.

Best Wishes
Richard
________________________________

bkmgit:
New version seems mostly ok for me.  Lines should be broken at 80
characters. Is it case insensitive for all versions of Mac OS?

gcapes:
Before merging this, can we please confirm the case-sensitivity on Windows? I thought that Windows was the outlier here not being case sensitive, while linux and macos were. But after a google search it seems that indeed macos is case insensitive too, so it's just linux which is case sensitive(?)

I know it's not really the OS, but rather the filesystem used, but I'm talking default behaviour here.

omsai:
This PR seems to be for #736. I still see multiple issues with the patch; for example the shell is not case sensitive, it's the file system that is. Additionally, the 80 character line wrapping previously requested has not yet been applied. But in any case it is unreasonable to ask the contributor to revise this PR for a 3rd time and rebase after such a long time so I opened a new PR #1499 and added the [contributors and coauthors](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) manually in the commit to get credit for their work.

PR #1396, opened 2023-05-18
Added a note about switching shells to default bash in MacOSX

jweekley-ucsc:


bkmgit:
@jweekley-ucsc Thanks for this and sorry for the delay in responding. It partially addresses https://github.com/swcarpentry/shell-novice/issues/1433 Happy to incorporate it, though maybe it is better added to https://carpentries.github.io/workshop-template/install_instructions/#shell see  https://github.com/carpentries/workshop-template/blob/gh-pages/_includes/install_instructions/shell.html#L124

PR #1414, opened 2023-07-16
updated path in 05-loop.md to be relative to home directory

theabro:
Currently, in the "Nelle's Pipeline: Processing Files" section of [episodes/05-loop.md](https://github.com/swcarpentry/shell-novice/blob/61534b49425a60bbfd78cfab5427a27b6bf91a3c/episodes/05-loop.md?plain=1#L597C13-L597C13) (line 597), the text reads

> Starting from her home directory, Nelle types:

however, the [code snippet](https://github.com/swcarpentry/shell-novice/blob/61534b49425a60bbfd78cfab5427a27b6bf91a3c/episodes/05-loop.md?plain=1#L600) just below this (line 600) then reads:

```bash
cd north-pacific-gyre
```

This path is incorrect, as if starting from the users home directory this line should read

```bash
cd Desktop/shell-lesson-data/north-pacific-gyre
```

### DIFF
--- a/episodes/05-loop.md
+++ b/episodes/05-loop.md
@@ -594,9 +594,10 @@ Since she's still learning how to use the shell,
 she decides to build up the required commands in stages.
 Her first step is to make sure that she can select the right input files --- remember,
 these are ones whose names end in 'A' or 'B', rather than 'Z'.
-Starting from her home directory, Nelle types:
+Moving to the `north-pacific-gyre` directory, Nelle types:
 
 ```bash
+$ cd
 $ cd Desktop/shell-lesson-data/north-pacific-gyre
 $ for datafile in NENE*A.txt NENE*B.txt
 > do
--- a/episodes/05-loop.md
+++ b/episodes/05-loop.md
@@ -597,7 +597,7 @@ these are ones whose names end in 'A' or 'B', rather than 'Z'.
 Starting from her home directory, Nelle types:
 
 ```bash
-$ cd north-pacific-gyre
+$ cd Desktop/shell-lesson-data/north-pacific-gyre
 $ for datafile in NENE*A.txt NENE*B.txt
 > do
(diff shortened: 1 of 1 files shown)


bkmgit:
Thanks for the contribution. The home directory may not be Desktop, so probably something further is needed.  Typically in bash one can check the output of `echo $HOME` to see what this is set to.

bkmgit:
Not sure why, it uses `starting from her home directory`. The section before that begins ["
We would like to modify each of the files in `shell-lesson-data/exercise-data/creatures`,"](https://github.com/swcarpentry/shell-novice/blob/61534b49425a60bbfd78cfab5427a27b6bf91a3c/episodes/05-loop.md?plain=1#L518) so maybe one could use relative paths
```
cd ../../north-pacific-gyre
```
and change the text "starting from her home directory" to "moving to the north-pacific-gyre directory" or something similar.

Alternatively, use
```
cd
cd Desktop/shell-lesson-data/north-pacific-gyre
```
Since this will then move to the home directory

bkmgit:
Can look at other changes for consistent paths in a further change since the home directory maybe different, and changes will be needed through the entire lesson.

theabro:
In 0b441e9e94b7b027b849a016a4e7e5279a88c828 I have amended the descriptive text to reference the directory moving to, and put in a `cd` to the home directory prior to the other `cd` command. Personally I preferred this over a relative path, in case the learner/instructor has somehow moved out of the previous directory. 

I agree that consistent paths could be an issue throughout the whole lesson. Currently it seems consistent on a Mac (except for this section), but any system where Desktop is not in the home directory will have problems. There may not be an ideal solution to this.

bkmgit:
@theabro Thanks. Created another issue to update paths https://github.com/swcarpentry/shell-novice/issues/1419.

PR #1432, opened 2023-08-10
Move callout on long vs short options

### LINKED ISSUE #1429: Consider moving callout about long/short options
The [callout about long/short options](https://swcarpentry.github.io/shell-novice/02-filedir.html#when-to-use-short-or-long-options) appears near the end of  the [Navigating Files and Directories lesson](https://swcarpentry.github.io/shell-novice/02-filedir.html). However long/short options are not actually introduced here.  It may help to move this callout earlier to the man/help section, where we see both long and short options in the man/help results, so we can note that they both mean the same thing and help the learner consider which to use.

bkmgit:
Place it where long and short options are first introduced.

Closes #1429

### DIFF
--- a/episodes/02-filedir.md
+++ b/episodes/02-filedir.md
@@ -263,6 +263,18 @@ Mandatory arguments to long options are mandatory for short options, too.
 
 :::::::::::::::::::::::::::::::::::::::::  callout
 
+### When to use short or long options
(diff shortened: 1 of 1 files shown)


ErinBecker:
Thank you @bkmgit!

PR #1438, opened 2023-10-17
Replace link to regular expressions lesson

jsteyn:
The link to the regular expressions lesson is broken. Replaced with https://librarycarpentry.org/lc-data-intro/01-regular-expressions.html


bkmgit:
Thanks. Build error seems unrelated.

PR #1443, opened 2023-11-09
Correct formatting

cknuepfer:
The variable name was not marked as verbatim which leads to wrong formatting of the following text.


bkmgit:
Thanks for the pull request. Please also raise an issue.
Line length seems too long, so it will not pass formatting check.
Please keep line lengths consistent with the rest of the file.

bkmgit:
@cknuepfer  Thanks.

PR #1460, opened 2024-07-23
Use `q` instead of `Q` to quit the man pages in 02-filedir.md

krishnakumarg1984:
Use `q` instead of `Q` to quit the man pages in `02-filedir.md`


bkmgit:
Thanks

PR #1467, opened 2024-09-25
Make printed output consistent with what is obtained

### LINKED ISSUE #1466: 05 - Loops wildcard matching
### What is the problem?

image
In this particular piece, my shell seems to print all files ending in A followed by files ending in B, instead of the alternate A's and B's in the output here. Not sure if it is shell dependent but a note might be helpful if it is.

### Location of problem (optional)

https://github.com/swcarpentry/shell-novice/blob/main/episodes/05-loop.md

bkmgit:
_If this pull request addresses an open issue on the repository, please add 'Closes #NN' below, where NN is the issue number._

Closes #1466

_Please briefly summarise the changes made in the pull request, and the reason(s) for making these changes._

Output shown to reader is not consistent wiht what is obtained

### DIFF
--- a/episodes/05-loop.md
+++ b/episodes/05-loop.md
@@ -605,10 +605,11 @@ $ for datafile in NENE*A.txt NENE*B.txt
 
 ```output
 NENE01729A.txt
-NENE01729B.txt
 NENE01736A.txt
+NENE01751A.txt
+
 ...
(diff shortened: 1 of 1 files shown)


sushma-4:
Looks good, thank you!

Issue #1398, opened 2023-05-22
Add OneDrive/OS info to inline instructor notes


ndporter:
The [instructor notes](https://swcarpentry.github.io/shell-novice/instructor/instructor-notes.html) have excellent detail on managing the complexities of finding desktop directories in various OS's with and without OneDrive, but I didn't realize those had been added. It would be great if someone took the time to convert those to inline notes in the [Navigating Files and Directories](https://swcarpentry.github.io/shell-novice/02-filedir.html) lesson.

murraycadzow:
Is it worth modifying the episode and the notes to mention people they can find the directory located on their desktop, then 'drag and drop' it into their terminal and it will automatically put the path in - this gets around pretty much all the different configuration issues and confusion that happens at the start of the lesson. Gets around the multiple OneDrives and the use of spaces in many of the enterprise onedrives.

I also believe OneDrive is now the default for Windows 11 users.

sstevens2:
Does the drag and drop option work for all OS's @murraycadzow?

jas58:
I can copy the absolute file path on a Windows machine.  

![Image](https://github.com/user-attachments/assets/cb9c9d16-ec38-4a4e-b208-efb2c2122622)

PR #1480, opened 2025-02-05
Moving onedrive solution inline

sstevens2:
This should solve issue #1398. I moved this to a callout as I think it is useful for many learners as well and then put part in an instructor note since it seemed more specific for helpers/instructors.  Feedback appreciated!

@jas58 brought this up in the Midwest Carpentries Community call so I was motivated to get this issue worked on.

### DIFF
--- a/episodes/02-filedir.md
+++ b/episodes/02-filedir.md
@@ -405,6 +405,35 @@ Note that if a directory named `Desktop` does not exist in your current working 
 this command will return an error. Typically, a `Desktop` directory exists in your
 home directory, which we assume is the current working directory of your bash shell.
 
+:::::::::::::: callout
+
+### OneDrive Variations
+
+If a Windows machine is backed up with OneDrive, your Desktop may
+  be rendered from a folder within OneDrive, which will not match the contents of `~/Desktop`.
+  The OneDrive desktop should be accessible using one of the following commands
+  (if the name of the enterprise/org isn't clear, look through the output of `ls` to find
+  the right folder):
+  
+  ```bash
+  $ cd "~/OneDrive - Name Of Enterprise/Desktop"
+  $ cd "C:/Users/Username/OneDrive - Name Of Enterprise/Desktop"
+  ```
+
+Note: The quotation marks are needed since the Unix shell interprets spaces to be meaningful separators between files/commands/options.
+
+::: instructor
+
+  One way to spot if the computer is using this kind of configuration is to look at files,
+  folders or links on the desktop. Usually the icon contains a shortcut/arrow symbol if it
+  is a link, or just the plain icon if the file is just saved in the `Desktop` folder.
+  Files synced with OneDrive contain an additional symbol indicating the sync status
+  (typically blue arrows for 'sync pending' or a green tick for 'synced').
+
+::::::::::::::
+
+::::::::::::::::::::::
+
 Your output should be a list of all the files and sub-directories in your
 Desktop directory, including the `shell-lesson-data` directory you downloaded at
 the [setup for this lesson](../learners/setup.md). (On most systems, the
--- a/episodes/02-filedir.md
+++ b/episodes/02-filedir.md
@@ -410,16 +410,23 @@ home directory, which we assume is the current working directory of your bash sh
 ### OneDrive Variations
 
 If a Windows machine is backed up with OneDrive, your Desktop may
-  have been moved within OneDrive and you will get an error when you try `ls -F Desktop`.
-  The OneDrive desktop should be accessible using one of the following commands
-  (if the name of the enterprise/org isn't clear, look through the output of `ls` to find
-  the right folder):
-  
-  ```bash
-  $ cd "~/OneDrive - Name Of Enterprise/Desktop"
-  $ cd "C:/Users/Username/OneDrive - Name Of Enterprise/Desktop"
-  ```
+have been moved within OneDrive and you will get an error when you try `ls -F Desktop`.
 
+To check if this is true for you, run the command `ls` and look through the output for a folder that has "OneDrive" in the name and note that folder name. It will likely include the name of your organization in it.  It may be something like `OneDrive - University of Example`. 
+  
+If you don't see a "OneDrive" folder, you may not be in your home folder or there may be another issue going on.  To check you are in your home folder, type `pwd` and it should show something like `C:/Users/Username` or `/c/Users/Username`.
+
+If you do see the "OneDrive" folder, going forward whenever the lesson or instructor uses the `Desktop` folder, you will need to include the OneDrive folder as well. For example if the instructor types...
+
+``` bash
+ls -F Desktop
+```
+
+You will need to type:
+
+```bash
+ls -F "OneDrive - Name Of Enterprise/Desktop"
+```
 Note: The quotation marks are needed for this folder because the Unix shell interprets spaces to be meaningful separators between files/commands/options but this folder has spaces in the name.
 
 ::: instructor
--- a/episodes/02-filedir.md
+++ b/episodes/02-filedir.md
@@ -424,7 +424,7 @@ Note: The quotation marks are needed since the Unix shell interprets spaces to b
 
 ::: instructor
 
-  One way to spot if the computer is using this kind of configuration is to look at files,
+  One way to spot if the computer is using this kind of configuration with OneDrive is to look at files,
   folders or links on the desktop. Usually the icon contains a shortcut/arrow symbol if it
   is a link, or just the plain icon if the file is just saved in the `Desktop` folder.
   Files synced with OneDrive contain an additional symbol indicating the sync status
--- a/episodes/02-filedir.md
+++ b/episodes/02-filedir.md
@@ -410,7 +410,7 @@ home directory, which we assume is the current working directory of your bash sh
 ### OneDrive Variations
 
 If a Windows machine is backed up with OneDrive, your Desktop may
-  be rendered from a folder within OneDrive, which will not match the contents of `~/Desktop`.
+  have been moved within OneDrive and you will get an error when you try `ls -F Desktop`.
   The OneDrive desktop should be accessible using one of the following commands
   (if the name of the enterprise/org isn't clear, look through the output of `ls` to find
   the right folder):
@@ -420,7 +420,7 @@ If a Windows machine is backed up with OneDrive, your Desktop may
   $ cd "C:/Users/Username/OneDrive - Name Of Enterprise/Desktop"
   ```
 
-Note: The quotation marks are needed since the Unix shell interprets spaces to be meaningful separators between files/commands/options.
+Note: The quotation marks are needed for this folder because the Unix shell interprets spaces to be meaningful separators between files/commands/options but this folder has spaces in the name.
 
 ::: instructor
 
--- a/episodes/02-filedir.md
+++ b/episodes/02-filedir.md
@@ -425,7 +425,7 @@ ls -F Desktop
 You will need to type:
 
 ```bash
-ls -F "OneDrive - Name Of Enterprise/Desktop"
+ls -F "OneDrive - University of Example/Desktop"
 ```
 Note: The quotation marks are needed for this folder because the Unix shell interprets spaces to be meaningful separators between files/commands/options but this folder has spaces in the name.
 
--- a/instructors/instructor-notes.md
+++ b/instructors/instructor-notes.md
@@ -221,23 +221,6 @@ as long as learners using Windows do not run into roadblocks such as:
   so that they can find it easily
   and watch it evolve.
 
-- If a Windows machine is backed up with enterprise OneDrive, their GUI desktop may
-  be rendered from a folder within OneDrive, which will not match the contents of `~/Desktop`.
-  The OneDrive desktop should be accessible using one of the following commands
-  (if the name of the enterprise isn't clear, look through the output of `ls` to find
-  the right folder):
-  
-  ```bash
-  $ cd "~/OneDrive - Name Of Enterprise/Desktop"
-  $ cd "C:/Users/Username/OneDrive - Name Of Enterprise/Desktop"
-  ```
-  
-  One way to spot if the computer is using this kind of configuration is to look at files,
-  folders or links on the desktop. Usually the icon contains a shortcut/arrow symbol if it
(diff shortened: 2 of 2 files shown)


jas58:
I might change the language for a learner to view (review) on their own, and me as a novice instructor.

episodes/02-filedir.md
```diff
@@ -405,6 +405,35 @@ Note that if a directory named `Desktop` does not exist in your current working
 this command will return an error. Typically, a `Desktop` directory exists in your
 home directory, which we assume is the current working directory of your bash shell.
 
+:::::::::::::: callout
+
+### OneDrive Variations
+
+If a Windows machine is backed up with OneDrive, your Desktop may
+  be rendered from a folder within OneDrive, which will not match the contents of `~/Desktop`.
```
sstevens2, line 413:
```suggestion
  have been moved within OneDrive and you will get an error when you try `ls -F Desktop`.
```

episodes/02-filedir.md
```diff
@@ -405,6 +405,35 @@ Note that if a directory named `Desktop` does not exist in your current working
...
+  The OneDrive desktop should be accessible using one of the following commands
+  (if the name of the enterprise/org isn't clear, look through the output of `ls` to find
+  the right folder):
+  
+  ```bash
+  $ cd "~/OneDrive - Name Of Enterprise/Desktop"
+  $ cd "C:/Users/Username/OneDrive - Name Of Enterprise/Desktop"
+  ```
```
sstevens2, line 421:
```suggestion
  Run the command `ls` and examine the output, looking for a folder that has "OneDrive" in the name and note that folder name. It will likely include the name of your organization in it.  It may be something like `OneDrive - University of Example`
  Going forward, whenever the lesson or instructor uses the `Desktop` folder, you will need to include the OneDrive folder as well. For example if the instructor types...

``` bash
ls -F Desktop
```

You will need to type:

```bash
ls -F "OneDrive - Name Of Enterprise/Desktop"
```

episodes/02-filedir.md
```diff
@@ -405,6 +405,35 @@ Note that if a directory named `Desktop` does not exist in your current working
...
+
+Note: The quotation marks are needed since the Unix shell interprets spaces to be meaningful separators between files/commands/options.
```
sstevens2, line 423:
```suggestion
Note: The quotation marks are needed for this folder because the Unix shell interprets spaces to be meaningful separators between files/commands/options but this folder has spaces in the name.
```

episodes/02-filedir.md
sstevens2: If the language in the above looks okay, I'll fix it not in a github comment.  It isn't interpreting my suggestion correctly

sstevens2:
@jas58 I rewrote the text a bit to try to address the issues you noted above. Please let me know if you see additional changes that should be made.

bkmgit:
Thanks for adding this.  What has this been tested on?

bkmgit:
Maybe it would be helpful not to work in the Desktop directory, but instead create a directory under $HOME wherever that is.

This is a larger change and could follow up this one.

jas58:
Thanks, Benson, We are trying to guide newer instructors (like me) or
novices reviewing the lesson on *how* to find home under Microsoft's new
OneDrive problem. How might you address that or write it out in case you
lost your voice during that part of the lesson?

bkmgit:
The text is ok. Am wondering if it is possible to avoid the problem by putting the files somewhere else on the local machine rather than in cloud storage. Do not use One Drive, so am not able to test it out. 

Maybe something like [rclone](https://rclone.org/) maybe better?

jas58:
No one wants to use OneDrive but many instructors are dealing with it
without clear guidance. These are novice machines often without admin
privilege and day 1 & 2 of their first workshop. I resist jumping to
complicated advance computer science fixes.  I would rather identify a
teaching moment to empower the learner and the Macintosh attendees with
greater awareness of how and why to understand their device and
troubleshoot independently later , too.  But I'll defer to Sarah.

Adamtaranto:
I agree with @bkmgit , best to suggest a solution that sidesteps the whole Onedrive issue.

@jas58 can you please test `echo $HOME/Desktop` or `echo ~/Desktop` from the GitBash shell on a machine with this Onedrive backup issue?

If that works, maybe just add an instructor note and avoid a tangent that won't apply to most students.

> Thanks, Benson, We are trying to guide newer instructors (like me) or novices reviewing the lesson on *how* to find home under Microsoft's new OneDrive problem. How might you address that or write it out in case you lost your voice during that part of the lesson?
>> On Sat, Feb 8, 2025, 7:48 AM Benson Muite ***@***.***> wrote: Maybe it would be helpful not to work in the Desktop directory, but instead create a directory under $HOME wherever that is. This is a larger change and could follow up this one. — Reply to this email directly, view it on GitHub

bkmgit:
Does the problem also occur on Windows Subsystem for Linux? It would be helpful to have this fixed upstream as well.

Ideally, the problem is avoidable, but if not a discussion on remote storage is probably needed. Many people doing research would benefit from cloud backups of their data. Can test out other solutions, using a buggy implementation is not a long term solution, though temporary fixes ate useful to ensure the lesson can be used.

jas58:
Yes, replacing the GitBash install with WSL is under exploration for a long
term fix to the Git workshop as a whole,  but that is very long term.  I'm
trying to help instructors who've struggles for 14 months with this
Microsoft enforced "feature". :sad face

sstevens2:
Putting the lesson data files in a different folder could be a long term solution.  However, we might want to think carefully about it. In my experience, being able to navigate to your desktop (a location most users are very familiar with) is very useful and an important learning opportunity for learners that helps them connect that the locations they are seeing in the unix shell are the same locations they see in file explorer/finder.

Perhaps this is related to my institution's adoption of one drive, but in the workshop I taught in Jan, more than half of the windows users had this One drive issue.  It is possible to work around it by talking through it with learners as I described in this PR, and it helps learners know that onedrive is changing their file structure. I could move it to an inline instructor note as mentioned in the original issue if you prefer. Given @jas58 and my experience teaching with it I felt a callout could be useful for instructors. As a callout, it could still be skipped by instructors who don't feel it is needed but will be available for all instructors and for folks working through this lesson outside a workshop.

Switching to WSL2 is a big change that I don't think we are ready for at the moment.  When I've tried to use it instead of gitBash in past workshops, there were various installation issues that learners ran into, some of which required updating bios settings. gitbash has worked much better for us.

episodes/02-filedir.md
```diff
@@ -405,6 +405,42 @@ Note that if a directory named `Desktop` does not exist in your current working
...
+If you do see the "OneDrive" folder, going forward whenever the lesson or instructor uses the `Desktop` folder, you will need to include the OneDrive folder as well. For example if the instructor types...
+
+``` bash
+ls -F Desktop
+```
+
+You will need to type:
+
+```bash
+ls -F "OneDrive - Name Of Enterprise/Desktop"
```
sstevens2, line 428:
```suggestion
ls -F "OneDrive - University of Example/Desktop"
```

ndporter:
I may come back to try and improve the language, but for now I'll leave a couple of comments based on my experience teaching this 3-4 times a year:
1. This is a common issue for Windows users. In my setting (a University with most faculty using managed computers where they are not admin, with Microsoft services campuswide), typically around 50% of Windows users find that the files they save on their desktop end up in OneDrive
2. I think it's useful to leave this in (e.g. not just move to a directory under `Home`/`~` for the reasons stated above - that the learners will likely need to know about this quirk if they're going to successfully navigate the command line on their computers.
3. (this is the one that would lead to updating language) - It is possible to have at least 4 different Desktop folders without realizing it: a local user Desktop, a local public Desktop (e.g. "all users"), a `OneDrive/Desktop` folder, and a managed `OneDrive - [institution]/Desktop` folder. In order to effectively help learners find their data, it would be valuable to explicitly have all the options or at least examples of what they look like in a callout and explain that you may have any or all of them on your computer.

As an aside, I didn't realize about the syncing marks, which I'll remember for myself in the future.

sstevens2:
Thanks @ndporter!  I'd really appreciate your review of the language and suggestions on how to address point number 3 in your list.

jas58:
Thanks, @ndporter @sstevens2 for the suggested language.
I'm unaware of this "icon" referenced.  Can it be described? Or its location in Explorer or the terminal?
Is it the 3-hue blue cloud on the left of the word OneDrive?
Otherwise, the language as an instructor callout box will be an improvement.  
If an image helps to explain the process or visual to copy a filepath, feel free to add the image from my comment in https://github.com/swcarpentry/shell-novice/issues/1398

@bkmgit @Adamtaranto I'm sorry I was short-tempered this past weekend.  I hadn't realized the different experiences/histories between maintainers and instructors.  Thank you for your work to keep the lessons clean and clear.

bkmgit:
Are there similar problems with [msys2](https://www.msys2.org/) or [cygwin](https://cygwin.com/)? Msys2 has an option to install [Konsole](https://apps.kde.org/konsole/) as a terminal emulator, but as many people will likely also use git, being able to install git and use git from the terminal is useful.

It would be good to figure out where the OneDrive problem comes from so it can be fixed.

Do many people use the portable version of Git for Windows, or is this something most institutions will install for all learners? If WSL setup is problematic, can leave it as an option, but suggest something else as default.

bkmgit:
It appears Cygwin was used around 2013, then a switch to msysgit/gitbash occurred. Mysys2 probably has better path management, git and other linux tools can also be installed in it. Cygwin has more tools, but is bigger, not sure about differences in user experience at present.

If there are multiple Desktop folders, and remote syncing probably adding an exercise/discussion on remote storage is needed. If we want most people to have a similar experience, may need to use something other than/in addition to OneDrive. 

Typically advise people not to use Desktop when using this lesson, a users home directory is better. Navigating to the Desktop is a useful skill.

bkmgit:
To clarify, am ok merging this, though would much prefer to use $HOME for the majority of the lesson and have an exercise discussing desktop and remote storage. Can do migration to using $HOME and merge this as a stop gap measure for the time being.

If there are opinions based on recent experiences using other options for using linux tools on windows, it would be helpful to have these. Maintenance of gitforwindows is also a volunteer project, but if many users have trouble with how OneDrive is handled, it would be good to raise an issue upstream

Adamtaranto:
@sstevens2 @jas58 Can one of you please test `echo $HOME` and echo `~/` from on a machine with this Onedrive desktop issue? 

If possible please test from GitBash shell and from a WSL session. 

We are reviewing the general lesson setup instructions and it will be useful to know how Onedrive affects the different options.

sstevens2:
Let me know if I'm missing replying to something.  I'll try to cover a couple of things I noted from the convo above.

I don't think this is related to gitbash/cygwin/etc.  It is that in the windows file system One Drive moves the desktop folder into a subfolder of itself, instead of it being directly in the home folder.  I suspect in WSL2 it won't be an issue since they are separate partitions with their own home folders (at least last I checked).

I'm a bit hesitant to use the $HOME option at this point because it brings up system variables when learners are concentrating on learning how to use commands for the first time and how to navigate their file system.

I can't test it as I don't have or want to have onedrive setup on my personal computer, and my primary work computer is a mac.  I know the solution I proposed from helping many people through this issue in workshops.

@jas58 I think the icon mentioned is in the file explorer but I can't check either.  We can update the language to say the "icon in the file explorer" if that helps? I didn't write that text (I took it from the instructor notes) and it was a bit harder for me to evaluate as I've not looked at one drive in the file explorer much.

ndporter:
> @sstevens2 @jas58 Can one of you please test `echo $HOME` and `echo ~/` from on a machine with this Onedrive desktop issue?
> 
> If possible please test from GitBash shell and from a WSL session.
> 
> We are reviewing the general lesson setup instructions and it will be useful to know how Onedrive affects the different options.

In Git Bash, both commands just return the user home directory (`/c/Users/ndporter` for me). 

That directory contains both `Desktop` and `OneDrive/Desktop` (and would contain a third desktop in `One Drive - Virginia Tech - ndporter` or something like that if I had triggered the creation of that version.

In WSL, both commands return `/home/ndporter` (my username), though the system loads to my user directory `/mnt/c/Users/ndporter`. There is no OneDrive folder of either sort that appears.

Practically, it appears that WSL runs in `//wsl$` (which can only be found in Windows by knowing the address or looking in "Network" (not "This PC").  

So if we use WSL's home directory to solve the OneDrive issue, we basically have to teach people to treat WSL like a separate computer (and also get them to download the data into WSL, which Windows may not readily expose, especially if network neighborhood is turned off - as it is for my work computer).

bkmgit:
@ndporter Thanks for checking what WSL returns for $HOME.

jas58:
I'm adding two screenshots.  Maybe I didn't type in the right thing.
StubbsCommandTest
StubbsWSLtest
One might have the little cloud icon Sarah mentioned.

It looks like my enterprise managed laptop (which I'm due to trade in) is still blissfully OneDrive free, and hence why I as an instructor was confused and unable to help leaners in workshops.

bkmgit:
@jas58 Thanks. In gitbash may need to check `$HOME` rather than `$home`. It is possible that the variable is not set.

jas58:
[Folder of OneDrive Screenshots with Windows Explorer for context and Bash window](https://drive.google.com/drive/folders/1-H54anO4Hcz3F0EzBwg5fTVsX9c4kzoK?usp=sharing)

jas58:
[OneDriveScreenShots.zip](https://github.com/user-attachments/files/20129489/OneDriveScreenShots.zip)

jmjamison:
I'm another windows user but I've always ignored OneDrive.   I ran into a similar issue for the lc-shell lesson.  The solution I'm working on is to add a short note in the instructor notes and a reference in the episode 2.   I hope that I've added enough to help out a non-windows instructor.   (https://github.com/LibraryCarpentry/lc-shell/issues/299)

When teaching, Tim Dennis (@jt14den) got us in the habit of putting our operating system along with our names (ex. Jamie Win/Linux)  so both users and other instructors know where to send os specific questions.  It really helps to have at least one of each.

bkmgit:
Library carpentry have decided to put one drive as supplementary material with a note in the main lesson:                                                                  

- https://github.com/LibraryCarpentry/lc-shell/pull/302                                                                                    
- https://github.com/LibraryCarpentry/lc-shell/pull/303

PR #1487, opened 2025-04-23
Check goo script arguments to prevent data loss.

tkphd:
Often, the Instructor and Learners will run `goostats.sh` without arguments. The resulting usage message unhelpfully requests two files -- without warning that the second will be written into!

This PR updates both "goo" scripts to check file arguments for existence before continuing. 

`goostats.sh`:  If `$1` is absent, or `$2` is present, `exit 2` (invalid arguments).

`goodiff.sh`:  If `$1` or `$2` is absent, `exit 2` (invalid arguments).

"Usage" also updated to use `$0` rather than hard-coding the script names.


bkmgit:
Thanks for this. Any reason for changing the return code to 0 from 1 when the incorrect number of arguments are given?

PR #1493, opened 2025-06-18
adding linux possible clear exception

### LINKED ISSUE #1492: Differences between terminals for "clear"
### How could the content be improved?

On some terminals, including the default one that came with my PopOS/Gnome system, the `clear` command clears the scrollbuffer as well as the terminal output. This makes it impossible (as far as I know) to scroll up to see the content before the `clear` command was executed. This makes the last statement of the referenced section incorrect.

### Which part of the content does your suggestion apply to?

https://github.com/swcarpentry/shell-novice/blob/c3636d4da767d64a0bfb97d8d66ca2dd79334580/episodes/02-filedir.md?plain=1#L184-L188

sstevens2:
fixes #1492

I added language that for some linux systems `clear` might also clear the scrollbuffer.  

I'd also recommend making this callout a spoiler instead since it is optional content instead of content that should be highlighted.  I can switch this back if maintainers disagree.

### DIFF
--- a/episodes/02-filedir.md
+++ b/episodes/02-filedir.md
@@ -184,7 +184,7 @@ are **files** in the current working directory.
 ## Clearing your terminal
 
 If your screen gets too cluttered, you can clear your terminal using the
-`clear` command. You can still access previous commands using <kbd>↑</kbd>
+`clear -x` command. You can still access previous commands using <kbd>↑</kbd>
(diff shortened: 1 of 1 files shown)


bkmgit:
Investigating this.  Please try to wrap text at 80 characters to help readability of source files.

sstevens2:
I'll wrap to 80 char as requested. Going forward we might want to consider semantic line breaks instead as they are more readable and accessible.

PR #1503, opened 2025-09-02
Update file existence checks in goostats.sh

jsteyn:
The `-a` option for `if` does not work in all environments. I have tested the script below in three different environments, Git Bash, MobaXTerm and RedHat Linux and they all produce the following output:

# The Script
```
#!/bin/bash
touch checkForfile.md
echo WITH -a
if [ -a checkForfile.md ]; then echo yes; else echo no;fi
if [ -a NocheckForfile.md ]; then echo yes; else echo no;fi
echo WITH ! -a
if [ ! -a checkForfile.md ]; then echo no; else echo yes;fi
if [ ! -a NocheckForfile.md ]; then echo no; else echo yes;fi
echo WITH -e
if [ -e checkForfile.md ]; then echo yes; else echo no;fi
if [ -e NocheckForfile.md ]; then echo yes; else echo no;fi
echo WITH ! -e
if [ ! -e checkForfile.md ]; then echo no; else echo yes;fi
if [ ! -e NocheckForfile.md ]; then echo no; else echo yes;fi
echo WITH -f
if [ -f checkForfile.md ]; then echo yes; else echo no;fi
if [ -f NocheckForfile.md ]; then echo yes; else echo no;fi
echo WITH ! -f
if [ ! -f checkForfile.md ]; then echo no; else echo yes;fi
if [ ! -f NocheckForfile.md ]; then echo no; else echo yes;fi
```

# The output
```
 $ ./checkforfile.sh
WITH -a
yes
no
WITH ! -a
no
no
WITH -e
yes
no
WITH ! -e
yes
no
WITH -f
yes
no
WITH ! -f
yes
no
```

# Expected output
The expected output for all tests would be:
```
yes
no
```
However, the `-a` option does not work if used in conjunction with not ('!')

_If this pull request addresses an open issue on the repository, please add 'Closes #NN' below, where NN is the issue number._

_Please briefly summarise the changes made in the pull request, and the reason(s) for making these changes._

_If any relevant discussions have taken place elsewhere, please provide links to these._

<details>

For more guidance on how to contribute changes to a Carpentries project, please review [the Contributing Guide](CONTRIBUTING.md) and [Code of Conduct](https://docs.carpentries.org/topic_folders/policies/code-of-conduct.html).

[...]

PR #1509, opened 2025-09-29
corrected goostats.sh in zip

7anooch:
Corrects the copy of `goostats.sh` in the zipped data file to match the corrected version of the script (to match PRs #1505 and #1503).

No other changes were made to the zip file.


bkmgit:
@7anooch Thanks.

PR #1510, opened 2025-10-25
provide link to "Options" in glossary

cknuepfer:
The glossary reference in line 800 is not rendered because the link is missing.


bkmgit:
@cknuepfer Thanks.

PR #1513, opened 2025-10-31
Clarify `cp` behavior

katrinabrock:
Current wording of

> If given three or more arguments, `cp` throws an error such as the one below,

is not accurate. The intended meaning seems to be "If `cp` is given more than three filenames as arguments, it throws an error such as the one below."

Tried my best to reword in a way that is clear, concise, and accurate.

### DIFF
--- a/episodes/03-create.md
+++ b/episodes/03-create.md
@@ -515,7 +515,7 @@ cp: -r not specified; omitting directory 'thesis'
 ### Destination Can Be New Name or Directory
 
 When we use two arguments with `mv` or `cp`,
-the second argument can be the locationa and new name
+the second argument can be the location and new name
 we want our file to have. Alternatively,
 if we provide the path to an existing directory as the second argument,
 `mv`/`cp` will interpret that to mean that we want the file copied/moved


episodes/03-create.md
```diff
@@ -726,11 +726,15 @@ $ cp minotaur.dat unicorn.dat basilisk.dat
 
 ## Solution
 
-When `cp` is given two arguments and the second is a destination directory
-`cp` copies the files to the destination directory.
```
bkmgit: It is helpful to distinguish between two and three arguments because for two arguments, if the second argument is not a directory but a filename, then the first file is copied and renamed.

episodes/03-create.md
```diff
@@ -726,11 +726,15 @@ $ cp minotaur.dat unicorn.dat basilisk.dat
...
+When `cp` is given three or more arguments,
+the last argument is interpretted as the "destination
+directory".
+In that case, `cp` copies all files (or directories)
+into that "destination directory".
```
bkmgit: The above is nicely explained.  Can it be made shorter?

episodes/03-create.md
```diff
@@ -726,11 +726,15 @@ $ cp minotaur.dat unicorn.dat basilisk.dat
...
+If the final argument is not a directory,
+it's unclear what should be copied where so 
+`cp` throws an error such as the one below.
```
bkmgit:
```suggestion
If the final argument is not a directory,
the behavior is undefined, so 
`cp` throws an error such as the one below.
```

katrinabrock: I think "undefined behavior" might be a phrase that learners are unfamiliar with. I'll try to think of a way to say the same thing in simpler words.

episodes/03-create.md
bkmgit: Having it in the main text is a good idea.

bkmgit: Perhaps move this out of the challenge block and make it something that is demonstrated.

bkmgit:
Thanks. Merged.

PR #1515, opened 2025-11-07
Gentler introduction to recursion

katrinabrock:
This is an unsolicited PR so feel free to close without merging you don't think this needs fixing. Goal is to actually explain what learners need to know about recursion rather than linking out to wikipedia.

Changes:
- Explain the concept in simple words when first introduced
- Remind learners of the in-context meaning throughout

This is unfortunately a net add of content. I think it is possible to make the explanation(s) more concise, but likely will still be a net add since it adds an explanation that wasn't there before. I think overall it makes the lesson more clear to a learner who is not already familiar with the concept of recursion.

### DIFF
--- a/episodes/03-create.md
+++ b/episodes/03-create.md
@@ -89,11 +89,22 @@ in a single operation:
 $ mkdir -p ../project/data ../project/results
 ```
 
+Let's use `ls -F` to list the new directories we just created in the
+`project` directory:
+
+```bash
+$ ls -F ../project
+```
+
+```output
+data/  results/
+```
+
 :::::::::::::::::::::::::::::::::::::::::  instructor
 
-Discussion of `ls -R` has been removed in favor of
-introducing it in `cp` context, and explaining more deeply
-in `grep`/`find` context.
+One could use `ls -R` to list a directory hierarchy, but rather
+than introducing it in `cp` context, `find` can be used instead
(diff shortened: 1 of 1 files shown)


learners/reference.md
```diff
@@ -160,6 +160,12 @@ so that the shell will not try to expand the `*` wildcard.
 :   (REPL): A [command-line interface](#command-line-interface) that reads a command from the user,
 executes it, prints the result, and waits for another command.
 
+[recursive]{#recursive}
+:   A recursive precudure is one where one step of the procedure is to apply the procedure itself
+(usually to a different input).
+In the context of shell, it often refers to applying a command to a directory and to all directories
+contained within that directory.
```
bkmgit:
```suggestion
:   A recursive procedure is one where the output of an iteration
is the input of another iteration of that procedure.
```

Is context of the shell lesson helpful? Maybe should be added to
other terms if it is.

omsai:
@katrinabrock I second the suggestion from @bkmgit about describing recursion in the glossary, but I do not think we should be discussing the concept in the main text because it adds unnecessary cognitive load. I would only want the learners to know about `cp -r` for copying directories with the minimum explanation for why that is necessary. Using `find` or `grep` is more typical than `ls -R`; I don't see as much use of `ls -R` to focus on it here.

PR #1519, opened 2025-11-28
Remove note about nano and the SWC Windows Installer

tobyhodges:
The Instructor Notes refer to the Software Carpentry Windows Installer, and a potential issue with `nano` when using it. We have not used the Installer for a long time and have received no reports of `nano` being unavailable on Windows systems after following the setup instructions. This PR removes that item from the Instructor Notes.


dpshelio (approved):

bkmgit:
Thanks.

